PR #15, opened 2020-05-06
Fix move on multiple selection files/folders. Fix same id file/folder…

juanitopons:
- FIX: Multiple move fixed for folders & files.
- FIX: Bad selection on same folder/file id


juanitopons:
Actualizado!
https://github.com/internxt/drive-web/pull/15/commits/f2209503c2c742cfb2b36e5fd8c6e05ad2fcea28

PR #69, opened 2021-05-10
Fix Folders Sync sign

Siralde:
After uploading one file, the code call getFolderContent making the app to render each time after one file was uploaded to the network.

Now after all of the files are uploaded the app is rendered again. Solving the problem of the folder sync sign

### DIFF
--- a/src/components/xcloud/XCloud.js
+++ b/src/components/xcloud/XCloud.js
@@ -992,7 +992,7 @@ class XCloud extends React.Component {
           <NavigationBar
             showFileButtons={true}
             showSettingsButton={true}
-            By={this.createFolder}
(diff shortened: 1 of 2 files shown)


src/components/xcloud/XCloud.js
```diff
@@ -991,7 +992,7 @@ class XCloud extends React.Component {
           <NavigationBar
             showFileButtons={true}
             showSettingsButton={true}
-            createFolder={this.createFolder}
+            By={this.createFolder}
```
JAlbertoGonzalez: By miss typo

JAlbertoGonzalez (changes requested):

PR #149, opened 2021-08-30
[DW2-202]: feature/display-progress-folder-upload

AlexMenor:
Until now uploading a folder didn't update the notification at all and user could be waiting with no feedback for a while. 

- Added item counting under a root folder
- Update the notification after each item under root folder has been uploaded


AlexMenor:
"countItemUnderRoot" function has been added inside createFolderTreeStructureThunk.ts as I think it probably is not needed anywhere else. Let me know if it's better to extract it from the beginning  or on demand @JAlbertoGonzalez

PR #253, opened 2021-12-17
[_]: fix/avoid-conflict-in-e2e-tests

AlexMenor:


AlexMenor:
The reason of the conflicts in e2e tests was recaptcha, not that the users already exist, but we should increase the bytes of the random emails anyway as was too low

PR #256, opened 2021-12-20
feature/use-sdk-register

sraver:
On here we are updating the register process to use the register function created on the SDK's auth module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "@iconscout/react-unicons": "^1.1.6",
     "@internxt/inxt-js": "=1.2.14",
     "@internxt/lib": "1.1.2",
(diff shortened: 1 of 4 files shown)


sraver:
@carlossalasamper @AlexMenor make sure not to merge this request until it's carefully tested

carlossalasamper:
@sg-gs We need to solve the Recatcha issue with e2e tests before merge this

sraver:
> @sg-gs We need to solve the Recatcha issue with e2e tests before merge this

@carlossalasamper I closed this PR in order to open another one with the whole auth module, where register and login is replaced.

https://github.com/internxt/drive-web/pull/258

PR #264, opened 2021-12-31
feature/storage-module

sraver:
The Storage module on the SDK includes all the calls to endpoints related to files and folders management.

On here, all the requests to those endpoint have been replaced by the use of the SDK Storage module.

Also some of the interfaces have been moved to the SDK, in an attempt of start centralizing them.

### DIFF
--- a/src/app/drive/services/file.service/index.ts
+++ b/src/app/drive/services/file.service/index.ts
@@ -46,11 +46,11 @@ export function deleteFile(fileData: DriveFileData): Promise<void> {
 }
 
 export async function moveFile(
-  file: DriveFileData, destination: number, bucketId: string
+  fileId: string, destination: number, bucketId: string
 ): Promise<StorageTypes.MoveFileResponse> {
   const storageClient = createStorageClient();
   const payload: StorageTypes.MoveFilePayload = {
-    fileId: file.fileId,
+    fileId: fileId,
     destination: destination,
     bucketId: bucketId,
     destinationPath: uuid.v4()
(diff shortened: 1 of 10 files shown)


src/app/drive/services/file.service/index.ts
```diff
@@ -41,49 +31,52 @@ export function updateMetaData(fileId: string, metadata: DriveFileMetadataPayloa
...
-
-  return response;
 }
 
 async function fetchRecents(limit: number): Promise<DriveFileData[]> {
-  const response = await httpService.get<DriveFileData[]>(`/api/storage/recents?limit=${limit}`);
-
-  return response;
+  const storageClient = createStorageClient();
+  return storageClient.recentFiles(limit);
```
sg-gs: `getRecentFiles` could be more descriptive

src/app/drive/services/folder.service.ts
```diff
@@ -210,35 +159,35 @@ async function fetchFolderTree(folderId: number): Promise<{
...
 }
 
-export async function moveFolder(folder: DriveFolderData, destination: number): Promise<MoveFolderResponse> {
-  const user = localStorageService.getUser() as UserSettings;
+export async function moveFolder(
+  folder: DriveFolderData, destination: number
+): Promise<StorageTypes.MoveFolderResponse> {
+  const storageClient = createStorageClient();
+  const payload: StorageTypes.MoveFolderPayload = {
+    folder: folder,
```
sg-gs: Why not just use `folderId` instead of requiring the entire data structure?

sg-gs (changes requested):

src/app/drive/services/folder.service.ts
sraver: I think because I was driven by the old code, but it's better only the ID. Changing it right now

src/app/drive/services/file.service/index.ts
sraver: I guess is a philosophy thing.. I'm quite of a detractor on the use of "get" word on getters. We can talk the reasons of it some day, but I'll change it now if you feel reads better.

sg-gs (approved):

PR #272, opened 2022-01-13
[SDK-15, SDK-21]: feature/add-users-share-referrals-modules

sraver:


src/app/auth/services/auth.service.ts
```diff
@@ -188,31 +189,25 @@ export const changePassword = async (newPassword: string, currentPassword: strin
...
-  }
-
-  if (response.status !== 200) {
-    analyticsService.track(email, 'error');
-    throw new Error(data.error);
-  }
+  const usersClient = createUsersClient();
 
-  analyticsService.track(email, 'success');
+  return usersClient.changePassword(<ChangePasswordPayload>{
```
sg-gs, line 197: Be careful, this is not working as intended. Changing the password with these changes corrupts something involved with the mnemonic, and the user won't be able to regenerate his/her old mnemonic nor download any file uploaded before changing the password. 

Steps for testing it: 
1- Signup
2- Upload a file, a pdf for example, that gets corrupted with any minimal change.
3- Change password
4- Logout
5- Login
6- Download the previously uploaded file

sg-gs (changes requested):

src/app/auth/services/auth.service.ts
sraver: @sg-gs I just tried again and it works fine here :/ What error are you having?

sraver: Since we already found what was the error here I'm closing this.

Did you review all the code or stopped when found this?

sg-gs (approved):

PR #284, opened 2022-01-19
[]: feature/improve-signup-ux

aboredvaro:


src/app/auth/views/SignUpView/SignUpView.tsx
```diff
@@ -307,131 +325,181 @@ const SignUpView = (props: SignUpViewProps): JSX.Element => {
...
-            register={register}
-            required={true}
-            minLength={{ value: 1, message: 'Password must not be empty' }}
-            error={errors.confirmPassword}
-          />
-
-          <div className="mt-1 mb-2">
-            <span className="text-red-60 text-sm font-medium">{bottomInfoError}</span>
+    <>
+      {isDoingRegister && (
```
sg-gs: Try to use a ternary operator here like: `isDoingRegister? <Component1>: <Component2>` in order to keep related components together. This change also prevents React of doing both checks to determine if the component should be rendered when one check is enough

sg-gs (changes requested):

sg-gs (approved):

PR #295, opened 2022-01-27
[_]: fix/blank-screen-on-signup

AlexMenor:
- What's happening currently is that If at any moment the user is authenticated but is not initialised, an empty div shows (the blank screen)
- In this PR that div is replaced by the animation that @aboredvaro made.
- The animation has some steps that were previously controlled by props, I put it inside instead as it doesn't rely on any particular external logic.
- The animation has been renamed as it is not used in SignUp anymore, but instead when an uninitialised user is authenticated.

We could also plan a refactor for the sign up page as it is critical for bounces and is extremely confusing right now, it is scary to modify it.


sg-gs (approved):
Considering a refactor for the signup is a great idea @AlexMenor

PR #301, opened 2022-02-01
[DW-430] feature/link-share-folder

sraver:
The modal of generating a share link has been modified so we can also share a folder.

Depending the kind of resource we are sharing the link is generated in one way or another.


sg-gs (approved):
Not merging this until download is available, isn't it? @sraver

sraver:
> Not merging this until download is available, isn't it? @sraver

We can, I created the _epic_ branch so we could split the task on different PRs. Like to I'll PR the other one against the _epic_. @sg-gs

sg-gs:
Oh, okay, haven't seen that

PR #304, opened 2022-02-08
feature/download-shared-folder

sraver:


sg-gs:
- No feedback/loading animation is being provided when the folder size is being calculated. Better check with @JoanMora @aboredvaro 
- Folder too big dialog 'Dismiss' and close buttons are not working 
- If I click 'Share' for a big folder and the request to calculate the size lasts, the 'Share' button is still enabled, therefore I can fire the request infinite times clicking the 'Share' button

@sraver

sraver:
Point 2 and 3 are fixed. Point 1 I cant replicate, it actually loads the loading icon on my side :/

sg-gs (approved):

PR #334, opened 2022-03-14
fix/upload-url-generation

sraver:
Fix the URL generation for uploading a file when the app should not use a proxy


AlexMenor:
Hi @sraver, I believe we had an env variable for this

sraver:
> Hi @sraver, I believe we had an env variable for this

Nice, I added the check into the code so it works as it seems would be expected.

AlexMenor:
@sraver can we avoid relying on the host somehow? Even if its including env variables.

sraver:
> @sraver can we avoid relying on the host somehow? Even if its including env variables.

I think we could, maybe checking the NODE_ENV, but I didn't want to change much the original logic cos I didn't know the idea or purpose behind.

@sg-gs You think we can switch the host checking for the env??

sraver:
Also, on another note. I added the same checking on the construction of the download URLs, otherwise on local the downloads were not working for me.

Those: https://github.com/internxt/drive-web/pull/334/files#diff-262c5a71d08e190d6ade5e990252eb1421ba2cabd7a2a524626956ff75a1ac0aR186-R190

AlexMenor:
> > @sraver can we avoid relying on the host somehow? Even if its including env variables.
> 
> I think we could, maybe checking the NODE_ENV, but I didn't want to change much the original logic cos I didn't know the idea or purpose behind.
> 
> @sg-gs You think we can switch the host checking for the env??

I was thinking that the logic behind using them should not be in the code, and rather just be configurable so the source of truth are just the env variables

PR #335, opened 2022-03-15
[DW-435] feature/unify-download-file

sraver:
Use same strategy to download a shared file as the one used when downloading files from a shared folder.


sraver:
@sg-gs I believe that was the idea, please check if it makes sense at all because it took me quite a while to understand all the logic of what is happening when downloading a file, and I'm not yet sure that I have the conditions on strategy selection correct. I used the same as when a file is downloaded form a shared folder tho, so it should be fine i think.

PR #407, opened 2022-08-11
[_]:feat/integrate-network-web-package

fabioespinosa:
~**WIP (don't merge):**~

Needs to yarn link the package for @internxt/network-web to work

Issues: 

- Types are not working (from the package), they are not being exported from the package (trying to figure out why)

~Purpose of this PR is just to test the package and approve changes in the package, not in drive-web, not yet~


src/app/network/download/v2.ts
```diff
@@ -1,34 +1,34 @@
-import { Network } from '@internxt/sdk/dist/network';
 import { sha256 } from '../crypto';
-import { NetworkFacade } from '../NetworkFacade';
+import { NetworkWeb } from '@internxt/network-web/_bundles/prod';
```
fabioespinosa: not sure if it is a good idea to import it like this (under bundles)

src/app/network/upload.ts
```diff
@@ -19,6 +19,7 @@ interface IUploadParams {
   mnemonic: string;
   progressCallback: UploadProgressCallback;
   abortController?: AbortController;
+  parts?: number;
 }
 
 export function uploadFileBlob(
```
fabioespinosa: this function should be removed (the package replaces it)

src/app/network/download/v2.ts
fabioespinosa: fixed

sg-gs (approved):

sg-gs:
We no longer work on this line, fixed with lasts improvements.

fabioespinosa:
Hope it helped a little bit :) :)

PR #409, opened 2022-08-17
[_]: feat/removed-limited-views-feature

vicblab:
Now the default and only max views of a shared link is infinite.


src/app/core/factory/sdk/index.ts
```diff
@@ -47,9 +47,9 @@ export class SdkFactory {
   }
 
   public createShareClient(): Share {
-    const apiUrl = this.getApiUrl();
+    const apiUrl = this.getApiV2Url();
     const appDetails = SdkFactory.getAppDetails();
-    const apiSecurity = this.getApiSecurity();
+    const apiSecurity = this.getNewApiSecurity();
```
PixoDev, line 65: whats the difference between the old and the new one?

fabioespinosa (approved):

PR #413, opened 2022-08-30
Feature/wt 6 move items to trash

foxkdev:
- Added Use Case to Move items to trash
- Modify views to call directly to use case
- Pending: Check with API.
- commented delete items to wait task to delete items


vicblab (approved):
Merged locally with no conflicts

PR #417, opened 2022-08-30
Feature/refactor use cases

foxkdev:
- Refactor to use Use Case Get Trash


vicblab (approved):
Changes merged locally without conflicts

PR #420, opened 2022-09-02
[WT-25]: Removed decimals from task logger progress

larry-internxt:
- Removed decimals from task logger progress


fabioespinosa (approved):
Seems ok to me

PR #444, opened 2022-09-28
[WT-129] test: make sure cypress works

JoanVicens:
- Updated sing up test to match new singup page
- Added `data-test` properties to html elements in order to be able to select them on cypress test
- Updated download file service to use `PatchedStreamApi` when is being tested with cypress. Without this the file was never downloaded making the test to always fail.

⚠️ There is a lot of linting going on


src/app/core/components/Navbar/AccountPopover.tsx
```diff
@@ -64,7 +64,7 @@ export default function AccountPopover({
       </Item>
       <Link
         to="/preferences"
-        className="flex cursor-pointer items-center py-2 px-3 text-gray-80 hover:text-gray-80 hover:bg-gray-1 active:bg-gray-5 no-underline"
+        className="flex cursor-pointer items-center py-2 px-3 text-gray-80 no-underline hover:bg-gray-1 hover:text-gray-80 active:bg-gray-5"
```
fabioespinosa, line 67: Is this from the prettier auto save rule? or is there any change of classes here? 🤔

src/app/drive/services/download.service/downloadFile.ts
```diff
@@ -97,19 +99,22 @@ export default async function downloadFile(
...
+  const isCypress = window['Cypress'] !== undefined;
 
   const writeToFsIsSupported = 'showSaveFilePicker' in window;
   const writableIsSupported = 'WritableStream' in window && streamSaver.WritableStream;
 
   let support: DownloadSupport;
 
-  if (isBrave) {
+  if (isCypress) {
+    support = DownloadSupport.PatchedStreamApi;
```
fabioespinosa, line 116: Nice find 👌

src/app/core/components/Navbar/AccountPopover.tsx
fabioespinosa: nevermind, it seems its rearranging of existing classes

fabioespinosa (approved):

src/app/drive/services/download.service/downloadFile.ts
```diff
@@ -57,10 +57,12 @@ function getBlobWritable(filename: string, onClose: (result: Blob) => void): Blo
         closed: Promise.resolve(undefined),
         desiredSize: 3 * 1024 * 1024,
         ready: Promise.resolve(undefined),
-        releaseLock: () => { null; },
+        releaseLock: () => {
+          null;
```
fabioespinosa: Sonar seems to think this is a bug

fabioespinosa: Which makes sense since it is not returning anything

fabioespinosa: Maybe we can remove the null and just comment 

// no op

redac: can we try this?

fabioespinosa: we can! Nice to see a random person commenting on this PR

fabioespinosa: @JoanVicens on the request of @redac :D let's try removing the null and adding // no op

redac: 👋 currently lurking while reading up on the project before contributing.

fabioespinosa: go for it, if you need anything let us know :)

JoanVicens: That made it!

fabioespinosa: Awesome!

redac: > go for it, if you need anything let us know :)

Hi again, it looks like i cannot access the @internxt/inxt-js package even with a valid token in .npmrc when running yarn install.. is this normal?

image

PR #448, opened 2022-10-03
[_]: Fix/checkout id not being retrieved

fabioespinosa:
- Fixes sessionId not appearing in analytics

### DIFF
--- a/src/app/payment/views/CheckoutSuccessView/CheckoutSuccessView.tsx
+++ b/src/app/payment/views/CheckoutSuccessView/CheckoutSuccessView.tsx
@@ -4,15 +4,22 @@ import { userThunks } from 'app/store/slices/user';
 import navigationService from 'app/core/services/navigation.service';
 import { AppView } from 'app/core/types';
 import analyticsService from 'app/analytics/services/analytics.service';
+import { useCallback } from 'react';
 
 const CheckoutSuccessView = (): JSX.Element => {
   const dispatch = useAppDispatch();
 
-  useEffectAsync(async () => {
+  const onCheckoutSuccess = useCallback(async () => {
     await dispatch(userThunks.refreshUserThunk());
-    await analyticsService.trackPaymentConversion();
+    try {
+      await analyticsService.trackPaymentConversion();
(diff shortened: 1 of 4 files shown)


fabioespinosa:
@JoanMora

src/app/payment/views/CheckoutSuccessView/CheckoutSuccessView.tsx
```diff
@@ -10,7 +10,7 @@ const CheckoutSuccessView = (): JSX.Element => {
 
   useEffectAsync(async () => {
     await dispatch(userThunks.refreshUserThunk());
-    analyticsService.trackPaymentConversion();
+    await analyticsService.trackPaymentConversion();
```
sg-gs: do not make the checkout depend on analytics

sg-gs (changes requested):

src/app/payment/views/CheckoutSuccessView/CheckoutSuccessView.tsx
fabioespinosa: it has a try catch inside though 🤔

sg-gs: It would be better not to rely on service as the component is responsible for its own behavior. So if the component has to handle the checkout success and sending the analytics is not critical for handling the success is the component who is responsible for ensuring that behavior really happens as expected. 

Therefore, if someone decides to remove the `try-catch` from the analytics, all the App will not be broken, and the side effects are avoided. 

'The checkout is suddenly broken'. Then you find that someone removed something non-directly related to the checkout but still break it, which is very serious, troublesome to debug and a nightmare that always returns to you.

In summary, a good idea will be to extract this function from the hook if possible and add a `try-catch` there to ensure that the component accomplishes its responsibilities always. Then the dependency from the analytics is avoided.

sg-gs (approved):

JoanMora (commented):
Can we send an error here?
window.rudderanalytics.track('Payment Conversion Error', {
msg: string
})

PR #453, opened 2022-10-06
[WT-104]: Feature/thumbnails

larry-internxt:
- [WT-102]: Generate thumbnail for .png, .jpeg, .img
- [WT-103]: Send the thumbnail to be uploaded along with the file, on every upload
- [WT-104]: Show the thumbnails when browsing in mosaic

[WT-102]: https://inxt.atlassian.net/browse/WT-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WT-103]: https://inxt.atlassian.net/browse/WT-103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WT-104]: https://inxt.atlassian.net/browse/WT-104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerGridItem/DriveExplorerGridItem.scss
+++ b/src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerGridItem/DriveExplorerGridItem.scss
@@ -25,6 +25,11 @@
         height: 50%;
         width: 50%;
       }
+
+      .file-thumbnail {
+        height: 100%;
+        width: 100%;
+      }
     }
   }
 
--- a/src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerGridItem/DriveExplorerGridItem.scss
+++ b/src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerGridItem/DriveExplorerGridItem.scss
@@ -1,43 +1,5 @@
 @layer components {
-  .file-grid-item {
-    @apply relative bg-white p-4 box-border rounded-lg;
-
-    &:hover, {
-      @apply bg-neutral-10;
-    }
-
-    &.selected {
-      @apply bg-blue-10;
-
-      box-shadow: inset 0 0 0 2px var(--color-blue-50);
-    }
-
-    &.is-dragging {
-      @apply opacity-50;
-    }
-
-    .file-grid-item-icon-container {
-      height: 70%;
-
-      @apply flex justify-center items-center w-full;
-
-      .file-icon {
-        height: 50%;
-        width: 50%;
-      }
-
-      .file-thumbnail {
-        height: 100%;
-        width: 100%;
-      }
-    }
-  }
-
-  .file-grid-item-actions-button {
-    @apply top-2 right-2 rounded-1/2 w-5 h-5 bg-white cursor-pointer font-bold text-blue-60 absolute transition opacity-0 group-hover:opacity-100;
-  }
-
-  .file-grid-item-name-span {
-    @apply whitespace-nowrap overflow-hidden overflow-ellipsis text-neutral-900 text-base px-1 hover:underline;
+  .grid-item-shadow {
+    box-shadow: inset 0 0 0 2px var(--color-blue-50);
   }
 }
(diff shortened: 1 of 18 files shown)


src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerGridItem/DriveExplorerGridItem.scss
```diff
@@ -25,6 +25,11 @@
         height: 50%;
         width: 50%;
       }
+
+      .file-thumbnail {
+        height: 100%;
```
fabioespinosa: Maybe with this we can use tailwind's `w-full h-full` so we keep everything in the same file

src/app/drive/services/file.service/uploadFile.ts
```diff
@@ -72,7 +73,21 @@ export async function uploadFile(
...
-    const response = await storageClient.createFileEntry(fileEntry);
+    let response = await storageClient.createFileEntry(fileEntry);
+    if (!response.thumbnails) {
+      response = {
+        ...response,
+        thumbnails: []
+      };
+    }
+
+    const thumbnail = await generateThumbnailFromFile(file, response.id, userEmail, isTeam);
```
fabioespinosa: Perhaps naming this as `generatedThumbnail` makes it clearer to then use `generatedThumbnail.thumbnail` instead of `thumbnail.thumbnail`

src/app/drive/services/thumbnail.service.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export interface ThumbnailToUpload {
+  fileId: number;
+  size: number;
+  type: string;
+  content: File;
+}
+
+const getImageThumbnail = async (file: File): Promise<File | null> => {
+  return await new Promise((resolve) => {
```
fabioespinosa: Here you can remove async and await to achieve the same functioanlity

fabioespinosa: Also does Resizer.imageFileResizer ever fail? if passed a broken image perhaps is good to catch the error in this Promise `reject` function

src/app/drive/types/file-types.ts
```diff
@@ -208,4 +208,11 @@ export const fileExtensionPreviewableGroups: fileExtensionsPreviewableDictionary
   [FileExtensionGroup.Default]: [],
 };
 
+export const thumbnailableImageExtension: string[] = [...imageExtensions['jpg'], ...imageExtensions['png'],
+...imageExtensions['bmp'], ...imageExtensions['gif']];
+
+export const thumbnailablePdfExtension: string[] = [...pdfExtensions['pdf']];
```
fabioespinosa: Here you can just leave it as:
`export const thumbnailablePdfExtension: string[] = pdfExtensions['pdf'];`

Since there will never be another group for thumbnailPdfExtensions, since its not generic like images

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -42,18 +49,72 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show }: FileViewerP
...
+
+            const thumbnail = await getThumbnailFrom(fileUpload);
+
+            //Don't upload generated thumbnail if it match in size and type with currentThumbnail
+            if (thumbnail && thumbnail.file && thumbnail.type &&
+              (!currentThumbnail ||
+                (currentThumbnail && (Number(currentThumbnail.size) !== Number(thumbnail.file.size)
+                  || String(currentThumbnail.type) !== String(thumbnail.type)
+                  || (Number(currentThumbnail.max_width)) !== (Number(thumbnail.max_width))
+                  || (Number(currentThumbnail.max_height)) !== (Number(thumbnail.max_height)))))) {
```
fabioespinosa: This `if` looks a bit complicated and difficult to read, perhaps creating a function called `compareThumbnails` is more readable, and then you can remove the comment as it is obvious what it is doing

fabioespinosa (changes requested):
Small detail with readability `if`

fabioespinosa (approved):
good job

PR #456, opened 2022-10-10
[_]:(feature) get discount coupon

xabg2:
get the black firday discount coupon that comes from website to do the checkout session

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "@iconscout/react-unicons": "^1.1.6",
     "@internxt/inxt-js": "=1.2.21",
     "@internxt/lib": "^1.2.0",
-    "@internxt/sdk": "1.4.9",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "@iconscout/react-unicons": "^1.1.6",
     "@internxt/inxt-js": "=1.2.21",
     "@internxt/lib": "^1.2.0",
-    "@internxt/sdk": "1.4.9",
+    "@internxt/sdk": "^1.4.10",
```
fabioespinosa: I guess for this we want to leave it without the `^`

fabioespinosa (changes requested):
Fix the requested changes and all good

fabioespinosa (approved):

PR #457, opened 2022-10-14
[WT-145] feature/Upload file plain_name and use new types of files and folders

CandelR:
- Enables to send plain_name with files data (folders plain_name is managed in drive-server)


src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerListItem/DriveExplorerListItem.tsx
```diff
@@ -96,7 +96,8 @@ const DriveExplorerListItem = ({ item }: DriveExplorerItemProps): JSX.Element =>
             title={items.getItemDisplayName(item)}
             onClick={onNameClicked}
           >
-            {items.getItemDisplayName(item)}
+            {/* TODO: CHANGE getItemDisplayName OF internxt/lib IN ORDER TO USE plain_name*/}
+            {item.plain_name ?? item.name}
```
fabioespinosa: Given we want this change to be silent yet, we want to store the file name unencrypted in the backend but not yet use it to show it in the front end.

Since we have to first unencrypt the existing file names, and we haven't done that yet.

So for the moment anything that is to display name, should still work the way it is. We just want to send the file name unencrypted.

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -337,51 +337,53 @@ export const uploadItemsParallelThunk = createAsyncThunk<void, UploadItemsPayloa
...
-          }
-
-          const castedError = errorService.castError(err);
-          const task = tasksService.findTask(tasksIds[index]);
-
-          if (task?.status !== TaskStatus.Cancelled) {
             tasksService.updateTask({
               taskId: taskId,
-              merge: { status: TaskStatus.Error },
+              merge: { status: TaskStatus.Success },
```
fabioespinosa, line 524: curious why we changed this to Success if it is on a .catch 🤔

fabioespinosa: Or is it because the diff makes it appear so? is it not on a catch?

fabioespinosa (changes requested):
Some slight changes in display and all good

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
CandelR: It looks like it has been changed because of the changes prettier has made, but in reality this lines of code
`if (task?.status !== TaskStatus.Cancelled) {`, `merge: { status: TaskStatus.Error },` are in the 376 and 379 now.

src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerListItem/DriveExplorerListItem.tsx
CandelR: Perfect, I'll change it back to the way it was!

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
fabioespinosa: Ah I see, makes sense

sg-gs (approved):

PR #460, opened 2022-10-14
[_]: Refactor/shared links password

fabioespinosa:
(this one is forlearning purposes only)

To show what changed with design PR


fabioespinosa:
@masterprog-cmd este PR es solo para que veas que cambio con tu rama para poder conectar shared links con password, no tienes que mergearlo ☮️

PR #471, opened 2022-10-21
[WT-155] Feature/Added Photos Trash UI

CandelR:
- Added photos trash UI

**IMPORTANT NOTES**
- ⚠️ The delete photos functionality does not work, first we have to change the call to delete the photos to send them to Trash (TRASHED) and not delete them directly (DELETED).⚠️


sg-gs:
Hey @CandelR, closing this at least for now, as it has been forbidden by the product roadmap. If in the future we come again, we can revive this.

PR #477, opened 2022-10-31
WT-169  feature/Hot reloading of drive-web with .env

CandelR:
- Added hot reloading in development env

⚠️ Need to test on Windows with `yarn dev:windows` ⚠️

- [ ] Test it on Windows SO

**Description:**
The https://github.com/kHRISl33t/runtime-env-cra package has been added so that environment variables can be modified while running drive-web in a local environment.

To change the env variables in running time, run the next commands and modify the `runtime-env.js` file: 
- Use `yarn dev:start`  to run on Mac OS or any Linux SO
- Use `yarn dev:windows` to run on Windows SO


src/app/analytics/services/analytics.service.ts
```diff
@@ -8,6 +8,8 @@ import localStorageService from 'app/core/services/local-storage.service';
 import { DevicePlatform, SignupDeviceSource } from 'app/core/types';
 import { DriveItemData } from 'app/drive/types';
 import { AnalyticsTrack } from '../types';
+import EnvService from 'app/core/services/dynamicEnv.service';
```
fabioespinosa: use relative imports here

PR #480, opened 2022-10-31
[WT-160]: Feat/share domain

fabioespinosa:


fabioespinosa:
https://github.com/internxt/drive-web/pull/537

PR #491, opened 2022-11-09
[WT-186] feature/add cypress test to drive view

CandelR:
- Added cypress tests:
   - Rename file
   - Download/remove folder 
   - Download Shared file 

Notes:
- ℹ️  Updated cypress in order to able navigation across domains in the same test 
- 🔴  At the moment it is not possible to test the functionality of uploading folders.


test/e2e/cypress/integration/share-file.spec.ts
```diff
@@ -1,28 +0,0 @@
-import * as path from 'path';
```
JoanVicens, line 1: Why is this test deleted?

JoanVicens:
The tests fail for me. On the second test of `signup.spec.ts` when executes the beforeEach and visits '/' it goes to /login so it fails and does not store the credentials. I'm running them on chrome 107:
imatge

I tested locally with a logged user and it redirects to /app so I assume it's something to do with cypress. Besides that, looks good to me

test/e2e/cypress/integration/share-file.spec.ts
CandelR: Because it doesn't work and I already check that feature in the new test `download-shared-file.spec.ts`. Where I first share the file and then with the copied shared link I download it. 
If you consider it more appropriate I could separate it.

CandelR:
> The tests fail for me. On the second test of `signup.spec.ts` when executes the beforeEach and visits '/' it goes to /login so it fails and does not store the credentials. I'm running them on chrome 107: imatge
> 
> I tested locally with a logged user and it redirects to /app so I assume it's something to do with cypress. Besides that, looks good to me

Okay! I check why and upload the fix! Maybe it has something to do with having upgraded to Cypress and activated this `experimentalSessionAndOrigin`

test/e2e/cypress/integration/share-file.spec.ts
JoanVicens: Nah, for me it's okay to delete it then 🙆🏼‍♂️

fabioespinosa (approved):

fabioespinosa:
Failing before & after this PR:

- change-password.spec.ts 
- download-shared-file.spec.ts
- unauthorized.spec.ts

PR #492, opened 2022-11-11
[_]: Feat/add-tracks-rudderanalytics

xabg2:
Tracks added for:
* Login
* Logout
* Upload file/folder
* Drop file/folder


src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -190,6 +194,11 @@ export const uploadFolderThunkNoCheck = createAsyncThunk<void, UploadFolderThunk
     try {
       root.folderId = currentFolderId;
 
+      window.rudderanalytics.track('Folder Upload Started', {
```
fabioespinosa: We want this to be imported from analytics service

src/app/analytics/services/analytics.service.ts
```diff
@@ -93,40 +93,32 @@ export function identifyPlan(newValue: number) {
   }
 }
 
-export function trackSignOut() {
-  /* window.analytics.track(AnalyticsTrack.SignOut);
-  window.analytics.reset(); */
+export function rudderanalyticsSignOut() {
+  window.rudderanalytics.track('User Logout');
 }
```
fabioespinosa, line 97: The only file that should contain any method or function with the word 'rudder' should be analytics.service.ts

All others should be called "trackSignout"

Later if we want to change from rudder to other analytics, we'd just have to change it here.

Tracking is not dependant on if we use rudder or any other service, so calling it trackSignOut is more maintainable through time than rudderanalyticsSignOut

fabioespinosa (changes requested):
checking it

src/app/auth/services/auth.service.ts
```diff
@@ -140,9 +140,6 @@ export const doLogin = async (
       };
     })
     .catch((error) => {
```
sg-gs: Remove this `catch` as the track is no longer there anymore

src/app/drive/services/file.service/uploadFile.ts
```diff
@@ -25,18 +25,18 @@ export async function uploadFile(
...
 
   try {
-    analyticsService.trackFileUploadStart({
-      file_size: file.size,
-      file_type: file.type,
-      folder_id: file.parentFolderId,
-      email: userEmail,
-      platform: DevicePlatform.Web,
-    });
+    // analyticsService.trackFileUploadStart({
```
sg-gs: why this analytic is commented now?

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -296,6 +307,21 @@ function countItemsUnderRoot(root: IRoot): number {
...
 }
+function getItemsSize(root: IRoot): number {
+  let size = 0;
+
+  const queueOfFolders: Array<IRoot> = [root];
+
+  while (queueOfFolders.length) {
+    const folder = queueOfFolders.shift() as IRoot;
+
+    size += folder.childrenFiles?.map((file) => file.size).reduce((a, b) => a + b, 0) ?? 0;
```
sg-gs: why not just: 

`folder.childrenFiles?.forEach((file) => size += file.size);`

?

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -225,6 +246,7 @@ export const uploadItemsThunkNoCheck = createAsyncThunk<void, UploadItemsPayload
     const errors: Error[] = [];
     const tasksIds: string[] = [];
     const abortController = new AbortController();
+    console.log(parentFolderId, '-', files);
```
sg-gs: remove this `console.log`

sg-gs (changes requested):

src/app/drive/services/file.service/uploadFile.ts
xabg2: Because if we throw the track in this file, when we upload a folder, we'll have a track for every file inside into.

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -296,6 +307,21 @@ function countItemsUnderRoot(root: IRoot): number {
 
   return count;
 }
+function getItemsSize(root: IRoot): number {
+  let size = 0;
+
+  const queueOfFolders: Array<IRoot> = [root];
+
+  while (queueOfFolders.length) {
```
sg-gs: `while (queueOfFolders.length > 0)` is better to understand, I am not sure about the side effects of checking is the `length` attribute is present

sg-gs (changes requested):

JoanMora (commented):
Replace uuid for userId on `.identify()`  calls

JoanMora:
Great job @masterprog-cmd!

src/app/analytics/services/analytics.service.ts
```diff
@@ -93,40 +93,33 @@ export function identifyPlan(newValue: number) {
...
-    signup_device_source: string;
-    acquisition_channel;
-  };
-  userId: string;
-}): void {
-  /* window.analytics.identify(payload.userId, payload.traits);
-  window.analytics.track(AnalyticsTrack.SignUp, payload.properties); */
-  trackSignUpServer(payload);
+export function rudderanalyticsSignUp(email: string, uuid: string): void {
+  window.rudderanalytics.identify(uuid, { email, uuid: uuid }, () => {
```
JoanMora: Change uuid by userId in event properties

src/app/analytics/services/analytics.service.ts
```diff
@@ -93,40 +93,33 @@ export function identifyPlan(newValue: number) {
...
+    window.rudderanalytics.track('User Signup', { email });
+  });
+}
+
+export function rudderanalyticsSignUpError(email: string, error: string): void {
+  window.rudderanalytics.track('User Signup Error', { email: email, error: error });
```
JoanMora: Replace error by message when sending this event, also name of the track must be `User Signup Failed`
Screenshot 2022-11-23 at 11 08 05 a m

src/app/analytics/services/analytics.service.ts
```diff
@@ -403,9 +390,10 @@ const analyticsService = {
   identify,
   identifyUsage,
   identifyPlan,
-  trackSignOut,
-  trackSignIn,
-  trackSignUp,
+  rudderanalyticsSignIn,
```
JoanMora: Why are we bound by the analytics library?  The action is the same regardless of what library we are using rudder or any other. I would keep track+actionToTrack as the naming convention. The name of the library used should only appear when this is loaded. @sg-gs

src/app/analytics/services/analytics.service.ts
```diff
@@ -93,40 +93,33 @@ export function identifyPlan(newValue: number) {
...
-export function trackSignOut() {
-  /* window.analytics.track(AnalyticsTrack.SignOut);
-  window.analytics.reset(); */
+export function rudderanalyticsSignOut() {
+  window.rudderanalytics.track('User Logout');
 }
 
-export function trackSignIn(payload: { email: string; userId: string }): void {
-  // window.analytics.track(AnalyticsTrack.SignIn, payload);
+export function rudderanalyticsSignIn(uuid: string, email: string): void {
```
JoanMora: I would advise receiving the payload as an object so we are not bound to the order of the parameters, we already experimented with tracks received with these two parameters interchanged.

src/app/analytics/services/analytics.service.ts
xabg2: This function doesn't exists

src/app/auth/components/SignUp/SignUp.tsx
```diff
@@ -115,9 +116,8 @@ function SignUp(props: SignUpProps): JSX.Element {
         dispatch(referralsThunks.initializeThunk());
       }
 
-      window.rudderanalytics.identify(xUser.uuid, { email, uuid: xUser.uuid });
-      window.rudderanalytics.track('User Signup', { email });
-      
+      analyticsService.rudderanalyticsSignUp(email, xUser.uuid);
```
JoanMora: Advisable to send payload as an object and if it is required userId by the library (as it is a standard) pass it with this convention to the service.

sg-gs:
Should this still be merged @JoanMora ?

PR #493, opened 2022-11-14
[WT-191]: feat/recents-panel-tests

rafijv:
Added tests:
- List recents
- Get link


test/e2e/cypress/integration/recents-panel.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+      cy.visit('/app/recents');
+    });
+
+    it('Should list recents', () => {
+      cy.contains('example.txt');
+    });
+
+    it('Should get link', () => {
+      cy.get('button.file-list-item-actions-button').click();
+      cy.get('a').contains('Get link').click();
```
fabioespinosa: Here we want to make sure the link is generated, try to find out if there is a way in cypress to check if something has been added to clipboard and also that the pop up saying "share link copied to clipboard" appears

test/e2e/cypress/integration/recents-panel.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+      cy.get('button.file-list-item-actions-button').click();
+      cy.contains('Delete link');
```
fabioespinosa: Here we also want to make sure the link is deleted, perhaps checking for "link deleted" and the link logo no longer there could be a good idea

fabioespinosa (changes requested):
We want to make sure in these tests that the user can perform the tasks, and that they are successful

test/e2e/cypress/integration/recents-panel.spec.ts
rafijv: There is no task to remove the link, it is the way to check that the link has been generated but I will find another approach to do it.

test/e2e/cypress/integration/recents-panel.spec.ts
rafijv: There are ways to check the clipboard content but in the link that is generated I don't know how to find a value that I can compare.
The pop up disappears and there is no time to check the content.
Another option could be to check the shared links section and see if the shared file is listed, is that ok?

test/e2e/cypress/integration/recents-panel.spec.ts
rafijv: comment added

fabioespinosa (approved):

PR #499, opened 2022-11-15
[WT-187] feat/account-panel-tests

rafijv:
Add e2e tests
- Invite friend
- Change first, last name

### DIFF
--- a/src/app/shared/components/Input/index.tsx
+++ b/src/app/shared/components/Input/index.tsx
@@ -15,6 +15,7 @@ export default function Input({
   onFocus,
   onBlur,
   autofocus = false,
+  name,
 }: {
   className?: string;
   label?: string;
@@ -29,6 +30,7 @@ export default function Input({
   onBlur?: () => void;
   message?: string;
   autofocus?: boolean;
+  name?: string;
 }): JSX.Element {
   const inputRef = useRef<HTMLInputElement>(null);
 
@@ -95,6 +97,7 @@ export default function Input({
           setIsFocused(false);
         }}
         value={value}
+        name={name}
       />
       {variant === 'password' && isFocused && (
         <div
--- a/test/e2e/cypress/integration/account-panel.spec.ts
+++ b/test/e2e/cypress/integration/account-panel.spec.ts
@@ -1,6 +1,7 @@
 describe('Account panel', () => {
   const firstName = 'Name';
   const lastName = 'Lastname';
+  const email = 'name@intx.com';
 
     beforeEach(() => {
       cy.clearLocalStorage();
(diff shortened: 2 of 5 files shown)


test/e2e/cypress/integration/account-panel.spec.ts
```diff
@@ -0,0 +1,28 @@
+describe('Account panel', () => {
+  const firstName = 'Name';
+  const lastName = 'Lastname';
+  const email = 'name@intx.com';
```
fabioespinosa: its good to keep this practice of using @inxt.com theres a typo there though

src/app/shared/components/Input/index.tsx
```diff
@@ -29,6 +30,7 @@ export default function Input({
   onBlur?: () => void;
   message?: string;
   autofocus?: boolean;
+  name?: string;
```
fabioespinosa, line 33: nice addition to the Input component

src/app/shared/components/Input/index.tsx
```diff
@@ -95,6 +97,7 @@ export default function Input({
           setIsFocused(false);
         }}
         value={value}
+        name={name}
```
fabioespinosa, line 100: i'm just curious if `name` is undefined will it render name="undefined" in the html 🤔  or it won't render name at all. 

If it is the former then there is no problem with this approach.

test/e2e/cypress/integration/account-panel.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      cy.login();
+      cy.visit('/preferences?tab=account');
+    });
+
+    it('Should edit name and lastname', () => {
+      cy.get('button').contains('Edit').click();
+      cy.get('input[name=firstName]').type(firstName);
+      cy.get('input[name=lastName]').type(lastName);
+      cy.get('button').contains('Save').click();
+      cy.contains(firstName);
```
fabioespinosa: Can you please check if this `contains` awaits the change? or does it check the already typed result.

test/e2e/cypress/integration/account-panel.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    });
+
+    it('Should invite friend', () => {
+      cy.get('input:first').should('have.attr', 'placeholder', 'Enter friend email').type(email);
+      cy.get('button').contains('Send invitation').click();
+      cy.get('button').contains('See all invitations').click();
+      cy.contains(email);
+    });
+
+});
```
fabioespinosa: Try to use prettier to always have a EOF line

fabioespinosa (changes requested):

src/app/shared/components/Input/index.tsx
rafijv: does not render name

test/e2e/cypress/integration/account-panel.spec.ts
rafijv: Yes, wait for the change

fabioespinosa (approved):

PR #501, opened 2022-11-16
[WT-210] bugfix/add-password-length-restriction-to-signup

CandelR:
- Added password length restriction in sign up form (20 characters) 

- 📸  Snapshots

image

### DIFF
--- a/src/app/auth/components/SignUp/SignUp.tsx
+++ b/src/app/auth/components/SignUp/SignUp.tsx
@@ -23,7 +23,7 @@ import testPasswordStrength from '@internxt/lib/dist/src/auth/testPasswordStreng
 import PasswordStrengthIndicator from 'app/shared/components/PasswordStrengthIndicator';
 import { useSignUp } from './useSignUp';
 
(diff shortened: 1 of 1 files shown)


src/app/auth/components/SignUp/SignUp.tsx
```diff
@@ -23,6 +23,8 @@ import testPasswordStrength from '@internxt/lib/dist/src/auth/testPasswordStreng
 import PasswordStrengthIndicator from 'app/shared/components/PasswordStrengthIndicator';
 import { useSignUp } from './useSignUp';
 
+const MAX_PASSWORD_LENGHT = 20;
```
fabioespinosa: typo here it should be length

fabioespinosa: MAX_PASSWORD_LENGTH

fabioespinosa (changes requested):
correct typo an it goes in 👌

fabioespinosa (approved):

PR #502, opened 2022-11-17
[WT-211] bugfix/Files with zero bytes fail to upload

CandelR:


src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -73,7 +84,12 @@ export const uploadItemsThunk = createAsyncThunk<void, UploadItemsPayload, { sta
 
     const storageClient = SdkFactory.getInstance().createStorageClient();
 
+    let zeroLengthFilesNumber = 0;
     for (const file of files) {
+      if (file.size === 0) {
+        zeroLengthFilesNumber = zeroLengthFilesNumber + 1;
+        continue;
```
fabioespinosa, line 91: I like this approach, but would be even cooler if we had a list of the names of the files that have 0 bytes in the notification so the client know which ones they are.

example: `readme.md, file_one.jpg and exam.pdf ` were not uploaded because they have 0 bytes

fabioespinosa (changes requested):

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
CandelR: Okay! I'll tell Joan from the product department, so he can give his approval, that it is he who has given me the text

fabioespinosa: Nice, do you have an update on this?

fabioespinosa: merging now, to be looked at in the future when our lead designer comes back

fabioespinosa (approved):

PR #504, opened 2022-11-18
[WT-188] feature/add-backup-view-tests

CandelR:
- added backup view cypress tests


test/e2e/cypress/integration/backups.spec.ts
```diff
@@ -0,0 +1,23 @@
+describe('Backups view', () => {
+  beforeEach(() => {
+    cy.clearLocalStorage();
+    cy.login();
+    cy.intercept('GET', '/api/backup/deviceAsFolder', { fixture: '/api/backups/deviceAsFolder.json' });
+    cy.intercept('GET', '/api/storage/v2/folder/130', { fixture: '/api/backups/storagev2folder130.json' });
+    cy.intercept('GET', '/api/storage/v2/folder/131', { fixture: '/api/backups/storagev2folder131.json' });
```
fabioespinosa, line 7: I like this approach, and given in drive-web you cannot create backups, its good enough to fulfill the task.

Although it occurs to me to test the functionality completely we might be able create some backups via pure HTTP requests and then test them without intercepting? 🤔 

(correct me if it's not possible)

fabioespinosa (changes requested):

test/e2e/cypress/integration/backups.spec.ts
CandelR: Yes, as long as the data we have after logging in to drive-web is enough to make the same calls that desktop makes to generate a backup. 
I think that @JoanVicens can surely confirm this for us.

JoanVicens: It's doable but I would rather (for the moment) add an e2e test to drive server to test his functionality. 
This will help us to achieve test independence from each other since to upload a backup, first we need to create a device, and to create a device we need a user. So we are going to rely on the auth test to have a user.

CandelR:
This test has already been added to this PR https://github.com/internxt/drive-web/pull/642

PR #506, opened 2022-11-22
[WT-215] bugfix/Backup deletion on Drive-web seems to not be working

CandelR:
- Solved deletion of backup items  (devices, folders and files) when its stored in storage


sg-gs (approved):

sg-gs:
Great job @CandelR

PR #507, opened 2022-11-23
[WT-219]: fix/taskLogger-refresh-when-file-is-uploaded-after-folder

rafijv:


sg-gs:
I am not sure if removing a configuration parameter and assuming that always a notification should be shown is a good idea, why not passing that parameter where the issue is? @rafijv @larry-internxt

sg-gs (approved):

PR #512, opened 2022-11-28
[WT-216] Fix: server key not stored on user creation

JoanVicens:
Contains the same changes as #509 plus handles the server not returning the keys. 
⚠️ To properly fix the issue with the password https://github.com/internxt/drive-server-wip/pull/120 changes need to be deployed first


JoanVicens:
Changed to draft, because noticed a possible error

sg-gs (approved):

PR #513, opened 2022-11-28
[WT-220]: feat/limit uploads to 1000 items

rafijv:


src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -458,31 +466,53 @@ declare module 'react' {
...
-    );
-  }
 
-  if (rootList.length) {
-    for (const root of rootList) {
+  let totalFilesToUpload: number;
+
+  totalFilesToUpload = files.length;
+
+  const countTotalItemsToUpload = (rootList) => {
```
fabioespinosa: rootList is not typed, it is of type `IRoot[]`

fabioespinosa: Mutating a variable outside the function -in this case `totalFilesToUpload`- is not the best practice.

Having a function that is self contained, can be a better practice. If we ever want to refactor and re-use this function in other context for IRoot types, we would be set to export it and use it right where needed.

Refactor this code block so that we only have:

`const countTotalItemsToUpload = files.length + countTotalItemsInIRoot(rootList)`

Where `countTotalItemsInIRoot` is a recursive function that returns the total number of items in an IRoot.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -458,31 +466,53 @@ declare module 'react' {
...
+  countTotalItemsToUpload(rootList);
+
+  if (totalFilesToUpload < 1000) {
+    if (files.length) {
+      // files where dragged directly
       await dispatch(
-        storageThunks.uploadFolderThunkNoCheck({
-          root,
-          currentFolderId,
+        storageThunks.uploadItemsThunkNoCheck({
```
fabioespinosa, line 496: Why are we changing to `uploadItemsThunkNoCheck` from `uploadFolderThunkNoCheck`. 

I see that `uploadItemsThunkNoCheck` was implemented by larry not so long ago, I think it's a good idea. However this PR only states that we are handling upload size limit, if you are doing any other change please state it in the description of the PR so that it is easier to review. In this case, please add why are we changing to `uploadItemsThunkNoCheck`

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -458,31 +466,53 @@ declare module 'react' {
...
           options: {
             onSuccess: onDragAndDropEnd,
           },
         }),
       );
     }
+    if (rootList.length) {
+      for (const root of rootList) {
+        await dispatch(
+          storageThunks.uploadFolderThunkNoCheck({
```
fabioespinosa, line 508: Same here, why is this change made and how is it related to 1000 items upload limit 🤔 

If it's not related, its ok, just add it to the description

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -458,31 +466,53 @@ declare module 'react' {
...
+            options: {
+              onSuccess: onDragAndDropEnd,
+            },
+          }),
+        );
+      }
+    }
+  } else {
+    notificationsService.show({
+      text: 'The maximum is 1000 files per upload.',
```
fabioespinosa, line 521: I think this warning is ok, although as an opinion, we might need a bigger warning sign.

Please check with Alvaro if we can have a prompt with a button of `ok` so that we are sure the user got the feedback >1000 files is too much. Instead of a vanishing one.

fabioespinosa (changes requested):
Changes in this PR solve the issue of > 1000 items warning successfully, but there are changes I am not sure why are they there and how are they -or if they are-  related to > 1000 items

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: Refactor done!

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: added dialog component

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: I have not replaced anything, I have only wrapped the block that already existed in an `if`
Captura de pantalla 2022-12-12 a las 9 31 54

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: I have not replaced anything, I have only wrapped the block that already existed in an `if`
Captura de pantalla 2022-12-12 a las 9 31 54

fabioespinosa:
Changing base to staging to check dialog

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: Captura de pantalla 2022-12-15 a las 17 35 49

fabioespinosa: nice

fabioespinosa (approved):

PR #514, opened 2022-11-30
[WT-233]: feat/modified behavior of selected items

rafijv:


src/app/drive/components/DriveExplorer/DriveExplorerItem/hooks/useDriveItemActions.tsx
```diff
@@ -179,9 +178,8 @@ const useDriveItemActions = (item: DriveItemData): DriveItemActions => {
   };
 
   const onItemClicked = (): void => {
-    isItemSelected(item)
-      ? dispatch(storageActions.deselectItems([item]))
-      : dispatch(storageActions.selectItems([item]));
+    dispatch(storageActions.clearSelectedItems());
+    dispatch(storageActions.selectItems([item]));
```
CandelR, line 182: With this change it might be necessary to add an enlargement of the click area of the checkbox, as this would make it difficult to select multiple items because the checkbox area is too small in the List view. 
Also, if I'm not mistaken, the possibility to select several items in the Grid view is lost.
Maybe we should look for an alternative to modify the behaviour when clicking on an item.
What do you think?

rafijv: @CandelR 
This is the behavior that Álavaro has required, the same as shared links.

CandelR: Yes yes, I understand that it meets the behaviour requested in the task. 
What I mean is that removing the multi-selection of items by clicking on the 3 dots or anywhere on each item in the list can make it difficult to select multiple items in the Checkbox, as it is very small. 
Also this change breaks the multiselection selection in grid view mode, with this change I it is only be able to select one item in the grid view mode.
By this I mean that maybe we should modify the way the `DriveExplorerListItem` component is structured to meet Alvaro's requirements instead of modifying the behaviour of the `onclick` in the item, so that when pressing on the three dots the `onItemClicked` function is not executed and the possibility to click elsewhere on the item to select it remains, also to not break in grid view the multi item selection.

rafijv: The things you are proposing are solved with the new design of Shared links but this is out of the scope of the issue that was a priority for November.
I started a conversation with Alvaro and Sergio yesterday, it was agreed to solve the Jira issue and try in a new task to equalize all the sections.
With this code I think we are closer to the new UX and solve the issue which is what was wanted.

CandelR (approved):

fabioespinosa:
Merging to staging to check behavior works

PR #515, opened 2022-11-30
[WT-235]: fix/hide link editing when selecting non-linked item

rafijv:


src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -113,6 +113,7 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
   const hasItems = items.length > 0;
   const hasFilters = storageFilters.text.length > 0;
   const hasAnyItemSelected = selectedItems.length > 0;
+  const isSelectedItemShared = selectedItems[0]?.shares?.length != 0;
```
fabioespinosa: Always use `!==` and `===`

fabioespinosa: We could also use >0 like the other two above, any reason to prefer !==0?

fabioespinosa: nevermind I see the typescript error: object is possibly undefined, I'll change it to `!==`

fabioespinosa (approved):

PR #530, opened 2022-12-21
[WT-238]: fix/ empty message when uploading a folder with name that already exist

rafijv:
I have added a validation to compare the uploaded folders with the already existing ones, in case they already exist, it will not make the call to the backend and the messages will be displayed.


CandelR (approved):

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -500,8 +508,9 @@ const uploadItems = async (props: DriveExplorerProps, rootList: IRoot[], files:
         }),
       );
     }
-    if (rootList.length) {
-      for (const root of rootList) {
+
+    if (foldersUploadedNotExist.length) {
```
sg-gs, line 512: why the length is needed to be checked?

sg-gs (changes requested):

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
rafijv: To check that it has elements, if there are no elements there is no need to execute the for

PR #535, opened 2022-12-22
[_]: chore/Add github action for share

larry-internxt:
- Added cd to share action


fabioespinosa:
https://github.com/internxt/drive-web/pull/537

PR #551, opened 2023-01-05
Feature/add lifetimes

aboredvaro:


src/app/core/views/Preferences/tabs/Plans/PlanSelector.tsx
```diff
@@ -23,7 +23,7 @@ export default function PlanSelector({ className = '' }: { className?: string })
   const priceButtons = subscription?.type === 'subscription' ? 'change' : 'upgrade';
 
   const [prices, setPrices] = useState<DisplayPrice[] | null>(null);
-  const [interval, setInterval] = useState<'month' | 'year'>('year');
+  const [interval, setInterval] = useState<'month' | 'year' | 'lifetime'>('year');
```
CandelR: Maybe would be good to begin to extract the repeated string values to constants. 

```
export const PAYMENT_INTERVAL = {
  MONTH: 'month' as const,
  YEAR: 'year' as const,
  LIFETIME: 'lifetime' as const,
};
```
If want to add the type also could add this:
`type PAYMENT_INTERVAL = typeof PAYMENT_INTERVAL[keyof typeof PAYMENT_INTERVAL];`

and use it in the useState:
`  const [interval, setInterval] = useState<PAYMENT_INTERVAL>(PAYMENT_INTERVAL.YEAR);`

aboredvaro: Since types comes from SDK I'll simply reuse the SDK Interface like this:
`const [interval, setInterval] = useState<DisplayPrice['interval']>('year');`

CandelR: okay 👍, but the main idea that what I propose is to avoid use magic strings. 
Avoiding magic strings we reach more solid code, avoiding possible typos, change the values more safety.
In this case it is only one file and it is not complicated, but if it is used in more places in the code it can get complicated.

CandelR (approved):

PR #557, opened 2023-01-13
[WT-288] feature/Use the browser database to store photos and files content

CandelR:
- Added new database object that stores the Blob of drive files and let to avoid redownload when the user wants to see the preview of the file that previously has seen


src/app/core/services/date.service.test.ts
```diff
@@ -0,0 +1,27 @@
+import moment from 'moment';
```
sg-gs: change `moment` for `dayjs` or other alternatives, `moment` is deprecated

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -53,65 +55,119 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show }: FileViewerP
   const isTeam = useAppSelector(sessionSelectors.isTeam);
   const userEmail: string = localStorageService.getUser()?.email || '';
 
+  const handleFileThumbnail = async (file: DriveFileData, fileBlob: Blob) => {
```
sg-gs: [File implements Blob](https://developer.mozilla.org/en-US/docs/Web/API/File), therefore you should be able to do the implicit coercion by typing the `fileBlob` variable as `File`. Also, as a suggestion, you can name it `file` then, instead of `fileBlob`

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -53,65 +55,119 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show }: FileViewerP
...
+      size: file.size,
+      type: file.type,
+      content: fileObject,
+      parentFolderId: file.folderId,
+    };
+
+    const thumbnailGenerated = await getThumbnailFrom(fileUpload);
+
+    if (
+      thumbnailGenerated &&
```
sg-gs: Please extract these comparisons in a boolean variable whose name explains why is the check being done, in two months we will be very grateful for doing that

sg-gs: Some questions that could arise and can help you to choose a naming for the variables:
- Why do I check `.file` property nullability?
- Why do I check `.type` property nullability?

sg-gs (changes requested):

src/app/drive/components/FileViewer/FileViewer.tsx
CandelR: Finally here I have removed the unnecessary comparisons leaving only the thumbnailGenerated.file. 
It was not possible to extract it into an explicitly named boolean because typescript cannot infer the type inside the conditional.

sg-gs (approved):

PR #563, opened 2023-01-24
[_](feature) Sort photos by takenAt, use networkBucketId

PixoDev:
- Display Photos sorted by takenAt date in descending order
- Use networkBucketId instead of the photos bucketId if the photo contains it

### DIFF
--- a/src/app/photos/components/Preview.tsx
+++ b/src/app/photos/components/Preview.tsx
@@ -8,6 +8,7 @@ import { photosSlice, PhotosState } from '../../store/slices/photos';
 import useIdle from '../../core/hooks/useIdle';
 import { PhotosItemType } from '@internxt/sdk/dist/photos';
 import * as Sentry from '@sentry/react';
+
 export default function Preview({
   onDownloadClick,
   onDeleteClick,
@@ -35,6 +36,7 @@ export default function Preview({
     const photo = items[previewIndex];
     const photoBucketId = photo.networkBucketId ? photo.networkBucketId : bucketId;
     if (!photoBucketId) return;
+
     getPhotoPreview({
       photo,
       bucketId: photoBucketId,
--- a/src/app/photos/components/Preview.tsx
+++ b/src/app/photos/components/Preview.tsx
@@ -69,15 +69,16 @@ export default function Preview({
         return setSrc(URL.createObjectURL(blob));
       })
       .catch((err) => {
+        if (abortController.signal.aborted) {
+          return;
+        }
+
         Sentry.captureException(err, {
           extra: {
             photoId: photo.id,
             bucketId: photoBucketId,
           },
         });
-        if (abortController.signal.aborted) {
-          return;
-        }
       });
 
     return () => {
--- a/src/app/photos/components/Preview.tsx
+++ b/src/app/photos/components/Preview.tsx
@@ -7,7 +7,7 @@ import { RootState } from '../../store';
 import { photosSlice, PhotosState } from '../../store/slices/photos';
 import useIdle from '../../core/hooks/useIdle';
 import { PhotosItemType } from '@internxt/sdk/dist/photos';
-
+import * as Sentry from '@sentry/react';
 export default function Preview({
   onDownloadClick,
   onDeleteClick,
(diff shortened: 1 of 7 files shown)


src/app/photos/components/Preview.tsx
```diff
@@ -30,43 +30,59 @@ export default function Preview({
...
+    const photoBucketId = photo.networkBucketId ? photo.networkBucketId : bucketId;
+    if (!photoBucketId) return;
+
+    getPhotoBlob({ photo, bucketId: photoBucketId, abortController })
+      .then((blob) => {
+        setItemType(photo.itemType);
+        return setSrc(URL.createObjectURL(blob));
+      })
+      .catch((err) => {
+        Sentry.captureException(err, {
```
sg-gs, line 78: avoid capturing aborted downloads moving the if before capturing exceptions on Sentry

src/app/photos/views/PhotosView.tsx
```diff
@@ -234,11 +233,21 @@ function PhotoItem({
...
+    if (photoBucketId) {
       getPhotoPreview({
         photo,
-        bucketId,
-      }).then(setSrc);
+        bucketId: photoBucketId,
+      })
+        .then(setSrc)
+        .catch((err) => {
+          Sentry.captureException(err, {
```
sg-gs, line 245: Same here, what if the download is aborted?

sg-gs (changes requested):

src/app/photos/views/PhotosView.tsx
PixoDev: this operation is not abortable, a user cannot click in the UI and abort the download operation, this operation runs one for each preview download

src/app/photos/components/Preview.tsx
```diff
@@ -7,7 +7,7 @@ import { RootState } from '../../store';
 import { photosSlice, PhotosState } from '../../store/slices/photos';
 import useIdle from '../../core/hooks/useIdle';
 import { PhotosItemType } from '@internxt/sdk/dist/photos';
-
+import * as Sentry from '@sentry/react';
 export default function Preview({
```
CandelR: nit: add a break line between imports and the funcions.

src/app/photos/components/Preview.tsx
```diff
@@ -30,43 +30,60 @@ export default function Preview({
...
-        photo,
-        bucketId,
-      }).then(setThumbnailSrc);
-    }
+    if (previewIndex === null) return;
+    setThumbnailSrc(null);
+    const photo = items[previewIndex];
+    const photoBucketId = photo.networkBucketId ? photo.networkBucketId : bucketId;
+    if (!photoBucketId) return;
+    getPhotoPreview({
```
CandelR: nit: also here, a break line before the if improve the readability

src/app/photos/views/PhotosView.tsx
```diff
@@ -13,7 +13,7 @@ import Preview from '../components/Preview';
 import ShareDialog from '../components/ShareDialog';
 import Skeleton from '../components/Skeleton';
 import Toolbar from '../components/Toolbar';
-
+import * as Sentry from '@sentry/react';
 export default function PhotosView({ className = '' }: { className?: string }): JSX.Element {
```
CandelR: nit: also here, add a break line

src/app/photos/components/Preview.tsx
```diff
@@ -30,43 +30,60 @@ export default function Preview({
...
-          setItemType(photo.itemType);
-          return setSrc(URL.createObjectURL(blob));
-        })
-        .catch((err) => {
-          if (abortController.signal.aborted) {
-            return;
-          }
+    const abortController = new AbortController();
+    const photo = items[previewIndex];
+    const photoBucketId = photo.networkBucketId ? photo.networkBucketId : bucketId;
```
CandelR, line 65: Two questions about this, in which case the photos not gonna has `networkBucketId` and it has `bucketId`?
Photos uploaded before implement `networkBucketId`?

And in which case the photo not gonna has one of both?

CandelR (changes requested):

src/app/photos/components/Preview.tsx
PixoDev: - networkBucketId could be undefined for Photos created before we introduced the property in photos-server
- the bucketId is a property that comes from Redux, so it can be undefined if is not initialized yet, but should exist always

Should not happen the case where both doesn't exists, at least the bucketId should has value

CandelR (approved):

sg-gs (approved):

PR #575, opened 2023-02-01
[WT-313]:(feature) Download multiple items as zip file

PixoDev:
Allow to download multiple selected files as a single zip with the files inside, including folders.

Tested on:

- Safari
- Brave
- Microsoft Edge
- Chrome
- Firefox

### DIFF
--- a/src/app/core/services/socket.service.ts
+++ b/src/app/core/services/socket.service.ts
@@ -22,7 +22,8 @@ export default class RealtimeService {
       auth: {
         token: getToken(),
       },
-      withCredentials: true
+      reconnection: isProduction(),
+      withCredentials: true,
     });
 
     this.socket.on('connect', () => {
@@ -40,8 +41,7 @@ export default class RealtimeService {
     });
 
     this.socket.on('connect_error', (error) => {
-      if (!isProduction())
-        console.error('[REALTIME] CONNECTION ERROR:', error);
+      if (!isProduction()) console.error('[REALTIME] CONNECTION ERROR:', error);
     });
   }
 
--- a/src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
+++ b/src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
@@ -1,70 +1,134 @@
 import { ActionReducerMapBuilder, createAsyncThunk } from '@reduxjs/toolkit';
-
-import storageThunks from '.';
 import { StorageState } from '../storage.model';
 import { RootState } from '../../..';
-import { DriveFileData, DriveFolderData, DriveItemData } from 'app/drive/types';
 import i18n from 'app/i18n/services/i18n.service';
 import notificationsService, { ToastType } from 'app/notifications/services/notifications.service';
-import { DownloadFileTask, DownloadFolderTask, TaskType } from 'app/tasks/types';
+import date from 'app/core/services/date.service';
+import { FlatFolderZip } from 'app/core/services/zip.service';
+import { downloadFile } from 'app/network/download';
+import { DriveItemData } from 'app/drive/types';
+import localStorageService from 'app/core/services/local-storage.service';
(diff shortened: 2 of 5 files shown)


src/app/core/services/socket.service.ts
```diff
@@ -22,7 +22,8 @@ export default class RealtimeService {
       auth: {
         token: getToken(),
       },
-      withCredentials: true
+      reconnection: isProduction(),
```
PixoDev, line 25: Stop the reconnection error spam on dev mode

CandelR: Nice 🚀

src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
```diff
@@ -1,70 +1,134 @@
...
-            options: { taskId },
-          }),
-        );
-      } else {
-        await dispatch(
-          storageThunks.downloadFileThunk({
-            file: item as DriveFileData,
-            options: { taskId },
-          }),
-        );
```
CandelR: Need to restore this because now is downloading every item as a zip

CandelR (changes requested):

src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
PixoDev: restored here: https://github.com/internxt/drive-web/pull/575/commits/152f6d3acbe854a2cc76dd35905e1c5edf1d16e6

src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
```diff
@@ -69,6 +79,126 @@ export const downloadItemsThunk = createAsyncThunk<void, DriveItemData[], { stat
...
+                const progress = downloadedBytes / totalBytes;
+
+                downloadProgress[index] = progress;
+
+                updateProgressCallback(calculateProgress());
+              },
+            },
+          });
+
+          folder.addFile(`${driveItem.name}.${driveItem.type}`, fileStream);
```
CandelR: We have implemented a cache system. If we add it here the speed of the download will be increased. 
Here's a suggestion, I think this will work well.

```suggestion
          const lruFilesCacheManager = await LRUFilesCacheManager.getInstance();
          const cachedFile = await lruFilesCacheManager.get(driveItem.id.toString());
          const isCachedFileOlder = checkIfCachedSourceIsOlder({ cachedFile, file: driveItem });

          let fileStream;
          if (cachedFile?.source && !isCachedFileOlder) {
            const blob = new Blob([new Uint8Array([1, 2, 3])], { type: 'application/octet-stream' });
            fileStream = blob.stream();
            updateProgressCallback(calculateProgress());
          } else {
            fileStream = await downloadFile({
              fileId: driveItem.fileId,
              bucketId: driveItem.bucket,
              creds: {
                user: user.bridgeUser,
                pass: user.userId,
              },
              mnemonic: user.mnemonic,
              options: {
                abortController,
                notifyProgress: (totalBytes, downloadedBytes) => {
                  const progress = downloadedBytes / totalBytes;

                  downloadProgress[index] = progress;

                  updateProgressCallback(calculateProgress());
                },
              },
            });
          }

          folder.addFile(`${driveItem.name}.${driveItem.type}`, fileStream);
```

These two lines:
```
const lruFilesCacheManager = await LRUFilesCacheManager.getInstance();
const cachedFile = await lruFilesCacheManager.get(driveItem.id.toString());
```

could be changes for this one:
`  const cachedFile = await getDatabaseFileSourceData({ fileId: file.id });`

when [this PR](https://github.com/internxt/drive-web/pull/574) has been merged.

CandelR (changes requested):

CandelR (approved):

PR #585, opened 2023-02-13
[WT-286]: feat/add lazy loading to photo library

rafijv:
Improved user experience by allowing scrolling while the rest of the images load.

### DIFF
--- a/src/app/photos/views/PhotosView.tsx
+++ b/src/app/photos/views/PhotosView.tsx
@@ -99,6 +99,7 @@ export default function PhotosView({ className = '' }: { className?: string }): 
               <Grid
                 selected={photosState.selectedItems}
                 photos={photosState.items}
+                isMorePhotos={photosState.thereIsMore}
                 onUserScrolledToTheEnd={fetchPhotos}
               />
             )}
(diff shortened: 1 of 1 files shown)


CandelR (approved):
@aboredvaro check if the skeleton when is loading more photos accomplish the design requirements

aboredvaro:
> 

Loading skeleton is good, but the problem is still showing the full sized photos (previews) instead of the photo/video smaller thumbnails. Load times will improve dramatically by showing thumbnails. I've attached a video to show how laggy it can get with only a few photos (same performance both on dev and prod environments):


https://user-images.githubusercontent.com/34165099/218747244-2958ad0b-e080-4f85-8007-4200a3b5126f.mov

aboredvaro (changes requested):
Loading skeleton is good, but the problem is still showing the full sized photos (previews) instead of the photo/video smaller thumbnails. Load times will improve dramatically by showing thumbnails instead of previews (smaller image sizes). I've attached a video to show how laggy it can get with only a few photos (same performance both on dev and prod environments):

https://user-images.githubusercontent.com/34165099/218747244-2958ad0b-e080-4f85-8007-4200a3b5126f.mov

aboredvaro (approved):

PR #596, opened 2023-02-16
[WT-326]: Feat/switching between files in preview file view

xabg2:
User can navigate between files within file preview view.

### DIFF
--- a/src/app/drive/components/FileViewer/FileViewer.tsx
+++ b/src/app/drive/components/FileViewer/FileViewer.tsx
@@ -1,4 +1,4 @@
-import { Suspense, Fragment, useState, useEffect } from 'react';
+import { Suspense, Fragment, useState, useEffect, useMemo } from 'react';
 import { Dialog, Transition } from '@headlessui/react';
 import fileExtensionService from '../../services/file-extension.service';
 import viewers from './viewers';
@@ -29,6 +29,7 @@ import {
 import { FileExtensionGroup, fileExtensionPreviewableGroups } from 'app/drive/types/file-types';
 import iconService from 'app/drive/services/icon.service';
 import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
+import { CaretLeft, CaretRight } from 'phosphor-react';
 
 interface FileViewerProps {
   file?: DriveFileData;
@@ -42,7 +43,7 @@ interface FileViewerProps {
 
 export interface FormatFileViewerProps {
   blob: Blob;
-  getFile: (direction: number) => void;
+  changeFile: (direction: string) => void;
 }
 
 const extensionsList = fileExtensionService.computeExtensionsLists(fileExtensionPreviewableGroups);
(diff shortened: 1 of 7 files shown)


src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -191,7 +224,25 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show, progress }: F
     } else if (!show) {
       setBlob(null);
     }
-  }, [show]);
+  }, [show, file]);
+
+  const DownoladButton = ({ background }: { background?: string }) => (
```
CandelR: typo: `DownoladButton` → `DownloadButton`

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -224,22 +275,28 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show, progress }: F
...
+                >
+                  <CaretLeft size={24} />
+                </button>
                 {blob ? (
                   <Suspense fallback={<div></div>}>
-                    <Viewer blob={blob} />
+                    <Viewer blob={blob} changeFile={changeFile} />
                   </Suspense>
-                ) : (
+                ) : progress ? (
```
CandelR: With this, when the progress is 0 the Download button will be displayed instead of the progress bar, if the component has to be displayed when the progress is less than 1 it would be better to change to progress < 1. To avoid flashing of the download button

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -66,6 +77,28 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show, progress }: F
...
+  function changeFile(direction: 'next' | 'prev') {
+    if (direction === 'next') {
+      if (fileIndex === totalIndex - 1) {
+        setCurrentFile?.(folderFiles[0]);
+      } else {
+        setCurrentFile?.(folderFiles[fileIndex + 1]);
+      }
+    } else {
+      if (fileIndex === 0) {
+        setCurrentFile?.(folderFiles[totalIndex.length - 1]);
```
CandelR: this should not work `totalIndex.length`. Is `totalIndex`

CandelR (changes requested):

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -191,7 +224,25 @@ const FileViewer = ({ file, onClose, onDownload, downloader, show, progress }: F
     } else if (!show) {
       setBlob(null);
     }
-  }, [show]);
+  }, [show, file]);
+
+  const DownoladButton = ({ background }: { background?: string }) => (
+    <div
+      className={`${
+        background && background
```
CandelR: This is the same as use only `background`.

CandelR (approved):

PR #605, opened 2023-02-21
Analytics key events client side tracking

JoanMora:
Adds visibility for upload and download file flows. 
Covers the start, completed, aborted and error states of downloads and uploads for files. 

Provides interfaces using the tracking plan.

Note that the following env variables need to be added, and the `.env.example` is updated accordingly.
```
# ANALYTICS KEYS
REACT_APP_ANALYTICS_ERROR_REPORTING_WRITE_KEY=
REACT_APP_CDP_DATA_PLANE=
````

### CODE REVIEW

File: src/app/drive/services/download.service/downloadFile.ts [hunk 1]
```diff
@@ -135,15 +124,19 @@ export default async function downloadFile(
   await downloadToFs(completeFilename, fileStreamPromise, support, isFirefox, abortController).catch((err) => {
     const errMessage = err instanceof Error ? err.message : (err as string);
 
-    trackFileDownloadError(userEmail, fileId, errMessage);
+    if (errMessage.includes('user aborted')) {
```
sg-gs: To ensure that the abort always is caught, check use:
` if (abortController.signal.aborted) {...}`

src/app/drive/services/download.service/downloadFile.ts
```diff
@@ -125,7 +105,16 @@ export default async function downloadFile(
     support = DownloadSupport.PatchedStreamApi;
   }
 
-  trackFileDownloadStart(userEmail, fileId, itemData.name, itemData.size, itemData.type, itemData.folderId);
+  const trackingDownloadProperties: TrackingPlan.DownloadProperties = {
+    file_download_id: analyticsService.getTrackingActionId(),
+    file_id: parseInt(fileId),
```
sg-gs: Is this the fileId from the network (string) or the autoincremental id (unsigned int)? Unless the interface is explicitly changed, I will recommend double-checking if this is a string or an int, parseInt could fail and all the `downloadFile` function will crash

sg-gs (changes requested):

src/app/drive/services/download.service/downloadFile.ts
JoanMora: Added

src/app/drive/services/download.service/downloadFile.ts
JoanMora: This is the network fileId. Changes added

src/app/drive/services/download.service/downloadFile.ts
```diff
@@ -125,7 +105,16 @@ export default async function downloadFile(
     support = DownloadSupport.PatchedStreamApi;
   }
 
-  trackFileDownloadStart(userEmail, fileId, itemData.name, itemData.size, itemData.type, itemData.folderId);
+  const trackingDownloadProperties: TrackingPlan.DownloadProperties = {
+    file_download_id: analyticsService.getTrackingActionId(),
+    file_id: typeof fileId === 'string' ? parseInt(fileId) : fileId,
```
sg-gs, line 110: Parsing a string with the hex format (like the fileId from the network) is going to give you 'NaN'

sg-gs (changes requested):

sg-gs:
Environment variables are set and ready.

sg-gs (approved):

CandelR (approved):

PR #614, opened 2023-02-27
[WT-330]: Feat/pc componentes iframe

xabg2:
Iframes created for PCComponentes campaign.
* Auth Iframe: It contains the two inputs (email/password) and the 'signup' button.
* Button Iframe: If both inputs are empty, it autoScrolls to the top of the page, otherwise it logs in.

In both cases, when the user registers, he is redirected to the checkout with the coupon and the 1 month free trial.


src/app/auth/views/Auth/AuthView.tsx
```diff
@@ -17,45 +17,57 @@ import { useEffect, useState } from 'react';
...
-
-    window.addEventListener('message', onRecieveMessage);
-
-    return () => {
-      window.removeEventListener('message', onRecieveMessage);
-    };
-  });
+  // useEffect(() => {
+  //   const onRecieveMessage = (e) => {
+  //     if (permitedDomains.includes(e.origin)) {
```
sg-gs: Remove these comments

PR #618, opened 2023-02-28
[WT-329] feat/add view to download app on mobile devices

rafijv:
A view has been added that forces users to download the app. This view is displayed once the user has logged in, so you must be able to register account and login.


src/app/drive/views/MobileView/MobileView.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <div>
+              <p className="text-base font-medium text-gray-100">{fullName}</p>
+              <p className="text-base text-gray-50">{props.user?.username}</p>
+            </div>
+          </div>
+          <button
+            onClick={onLogOutClicked}
+            className="cursor-pointer rounded-lg border border-gray-10 bg-white py-2.5 px-5 font-medium drop-shadow"
+          >
+            Log out
```
CandelR: Add text into translation json

CandelR (changes requested):
I have tried from Vercel and it shows a blank screen when logging in from a mobile phone. Have you tested that it works correctly?

src/app/drive/views/MobileView/MobileView.tsx
rafijv: I had tested it locally with ngrok and it worked, it is true that with Vercel it is not working, I will check it.

rafijv: It should be solved by now but I can't test it because vercel has failed.

sg-gs (approved):

sg-gs:
Waiting for @CandelR  regarding to Vercel @rafijv

CandelR (changes requested):
The responsive needs an update, the logout button is out of view, you have to scroll sideways to see it completely.
Screenshot_2023-03-01-17-00-42-40_e4424258c8b8649f6e67d283a50a2cbc

CandelR (approved):

PR #625, opened 2023-03-01
[_]: fix/Update PcComponentes design

xabg2:
Changes:
* Card text.
* Add form tag.
* Disable autocomplete for inputs

### DIFF
--- a/src/app/auth/components/PasswordInput/PasswordInput.tsx
+++ b/src/app/auth/components/PasswordInput/PasswordInput.tsx
@@ -21,6 +21,7 @@ interface InputProps {
   className?: string;
   autoFocus?: boolean;
   value?: string;
+  autoComplete?: string;
 }
 const PasswordInput = ({
   label,
@@ -37,6 +38,7 @@ const PasswordInput = ({
   onBlur,
   className,
   autoFocus,
+  autoComplete,
 }: InputProps): JSX.Element => {
   const [showPassword, setShowPassword] = useState(false);
   return (
@@ -48,7 +50,7 @@ const PasswordInput = ({
         min={0}
         required={true}
         autoFocus={autoFocus}
-        autoComplete="new-password"
+        autoComplete={autoComplete}
         {...register(label, {
           required,
           onChange: (e) => {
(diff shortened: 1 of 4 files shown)


src/app/auth/components/PasswordInput/PasswordInput.tsx
```diff
@@ -48,6 +48,7 @@ const PasswordInput = ({
         min={0}
         required={true}
         autoFocus={autoFocus}
+        autoComplete="new-password"
```
CandelR: avoid assign fixed props to generic components, this could cause wrong behaviours in different parts of the web

src/app/auth/components/TextInput/TextInput.tsx
```diff
@@ -40,6 +40,7 @@ export default function TextInput({
         type={type}
         disabled={disabled}
         placeholder={placeholder}
+        autoComplete="off"
```
CandelR: Same here, avoid assign fixed props to generic components, this could cause wrong behaviours in different parts of the web

src/app/auth/views/Auth/AuthView.tsx
```diff
@@ -29,101 +30,104 @@ const SignupAuth = () => {
...
+      <div className="flex w-full max-w-lg flex-col items-center space-y-2 pt-10 lg:w-max lg:items-start lg:pt-0">
+        <div className="flex w-full flex-col space-y-3 lg:flex-row lg:space-y-0 lg:space-x-3">
+          <div className="flex w-full">
+            <TextInput
+              placeholder={'Correo'}
+              label="email"
+              type="email"
+              className={'w-full'}
+              register={register}
+              minLength={{ value: 1, message: 'El correo no puede estar vacío' }}
```
CandelR, line 44: I assume there is no need to add translations because it is an element of a campaign with PCComponentes and it is only in Spanish.

src/app/auth/views/Auth/signup.ts
```diff
@@ -10,7 +10,7 @@ const postMessage = (data: Record<string, unknown>) => {
   window.top?.postMessage(data, CampaignLinks.PcComponentes);
 };
 
-const signup = async (data, dispatch, doRegister, setLoading, setError?) => {
+const signup = async (data, dispatch, doRegister?, setLoading?, setError?) => {
```
CandelR: Why are doRegister and setLoading optional, if their value is undefined won't it crash?

CandelR (changes requested):

CandelR (approved):

PR #632, opened 2023-03-08
[WT-280]: fix/move items dialog ux

rafijv:
Fixed some details that affected the UX when displaying the move item modal:

- Now it is avoided that when opening the modal the list of items is reloaded.
- Double clicking on the folders inside the modal navigates to that folder.
- Navigating to a folder in the modal does not reload the breadcrumb.
- Do not change the content of the 'move' button.
- Finally, the reloading of the folder list when closing the 'New folder' option has been avoided.

### DIFF
--- a/src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
+++ b/src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
@@ -15,7 +15,7 @@ import databaseService, { DatabaseCollection } from 'app/database/services/datab
 import CreateFolderDialog from '../CreateFolderDialog/CreateFolderDialog';
 import Breadcrumbs, { BreadcrumbItemData } from 'app/shared/components/Breadcrumbs/Breadcrumbs';
 import storageSelectors from 'app/store/slices/storage/storage.selectors';
-import { fetchFolderContentThunk } from 'app/store/slices/storage/storage.thunks/fetchFolderContentThunk';
+import { fetchDialogContentThunk } from 'app/store/slices/storage/storage.thunks/fetchDialogContentThunk';
 import Spinner from 'app/shared/components/Spinner/Spinner';
 import Button from 'app/shared/components/Button/Button';
 import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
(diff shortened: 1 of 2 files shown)


src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -81,10 +81,9 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
   }, [newFolderIsOpen]);
 
   const onShowFolderContentClicked = (folderId: number, name: string): void => {
-    dispatch(fetchFolderContentThunk(folderId))
+    dispatch(fetchDialogContentThunk(folderId))
       .unwrap()
       .then(() => {
-        setIsLoading(true);
```
CandelR, line 87: If you remove this line, won't the spinner no longer be displayed in the dialogue when the content is loading?
Maybe should be moved before the dispatch in order to make the loader appear when is fetching the content.

src/app/store/slices/storage/storage.thunks/fetchDialogContentThunk.ts
```diff
@@ -0,0 +1,52 @@
...
+        }),
+      );
+      databaseService.put(DatabaseCollection.Levels, folderId, items);
+    });
+  },
+);
+
+export const fetchDialogContentThunkExtraReducers = (builder: ActionReducerMapBuilder<StorageState>): void => {
+  builder.addCase(fetchDialogContentThunk.rejected, (state, action) => {
+    state.loadingFolders[action.meta.arg] = false;
```
CandelR: I think that set this loading to false is not necessary

rafijv: Removed

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
rafijv: With what you propose, everything continues to be reloaded with each action. I have moved it to useEffect so that it only reloads when the modal is opened.

CandelR (approved):
Approved but not merged in case @aboredvaro  has to review it

aboredvaro (approved):

PR #634, opened 2023-03-09
[WT-343]: Fix/Avatar load failing

larry-internxt:
Fixed issue where the user's avatar was being uploaded in a corrupted state


CandelR (approved):

sg-gs:
Why it was being corrupted and was being updated? Please provide a more detailed description @larry-internxt

larry-disabled:
> Why it was being corrupted and was being updated? Please provide a more detailed description @larry-internxt

I don't know exactly why they were getting corrupted, but what I know is that the continuous updates that were being made (for no reason) where part of the problem and that once the update was done, the avatar was then being downloaded corrupted forever..

So removing these updates has directly solved the problem, except for the avatars that are already corrupted by this bug, which will continue to be corrupted until manually updated

PR #638, opened 2023-03-20
[_]: feat/concurrent-uploads

sg-gs:


sg-gs:
Some feedback: 
- Upload progress when uploading a folder is not working
- We need to add this to uploading many files via drag & drop etc.

sg-gs:
Ready to go.

PR #639, opened 2023-03-20
[_]: Feat/blog sign up form

xabg2:
Iframe created to allow users to register from our blog.

### DIFF
--- a/src/app/auth/views/Auth/AuthView.tsx
+++ b/src/app/auth/views/Auth/AuthView.tsx
@@ -11,9 +11,9 @@ import signup from './signup';
 
 const textContent = {
   email: 'Correo',
-  password: 'Contraseña',
+  passwordLabel: 'Contraseña',
   emailEmpty: 'El correo no puede estar vacío',
-  passwordEmpty: 'La contraseña no puede estar vacía',
+  passwordLabelEmpty: 'La contraseña no puede estar vacía',
   buttonText: 'Obtén la oferta',
   legal: {
     line1: 'Al crear una cuenta aceptas',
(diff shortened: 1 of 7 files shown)


src/app/auth/views/Auth/SignupBlog.tsx
```diff
@@ -0,0 +1,131 @@
...
+  } = useForm<IFormValues>({
+    mode: 'onChange',
+    defaultValues: {
+      email: '',
+      password: '',
+    },
+  });
+
+  return (
+    <form onSubmit={handleSubmit((e) => signup(e, dispatch, doRegister, setLoading, true, setError))}>
```
CandelR: Is the only difference from the original SignupAuth the appRedirect parameter?
If so, you can reuse it by passing that parameter as a prop to avoid all that duplicate code.

CandelR (changes requested):

src/app/auth/views/Auth/AuthView.tsx
```diff
@@ -74,12 +94,9 @@ const SignupAuth = () => {
...
               type="submit"
               disabled={loading}
-              className={
-                'focus:outline-none shadow-xm relative flex h-11 w-full flex-row items-center justify-center space-x-4 whitespace-nowrap rounded-lg px-0 py-2.5 text-lg text-white transition duration-100 focus-visible:bg-orange-dark active:bg-orange-dark disabled:cursor-not-allowed disabled:text-white/75 sm:text-base'
-              }
-              style={{
-                backgroundColor: '#F26122',
-              }}
+              className={`focus:outline-none shadow-xm relative flex h-11 w-full flex-row items-center justify-center space-x-4 whitespace-nowrap rounded-lg ${
+                buttonColor || 'bg-pcComponentes-orange focus-visible:bg-orange-dark active:bg-orange-dark'
```
CandelR, line 98: Is the default colour bg-pcComponents-orange correct?

xabg2: Yes, because the modal is used for the pcComponents campaign mainly, so I recycled the registration component for this iframe.

CandelR (approved):

PR #640, opened 2023-03-20
[WT-344] - bugfix/Fixed percentage of bar logic

CandelR:
- Fixed space bar overflow in case the usage exceeds 100% of the maximum storage in the account tab of the preferences.


CandelR:
@aboredvaro Finally I have changed the logic so that in the case that a user has more than 100% of the space occupied because the available space is lowered or in the case that the user does not occupy more than 100% of the available space.

aboredvaro (approved):

sg-gs (approved):

PR #641, opened 2023-03-20
[WT-358] bugfix/Fix file preview option issues

CandelR:
- Fixed erroneous behaviour on actions of a shared file in preview

### DIFF
--- a/src/app/drive/components/FileViewer/components/TopBarActions.tsx
+++ b/src/app/drive/components/FileViewer/components/TopBarActions.tsx
@@ -19,22 +19,25 @@ import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
 import { useAppDispatch } from 'app/store/hooks';
 import { DriveItemData } from 'app/drive/types';
 import UilImport from '@iconscout/react-unicons/icons/uil-import';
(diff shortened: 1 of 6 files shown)


src/app/drive/components/FileViewer/components/TopBarActions.tsx
```diff
@@ -220,7 +239,19 @@ const TopBarActions: FC<TopBarActionsProps> = ({
...
           </Transition>
         </Menu>
-      ) : null}
+      )}
+      {!isAuthenticated && (
+        <Button
+          variant="secondary"
+          type="button"
+          onClick={() => {
+            window.location.href = process.env.REACT_APP_HOSTNAME + '/login';
```
sg-gs, line 248: why not use `navigationService.push('/login')`?

sg-gs (changes requested):

src/app/drive/components/FileViewer/components/TopBarActions.tsx
CandelR: Because I am not sure if it will work because it changes the domain and I have put the same navigation as there is ShareLayout.tsx to the login view in drive.

sg-gs (approved):

PR #643, opened 2023-03-21
[WT-348]: Feat/Add cancel subscription modal

larry-internxt:
- The cancel subscription button now shows a modal where users can give us details on why they are cancelling their subscription

### DIFF
--- a/src/app/drive/types/index.ts
+++ b/src/app/drive/types/index.ts
@@ -145,3 +145,8 @@ export enum DownloadFolderMethod {
   FileSystemAccessAPI = 'file-system-access-api',
   StreamSaver = 'stream-saver',
 }
+
+export enum FreeStoragePlan {
+  simpleName = '2GB',
+  storageLimit = 2147483648,
+}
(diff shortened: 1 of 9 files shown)


CandelR (approved):

src/app/drive/types/index.ts
```diff
@@ -145,3 +145,8 @@ export enum DownloadFolderMethod {
   FileSystemAccessAPI = 'file-system-access-api',
   StreamSaver = 'stream-saver',
 }
+
+export enum FreeStoragePlan {
+  simpleName = '2GB',
+  storageLimit = 2147483648,
```
sg-gs, line 151: Be careful, this could not be the case always, `payments.internxt.com` is the authority here and this app does in fact request plans from there

sg-gs (approved):

src/app/drive/types/index.ts
sg-gs: This could be problematic in the future, just to take it for reference.

PR #645, opened 2023-03-21
[WT-359]:(fix) Cancel folder upload correctly

PixoDev:
Cancel folder upload correctly, some patterns were causing troubles with this, see comments in code

### DIFF
--- a/src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
+++ b/src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
@@ -211,9 +211,10 @@ function getNextNewName(filename: string, i: number): string {
 
 export const uploadFolderThunkNoCheck = createAsyncThunk<void, UploadFolderThunkPayload, { state: RootState }>(
   'storage/createFolderStructure',
-  async ({ root, currentFolderId, options }, { dispatch, requestId }) => {
+  async ({ root, currentFolderId, options }, { dispatch }) => {
     options = Object.assign({ withNotification: true }, options || {});
 
+    const uploadFolderAbortController = new AbortController();
     let alreadyUploaded = 0;
     let rootFolderItem: DriveFolderData | undefined;
     const levels = [root];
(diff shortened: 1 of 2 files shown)


src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -224,7 +225,8 @@ export const uploadFolderThunkNoCheck = createAsyncThunk<void, UploadFolderThunk
       showNotification: !!options.withNotification,
       cancellable: true,
       stop: async () => {
-        const relatedTasks = tasksService.getTasks({ relatedTaskId: requestId });
+        uploadFolderAbortController.abort();
```
PixoDev, line 228: This abortController is the "parent" abortController, it aborts everything

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -604,7 +605,7 @@ export const uploadItemsParallelThunkNoCheck = createAsyncThunk<void, UploadItem
     const filesToUpload: FileToUpload[] = [];
     const errors: Error[] = [];
     const tasksIds: string[] = [];
-    const abortController = new AbortController();
+    const abortController = options?.abortController || new AbortController();
```
PixoDev, line 608: This allows to control this operation from another operation, the problem with the event driven cancellation is that if we are doing parallel operations, some of them can get in between the event emit and the actual operation cancel, is not a problem usually since the time is in ms, but for multiple operations it causes problems like this case, so now we can abort this from the parent directly if we pass this option

sg-gs: But aborting should throw an error on the main function which in turn should stop the entire execution. All the requests (as they have started using this controller), and all the functions too. Is not the case then?

PixoDev: abort doesn't throw an error, what throws an error is `abortController.signal.throwIfAborted()` by default an abort doesn't throw an error, it only flags the abortController as aborted. With this we are only notifying the "child" operations that the execution should stop via parent abortController.

Use case, we have an operation that creates multiple abortable operations, such this one. So now we use an abortController to abort all the operations, instead this was creating one abortController for each operation, so we would end with N abortControllers that would need to stop based on an event, with this we just pass down the abortController and abort from the parent.

sg-gs: Yeah, passing the abort controller is the best way. The issue here is that the abort controller, apart from using the signal to ensure any task that is going on if you want, should follow a procedural pattern where a reason is passed to the `abort` function as this will replicate any other error. Take into count that if you cancel a fetch via passing the signal, that fetch is going to fail, at least this is the recommended way to follow. We can go with this anyway.

sg-gs (approved):

PR #647, opened 2023-03-22
[WT-366] - bugfix/Files and folders reappearing after delete

CandelR:
- Added messages informing the user of the status during the moving items to Trash and the deletion of those items from Trash.
- Fixed behaviour that caused items to disappear from Drive if moving to Trash fails


sg-gs (approved):

sg-gs:
Just @aboredvaro approval and we will be ready to go

aboredvaro (approved):

PR #649, opened 2023-03-22
[WT-365]:(fix) Check if indexedDB is available for use in browser

PixoDev:
Now Drive Web works on Firefox Private Browsing mode


src/app/core/plugins/sentry.plugin.ts
```diff
@@ -9,7 +9,7 @@ const sentryPlugin: AppPlugin = {
       dsn: process.env.REACT_APP_SENTRY_DSN,
       integrations: [new BrowserTracing()],
       tracesSampleRate: 1.0,
-      debug: process.env.NODE_ENV !== 'production',
+      debug: process.env.NODE_ENV !== 'production' && process.env.REACT_APP_DEBUG == 'true',
```
sg-gs: strict equal could be better

sg-gs (commented):
I suggest handling the unavailability of Indexed DB as something internal to the cache, these changes are forcing us to change each method to be optional, which could not scale well if we add more and more places where the cache is being used. For example, if the cache can not cache things, the get, put, etc methods could return null or undefined, as they already do in these changes. Still, we could allow the `getInstance ` method to return the instance instead of returning null. 

This saves as many changes and helps the codebase to scale better, isn't it?

sg-gs:
wdyt @CandelR ?

src/app/core/plugins/sentry.plugin.ts
PixoDev: right

PixoDev:
> I suggest handling the unavailability of Indexed DB as something internal to the cache, these changes are forcing us to change each method to be optional, which could not scale well if we add more and more places where the cache is being used. For example, if the cache can not cache things, the get, put, etc methods could return null or undefined, as they already do in these changes. Still, we could allow the `getInstance ` method to return the instance instead of returning null.
> 
> This saves as many changes and helps the codebase to scale better, isn't it?

Note that we are checking if indexedDB is available, I added the check at the interface level, but what this checks is the availability of that provider.

If we provide an internal fallback like indexedDb not available => use localStorage we will need to deal with "what we can do that works with indexedDB and localStorage" since it has size restrictions, you can only store up to 5MB in localStorage.

About the pattern for checking it:

- We can throw an error if indexedDB is not available, this would force us to handle that error always making us wrapping this call in a try catch every time, less scalable in my opinion and makes the code flow a bit messy.

- Or we explicitily return null meaning that you cannot use that cache layer/implementation like I did, this will make the code flow less verbose with the optional chaining operator, and more explicit.

EDIT: I see what you mean, we can return the instance since the method will check for availablity too, so we can avoid the ? operator, will modify in the code

sg-gs (approved):

PR #654, opened 2023-03-28
[PB-147 WT-370]: bugfix/create-account-using-any-password

CandelR:
- Fixed not allowing registration with invalid password and redirection on login failure


CandelR:
⚠️ Before review this task it is necessary to merge the new end to end test https://github.com/internxt/drive-web/pull/642

sg-gs (approved):

PR #656, opened 2023-04-03
[WT-372]  feature/Retry action for failed downloads

CandelR:
- Allow the user to retry a failed/cancelled download of an item from the download information panel.

### DIFF
--- a/src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
+++ b/src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
@@ -34,7 +34,17 @@ export const downloadItemsThunk = createAsyncThunk<void, DownloadItemsThunkPaylo
 
     // * 1. Creates tasks
     for (const item of items) {
-      if (item?.taskId === undefined) {
+      const isRetryDownload = !!item.taskId;
+
+      if (isRetryDownload) {
+        tasksIds.push(item.taskId as string);
+        tasksService.updateTask({
+          taskId: item.taskId as string,
+          merge: {
+            status: TaskStatus.Decrypting,
+          },
+        });
+      } else {
         if (item.isFolder) {
           const taskId = tasksService.create<DownloadFolderTask>({
             action: TaskType.DownloadFolder,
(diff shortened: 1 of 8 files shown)


src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
```diff
@@ -32,27 +34,37 @@ export const downloadItemsThunk = createAsyncThunk<void, DriveItemData[], { stat
...
-      if (item.isFolder) {
-        const taskId = tasksService.create<DownloadFolderTask>({
-          action: TaskType.DownloadFolder,
-          relatedTaskId: taskGroupId,
-          folder: item,
-          compressionFormat: 'zip',
-          showNotification: true,
-          cancellable: true,
-        });
+      if (item?.taskId === undefined) {
```
sg-gs: why?

sg-gs: If `item?.taskId === undefined` means that is a download (which is a bit weird, btw), could be more understandable to use a variable for this:

`const taskIsADownload = item?.taskId === undefined;`

and then:

`if (taskIsADownload) {...}`

src/app/tasks/services/tasks.service/index.ts
```diff
@@ -83,6 +83,7 @@ class TaskManagerService {
     return {
       taskId: task.id,
       status: task.status,
+      item: task?.file || task?.folder,
```
sg-gs: why the task is maybe undefined if `task.id` and `task.status` are not?

sg-gs (changes requested):

src/app/store/slices/storage/storage.thunks/downloadItemsThunk.ts
CandelR: I added it at first because I thought the taskId was a number to avoid the problem of false true by negating a possible 0. Then I saw that the taskId is a string and I forgot to change this. 
I add as you say in a boolean variable that identifies the reason for the check because if not it is very likely that if you don't have context you won't understand the reason for the check.

src/app/tasks/services/tasks.service/index.ts
CandelR: It has no sense. I remove it

sg-gs (approved):

PR #657, opened 2023-04-04
[_]: fix/FR lang spell check

xabg2:
Add text and spell check for french lang.

### DIFF
--- a/src/app/core/views/Preferences/tabs/Account/Language.tsx
+++ b/src/app/core/views/Preferences/tabs/Account/Language.tsx
@@ -10,22 +10,31 @@ import es from 'dayjs/locale/es';
 import fr from 'dayjs/locale/fr';
 import dayjs from 'dayjs';
 
-function getLanguage(): string {
-  const deviceLanguage = localStorageService.get('language') as string;
-  return deviceLanguage;
-}
-
-const deviceLang = getLanguage();
+const currentLang = {
+  es: 'Español (ES)',
+  fr: 'Français (FR)',
+  en: 'English (EN)',
+};
 
 export default function Language(): JSX.Element {
   const { translate } = useTranslationContext();
-  const [lang, setLang] = React.useState<string>(deviceLang === 'es' ? 'es' : 'en');
-  const [currentLangText, setCurrentLangText] = React.useState<DefaultTFuncReturn>(
-    deviceLang === 'es' ? 'Español (ES)' : 'English (US)',
-  );
+  const [lang, setLang] = React.useState<string>();
+  const [currentLangText, setCurrentLangText] = React.useState<DefaultTFuncReturn>();
 
   useEffect(() => {
-    localStorageService.set('language', lang);
+    if (localStorageService.get('language')) {
+      setCurrentLangText(currentLang[localStorageService.get('language') as string]);
+      setLang(localStorageService.get('language') as string);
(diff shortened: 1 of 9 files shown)


src/app/i18n/services/i18n.service.ts
```diff
@@ -3,11 +3,13 @@ import i18next from 'i18next';
 import localStorageService from 'app/core/services/local-storage.service';
 import envService from 'app/core/services/env.service';
 
-let deviceLang = navigator.language.split('-')[0];
+export const currentLang = {
+  es: 'es',
+  fr: 'fr',
+  en: 'en',
+};
```
CandelR: I think this is not being used anywhere, if so it is better to delete it to avoid having code that is not going to be used

src/app/core/views/Preferences/tabs/Account/Language.tsx
```diff
@@ -10,22 +10,31 @@ import es from 'dayjs/locale/es';
...
+  useEffect(() => {
+    if (localStorageService.get('language')) {
+      setCurrentLangText(currentLang[localStorageService.get('language') as string]);
+      setLang(localStorageService.get('language') as string);
+    } else {
+      setCurrentLangText(currentLang[i18next.language]);
+      setLang(i18next.language);
+    }
+    localStorageService.set('language', lang as string);
+    console.log('currentLang', currentLangText, 'deviceLang', i18next.language);
```
CandelR: forgot to remove the console.log

src/app/core/views/Preferences/tabs/Account/Language.tsx
```diff
@@ -10,22 +10,31 @@ import es from 'dayjs/locale/es';
...
   const { translate } = useTranslationContext();
-  const [lang, setLang] = React.useState<string>(deviceLang === 'es' ? 'es' : 'en');
-  const [currentLangText, setCurrentLangText] = React.useState<DefaultTFuncReturn>(
-    deviceLang === 'es' ? 'Español (ES)' : 'English (US)',
-  );
+  const [lang, setLang] = React.useState<string>();
+  const [currentLangText, setCurrentLangText] = React.useState<DefaultTFuncReturn>();
+
+  useEffect(() => {
+    if (localStorageService.get('language')) {
```
CandelR: As you are accessing to the language of local storage in different points of the useEffect you could add it into a constant to avoid acces more than once.
for example: 
const localStorageLanguage = localStorageService.get('language');

src/app/core/views/Preferences/tabs/Account/Language.tsx
```diff
@@ -10,22 +10,31 @@ import es from 'dayjs/locale/es';
...
+      setCurrentLangText(currentLang[localStorageService.get('language') as string]);
+      setLang(localStorageService.get('language') as string);
+    } else {
+      setCurrentLangText(currentLang[i18next.language]);
+      setLang(i18next.language);
+    }
+    localStorageService.set('language', lang as string);
```
CandelR: This line is not necessary, at this point you are adding the `lang` value of current render, that would be undefined. 
With the other useEffect that it is executed when `lang` state changes it is adding the new lang value to the local storage.

CandelR (changes requested):

CandelR (approved):

PR #659, opened 2023-04-05
[WT-373] feature/update the cancellation form to only have text area input

rafijv:
Updated the unsubscribe form to have only the feedback field

### DIFF
--- a/src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
+++ b/src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
@@ -92,7 +92,7 @@ export default function CurrentPlanExtended({ className = '' }: { className?: st
 
   useEffect(() => {
     dispatch(planThunks.initializeThunk()).unwrap();
-  }, [userSubscription]);
+  }, [cancellingSubscription]);
 
   return (
     <Section className={className} title={translate('views.account.tabs.billing.currentPlan')}>
--- a/src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
+++ b/src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
@@ -45,7 +45,6 @@ export default function CurrentPlanExtended({ className = '' }: { className?: st
     setCancellingSubscription(true);
     try {
       await paymentService.cancelSubscription();
-      await dispatch(planThunks.initializeThunk()).unwrap();
       notificationsService.show({ text: translate('notificationMessages.successCancelSubscription') });
       setIsCancelSubscriptionModalOpen(false);
       trackCanceledSubscription({ feedback });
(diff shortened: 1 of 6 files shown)


sg-gs:
Hey @rafijv, seems good to me. But I have seen that when canceling, the space is updated, but the plan type does not. See the picture below. (I have a free plan after canceling, but I am still seeing the old one). Could you give it a look, please?

Captura de pantalla 2023-04-05 a las 17 50 10

sg-gs (approved):

rafijv:
> Hey @rafijv, seems good to me. But I have seen that when canceling, the space is updated, but the plan type does not. See the picture below. (I have a free plan after canceling, but I am still seeing the old one). Could you give it a look, please?
> 
> Captura de pantalla 2023-04-05 a las 17 50 10

@sg-gs @CandelR 
Right, it doesn't change until I refresh the page, it's a status problem but I don't know how to reproduce it locally, is there any way to change the plan so I can cancel it later?

CandelR:
@rafijv After testing the change in the Preview, it remains in an infinite loop asking for the Plan data after entering Settings.  I have tried with a free account.

src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
```diff
@@ -90,6 +90,10 @@ export default function CurrentPlanExtended({ className = '' }: { className?: st
     return plan.usageDetails?.total || -1;
   };
 
+  useEffect(() => {
+    dispatch(planThunks.initializeThunk()).unwrap();
+  }, [userSubscription]);
```
CandelR: It seems to cause an infinite loop

CandelR (changes requested):

src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
rafijv: It happens with the free plan, I have already changed it

src/app/core/views/Preferences/tabs/Billing/CurrentPlanExtended.tsx
```diff
@@ -90,6 +90,10 @@ export default function CurrentPlanExtended({ className = '' }: { className?: st
     return plan.usageDetails?.total || -1;
   };
 
+  useEffect(() => {
+    dispatch(planThunks.initializeThunk()).unwrap();
+  }, [cancellingSubscription]);
```
CandelR: This will cause it to run several times during the cancellation process, when set to true and false. Also on line 48 this thunk is also called during cancellation in the function that cancels the subscription.
If it only has to be executed when the subscription is cancelled, wouldn't it be better to execute this thunk at the end of the cancellation function?

rafijv: I've tested it a couple of times with the 1 second setTimeout and it works.

CandelR (approved):

PR #664, opened 2023-04-11
[_]: Feat/banner for lifetime campaign

xabg2:
* New banner for the lifetime campaign. 
* Fixed a small problem with the language box.

### DIFF
--- a/src/app/banners/Banner.tsx
+++ b/src/app/banners/Banner.tsx
@@ -1,11 +1,10 @@
 import React, { useEffect } from 'react';
 import { X } from 'phosphor-react';
-import Hearts from 'assets/images/banner/hearts.png';
-import NeonBlur from 'assets/images/banner/neonBlur.png';
+import NeonBlur from 'assets/images/banner/infinity.svg';
 import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
 
 const Banner = () => {
-  const [showLifetimeBanner, setShowLifetimeBanner] = React.useState(false);
+  const [showLifetimeBanner, setShowLifetimeBanner] = React.useState(true);
   const { translate } = useTranslationContext();
   const onClose = () => {
     setShowLifetimeBanner(false);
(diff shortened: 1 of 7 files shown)


src/app/banners/Banner.tsx
```diff
@@ -1,11 +1,10 @@
 import React, { useEffect } from 'react';
 import { X } from 'phosphor-react';
-import Hearts from 'assets/images/banner/hearts.png';
-import NeonBlur from 'assets/images/banner/neonBlur.png';
+import NeonBlur from 'assets/images/banner/infinity.svg';
 import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
 
 const Banner = () => {
-  const [showLifetimeBanner, setShowLifetimeBanner] = React.useState(false);
+  const [showLifetimeBanner, setShowLifetimeBanner] = React.useState(true);
```
CandelR: I would add the opening and closing logic and the possible variables outside the Banner component, so you can reuse it for each banner you need for the different campaigns by creating a wrapper.

src/app/core/views/Preferences/tabs/Account/Language.tsx
```diff
@@ -25,8 +25,8 @@ export default function Language(): JSX.Element {
 
   useEffect(() => {
     if (localStorageLanguage) {
-      setCurrentLangText(currentLang[localStorageLanguage as string]);
-      setLang(localStorageLanguage as string);
+      setCurrentLangText(currentLang[localStorageLanguage]);
+      setLang(localStorageLanguage);
```
CandelR: If you always have to change both states at the same time, add it to a function, I see that the code is repeated 3 times.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -155,24 +154,10 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
...
-    if (!isSignUpTutorialCompleted && currentTutorialStep === 1 && successNotifications.length > 0) {
-      setShowSecondTutorialStep(true);
-    }
-  }, [successNotifications]);
+  const plan = useSelector<RootState, PlanState>((state) => state.plan);
 
   const showTutorial =
-    useAppSelector(userSelectors.hasSignedToday) &&
-    !isSignUpTutorialCompleted &&
-    (showSecondTutorialStep || currentTutorialStep === 0);
```
CandelR: Why did you remove it? This will break the onboarding tutorial

CandelR (changes requested):

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
xabg2: Ups. It is my fault

CandelR (approved):

PR #668, opened 2023-04-18
[PB-337]:fix/added validation for uploaded file size

rafijv:
Until now, it was checked if the space used exceeded the limit of the user's plan. Now I have changed it to check the space used plus the files uploaded.


src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -61,8 +61,10 @@ export const uploadItemsThunk = createAsyncThunk<void, UploadItemsPayload, { sta
     try {
       const planLimit = getState().plan.planLimit;
       const planUsage = getState().plan.planUsage;
+      const uploadItemsSize = Object.values(files).reduce((acum, file) => acum + file.size, 0);
+      const totalItemsSize = uploadItemsSize + planUsage;
 
-      if (planLimit && planUsage >= planLimit) {
+      if (totalItemsSize >= planLimit) {
```
CandelR: It seems that has removed the `planLimit` truly checking before the comparison with the usage. It is not necessary that checking?

CandelR (approved):

PR #669, opened 2023-04-19
[MKT-55]: Feat/update 404 page design

xabg2:
Update 404 page design.
Links added to:
* Drive Home page.
* Help support page.
* Open mail to contact us.(hello@internxt.com)

### DIFF
--- a/src/app/i18n/locales/en.json
+++ b/src/app/i18n/locales/en.json
@@ -33,18 +33,15 @@
     "description": "The page you are looking for could not be found.",
     "card1": {
       "title": "Return Home",
-      "description": "Go back to Internxt Drive.",
-      "urlRedirect": "https://drive.internxt.com/app"
+      "description": "Go back to Internxt Drive."
     },
     "card2": {
       "title": "Visit Help Center",
-      "description": "Troubleshoot common issues and browse our FAQ.",
-      "urlRedirect": "https://help.internxt.com/en/"
+      "description": "Troubleshoot common issues and browse our FAQ."
     },
     "card3": {
       "title": "Contact Support",
-      "description": "Reach out to our Support Team at hello@internxt.com.",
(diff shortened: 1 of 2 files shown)


src/app/i18n/locales/en.json
```diff
@@ -27,6 +27,26 @@
     },
     "cta": "Get the deal"
   },
+  "notFound": {
+    "header": "Error 404",
+    "title": "Nothing to see here...",
+    "description": "The page you are looking for could not be found.",
+    "card1": {
```
CandelR, line 34: French and Spanish translations are missing

src/app/i18n/locales/en.json
```diff
@@ -27,6 +27,26 @@
...
     "cta": "Get the deal"
   },
+  "notFound": {
+    "header": "Error 404",
+    "title": "Nothing to see here...",
+    "description": "The page you are looking for could not be found.",
+    "card1": {
+      "title": "Return Home",
+      "description": "Go back to Internxt Drive.",
+      "urlRedirect": "https://drive.internxt.com/app"
```
CandelR: Not add url to translation files. Use the navigationService in the cards object that has you created to redirect the user to other view of the web, use constants or what do you think fits better depend of the case.

CandelR (changes requested):

src/app/i18n/locales/en.json
xabg2: For the moment it will only be in English because the website has the same behavior.

xabg2: So, as English is the fallback lang, it always appears in english for the time being.

src/app/i18n/locales/en.json
xabg2: Fixed

CandelR (approved):

PR #671, opened 2023-04-20
[PB-20]: feature/provide incentive during cancellation

rafijv:
A new dialog has been added to offer an incentive to users who wish to cancel their account.

The dialog is displayed if the offer has not been previously taken advantage of, if the customer accepts the offer, the dialog is closed and a temporary message is displayed indicating that the offer has been applied. The UI is also refreshed.

### DIFF
--- a/src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
+++ b/src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
@@ -5,6 +5,7 @@ import Button from '../../../../../shared/components/Button/Button';
 import Modal from '../../../../../shared/components/Modal';
 import { FreeStoragePlan } from '../../../../../drive/types';
 import sizeService from '../../../../../drive/services/size.service';
+import paymentService from 'app/payment/services/payment.service';
 
 const CancelSubscriptionModal = ({
   isOpen,
@@ -24,30 +25,43 @@ const CancelSubscriptionModal = ({
   cancelSubscription: (feedback: string) => void;
 }): JSX.Element => {
   const { translate } = useTranslationContext();
-  const [step, setStep] = useState<1 | 2>(1);
+  const [step, setStep] = useState<1 | 2 | 3>(1);
+  const [couponAvailable, setCouponAvailable] = useState(false);
 
   useEffect(() => {
-    if (isOpen) {
-      setStep(1);
+    paymentService.getCoupon().then((coupon) => setCouponAvailable(coupon.elegible));
+  }, []);
+
+  useEffect(() => {
+    if (!couponAvailable && isOpen) {
+      setStep(2);
     }
-  }, [isOpen]);
+  }, [couponAvailable]);
 
   return (
     <Modal isOpen={isOpen} onClose={onClose}>
-      <h1 className="text-2xl font-medium text-gray-80">
-        {translate('views.account.tabs.billing.cancelSubscriptionModal.title')}
-      </h1>
-      <h2 className="text-base font-light text-gray-50">{step} of 2</h2>
-      {step === 1 ? (
-        <Step1
+      {(step === 2 || step === 3) && (
+        <>
+          <h1 className="text-2xl font-medium text-gray-80">
+            {translate('views.account.tabs.billing.cancelSubscriptionModal.title')}
+          </h1>
+          <h2 className="text-base font-light text-gray-50">{step - 1} of 2</h2>
+        </>
+      )}
+      {step === 1 && <Step1 currentPlanName={currentPlanName} onClose={onClose} setStep={setStep} />}
+
+      {step === 2 && (
+        <Step2
           currentPlanName={currentPlanName}
           onClose={onClose}
           setStep={setStep}
           currentPlanInfo={currentPlanInfo}
           currentUsage={currentUsage}
         />
-      ) : (
-        <Step2
+      )}
+
+      {step === 3 && (
+        <Step3
           currentPlanName={currentPlanName}
           onClose={onClose}
           cancellingSubscription={cancellingSubscription}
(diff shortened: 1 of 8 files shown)


src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
```diff
@@ -24,30 +28,60 @@ const CancelSubscriptionModal = ({
...
+  useEffect(() => {
+    couponAvailable && setStep(1);
+  }, [couponAvailable]);
+
+  const applyCoupon = async () => {
+    try {
+      await paymentService.applyCoupon();
+      notificationsService.show({ text: translate('notificationMessages.successApplyCoupon') });
+    } catch (error) {
+      console.error(error);
```
sg-gs: Remove this `console.error`

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
```diff
@@ -24,30 +28,60 @@ const CancelSubscriptionModal = ({
...
+      notificationsService.show({
```
sg-gs: What if the coupon has been already applied and the user is not eligible? Do we show something more specific? If we are working on reducing the churn, this could happen a lot, even if we handle the modal not being shown. (p.e: a user that left a window before applying the coupon opened with the modal, a lot of people could try to trick us). 

So, if someone tries something tricky or any bug is introduced in the future on the front end, showing that the user is not eligible will be enough to avoid unnecessary tickets of people complaining about the coupon appliance failing.

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
```diff
@@ -24,30 +28,60 @@ const CancelSubscriptionModal = ({
...
+        text: translate('notificationMessages.errorApplyCoupon'),
+        type: ToastType.Error,
+      });
+    } finally {
+      onClose();
+      setTimeout(() => {
```
sg-gs, line 47: Why? It this for refreshing the UI? Then, we should just refresh the information in the case that an error does not happen, isn't it?

src/app/payment/services/payment.service.ts
```diff
@@ -82,6 +83,16 @@ const paymentService = {
     return paymentsClient.getPrices();
   },
 
+  async getCoupon(): Promise<CouponAvailable> {
+    const paymentsClient = await SdkFactory.getInstance().createPaymentsClient();
```
sg-gs, line 87: It could be less verbose if we do something like this: 

`return SdkFactory.getInstance().createPaymentsClient().getCoupon();`

src/app/payment/services/payment.service.ts
```diff
@@ -82,6 +83,16 @@ const paymentService = {
...
+    return paymentsClient.getCoupon();
+  },
+
+  async applyCoupon(): Promise<CouponAvailable> {
+    const paymentsClient = await SdkFactory.getInstance().createPaymentsClient();
+    return paymentsClient.applyCoupon();
```
sg-gs, line 93: It could be less verbose if we do something like this: 

`return SdkFactory.getInstance().createPaymentsClient().applyCoupon();`

sg-gs (changes requested):

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
rafijv: If the coupon has been applied, the modal is not displayed, but I am not sure if I understood you.
Do you want to change the message that is displayed in case the coupon cannot be applied?

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
rafijv: It's for this
I change it

src/app/payment/services/payment.service.ts
rafijv: To do it this way you would have to change createPaymentsClient() and SDK
It seems that this refactoring is out of the scope of the task.

src/app/payment/services/payment.service.ts
rafijv: To do it this way you would have to change createPaymentsClient() and SDK
It seems that this refactoring is out of the scope of the task.

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
sg-gs: So, assume that I open a window and the modal appears, but I left the web app there, without pressing the apply button. Then I open another window and the modal appears again - as I have not still the coupon applied - and then I apply it. After doing that, I return to the first window and press the apply button. I am going to get a generic error. 

This is a corner case and is not too important, but some users are going to try that. 

Also, even ignoring that case, if we introduce some bug unintentionally and the modal shows when should not, we should still show the proper error message, which in that case would be 'The coupon has been already applied' or something like this.

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
rafijv: Done

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
rafijv: Done

src/app/core/views/Preferences/tabs/Billing/CancelSubscriptionModal.tsx
rafijv: Done

sg-gs (approved):

PR #677, opened 2023-04-26
[_]: fix/Buy lifetime plan with coupon when user has subscription

xabg2:
Now, user can buy a lifetime plan from internxt.com/lifetime with the discount when he has a subscription activated.


src/app/payment/views/CheckoutView/CheckoutPlanView.tsx
```diff
@@ -78,13 +76,22 @@ export default function CheckoutPlanView(): JSX.Element {
...
+                customer_email: user.email,
+                mode: mode,
+              }))
+            : (response = await paymentService.createCheckoutSession({
+                price_id: planId,
+                success_url: `${window.location.origin}/checkout/success`,
+                cancel_url: 'https://drive.internxt.com/preferences?tab=plans',
+                customer_email: user.email,
+                mode: mode,
+              }));
```
CandelR: As the only difference the object has is the coupon, perhaps adding an interface to define the type and putting the code like this makes it a bit clearer what changes there are between one structure and the other.
Adding the interface:
```
interface CheckoutOptions {
  price_id: string;
  coupon_code?: string;
  success_url: string;
  cancel_url: string;
  customer_email: string;
  mode: string;
}
```
```suggestion
          const checkoutOptions: CheckoutOptions = {
            price_id: planId,
            success_url: `${window.location.origin}/checkout/success`,
            cancel_url: 'https://drive.internxt.com/preferences?tab=plans',
            customer_email: user.email,
            mode: mode,
          };

          if (coupon) {
            checkoutOptions.coupon_code = coupon;
          }

          response = await paymentService.createCheckoutSession(checkoutOptions);
```

CandelR (changes requested):

src/app/payment/views/CheckoutView/CheckoutPlanView.tsx
xabg2: Done

CandelR (approved):

PR #697, opened 2023-05-17
[PB-442]: bugfix/stop upload on lost connection

CandelR:
- Added error handler to notify user when lost internet connection
- Disabled auto retry in upload when the user cancels
- Added sentry breadcrumbs to better identify bugs and their replication.


sg-gs (changes requested):
Some translations are missing, apart from that LGTM

sg-gs (approved):

PR #703, opened 2023-05-25
[_]: feat/Russian translation

xabg2:
Translate Drive Web to Russian


src/app/core/views/Preferences/tabs/Account/Language.tsx
```diff
@@ -133,6 +139,16 @@ export default function Language(): JSX.Element {
...
+              key={'ru'}
+              onClick={() => {
+                setLang('ru');
+                i18next.changeLanguage('ru');
+                dayjs.locale(ru);
+              }}
+            >
+              <p>{translate('lang.ru')}</p>
+            </MenuItem>,
           ]}
```
CandelR: This is getting too long, to facilitate the maintenance of the code, better to change it for a component that, given an array with the different languages, generates as many `MenuItem` as needed.

CandelR (approved):

PR #706, opened 2023-05-31
[MKT-100]: feat/Summer banner

xabg2:
Banner to promote the summer offer.


src/app/banners/BannerWrapper.tsx
```diff
@@ -1,23 +1,40 @@
...
-  useEffect(() => {
-    if (!localStorage.getItem('showLifetimeBanner')) {
-      setTimeout(() => {
+  function isBannerOpen() {
+    if (isNewAccount) {
+      if (userPlan === 'free' && isTutorialCompleted && !localStorageService.get('showSummerBanner')) {
+        setShowBanner(true);
+      }
+    } else {
+      if (userPlan === 'free' && !localStorageService.get('showSummerBanner')) {
```
CandelR: You could reduce this conditions to:
```suggestion
const shouldShowBanner = userPlan === 'free' && !localStorageService.get('showSummerBanner');

if ((isNewAccount && isTutorialCompleted && shouldShowBanner) || (!isNewAccount && shouldShowBanner)) {
  setShowBanner(true);
}
```

src/app/banners/BannerWrapper.tsx
```diff
@@ -1,23 +1,40 @@
...
+  const onCloseBanner = () => {
     setShowBanner(false);
-    localStorage.setItem('showLifetimeBanner', 'false');
+    localStorage.setItem('showSummerBanner', 'false');
   };
 
-  useEffect(() => {
-    if (!localStorage.getItem('showLifetimeBanner')) {
-      setTimeout(() => {
+  function isBannerOpen() {
```
CandelR: This name is more like a boolean variable name. Try using a verb to name functions to add a reference to what it does. 
For example: handleBannerDisplay()

src/app/banners/BannerWrapper.tsx
```diff
@@ -1,23 +1,40 @@
...
-      }, 10000);
+      }
     }
-  }, []);
+  }
 
-  return <Banner showBanner={showBanner} onClose={onClose} />;
+  useEffect(() => {
+    isBannerOpen();
+  }, [isTutorialCompleted, localStorageService.get('showSummerBanner'), userPlan, isNewAccount]);
```
CandelR: I think that localStorageService.get('showSummerBanner') dependency is not necessary.  
Besides, I think it can cause strange behaviour, because it is not a variable, it is a function that will be executed with each change creating new references to the value.

src/app/banners/SummerBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <div className="flex flex-col items-center justify-center space-y-9 text-center">
+            <InternxtLogo />
+            <div className="flex flex-col items-center justify-center space-y-6">
+              <div className="flex max-w-[443px] items-center justify-center rounded-xl bg-mint px-10 py-3 text-white">
+                <p className="text-4xl font-semibold">SUMMER DEAL!</p>
+              </div>
+              <div className="flex flex-col items-center text-white">
+                <p className="text-8xl font-bold">90% off</p>
+                <p className="pt-1 text-3xl font-bold">2TB PLAN FOR 1 YEAR</p>
+              </div>
```
CandelR, line 82: Should the banner only be displayed in English?

CandelR (changes requested):

src/app/banners/BannerWrapper.tsx
xabg2: Okay, I'll remove the dependency.

xabg2: Done

CandelR (approved):

PR #708, opened 2023-06-01
[PB-256] feature/Integrate pagination in front end

CandelR:
- Added drive pagination and disabled items sorting


src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -212,10 +215,21 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
   }, []);
 
   useEffect(() => {
-    if (isTrash && !hasMoreTrashFolders) {
-      getMoreTrashItems().catch((error) => errorService.reportError(error));
+    if ((!isTrash && !hasMoreFolders) || (isTrash && !hasMoreTrashFolders)) {
+      fetchItems();
```
sg-gs, line 220: where is the `catch` for this?

sg-gs (changes requested):

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
CandelR: I've added it to the trash fetch logic

sg-gs (approved):

PR #714, opened 2023-06-20
[MKT-115]: feat/Allow different payments methods

xabg2:
Allow new payments methods:
* PayPal
* iDEAL
* Sofort
* bancontact


src/app/payment/services/payment.service.ts
```diff
@@ -58,7 +58,7 @@ const paymentService = {
     return paymentsClient.getSetupIntent();
   },
 
-  async getDefaultPaymentMethod(): Promise<PaymentMethod> {
+  async getDefaultPaymentMethod(): Promise<PaymentMethod | any> {
```
sg-gs: why?

sg-gs (changes requested):

src/app/payment/services/payment.service.ts
xabg2: Because the returned object usually has this structure:
```export interface PaymentMethod {
  id: string;
  card: {
    brand: 'amex' | 'diners' | 'discover' | 'jcb' | 'mastercard' | 'unionpay' | 'visa' | 'unknown';
    exp_month: number;
    exp_year: number;
    last4: string;
  };
  created: number;
}
```
Then, when it is not a card, it returns nothing and is detected as a free plan. That's why I put any, so that if it returns an object different from this one, we can still get it. But yes, I have added the correct typing in both [payment.service.ts](https://github.com/internxt/drive-web/pull/714#:~:text=src/app/payment/services/payment.service.ts) and PaymentMethod.tsx.

sg-gs (approved):

CandelR (approved):

PR #725, opened 2023-07-03
[PB-665]: feat/user can request access to a file or folder

rafijv:
Added new view that allows users to request access to files or folders


src/app/drive/views/RequestAccess/RequestAccess.tsx
```diff
@@ -0,0 +1,99 @@
...
+function RequestAccess(): JSX.Element {
+  const user = useSelector<RootState, UserSettings | undefined>((state) => state.user.user);
+  const { translate } = useTranslationContext();
+
+  const [messageText, setMessageText] = useState<string>('');
+  const [messageTextLimit, setMessageTextLimit] = useState<boolean>(false);
+  const [requestSent, setRequestSent] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (messageText.length >= 950) {
```
CandelR: Avoid use magic numbers. Add it to a const in order to know what represents the number or what is its purpose with the name

src/app/core/config/app.json
```diff
@@ -218,6 +218,13 @@
       "path": "/verify-email/:token",
       "exact": false
     },
+    {
+      "id": "request-access",
+      "layout": "empty",
+      "path": "/request-access",
```
CandelR, line 224: Does the url not have to have a file or folder id?

src/app/core/config/views.ts
```diff
@@ -68,6 +69,7 @@ const views: Array<{
   { id: AppView.SharePhotosToken, component: SharePhotosView },
   { id: AppView.RedirectToApp, component: RedirectToAppView },
   { id: AppView.VerifyEmail, component: VerifyEmailView },
+  { id: AppView.RequestAccess, component: RequestAccess },
```
CandelR: Now it is accessible, hide it until the task is completed if this is just the UI.

CandelR (changes requested):
The Change button will be moved out of the div if the email is long.
image

src/app/core/config/app.json
rafijv: I don't know how this is going to work, we will have to take it up again when the back is finished and we start with the logic.

src/app/core/config/views.ts
rafijv: I have commented it so it is not available

src/app/drive/views/RequestAccess/RequestAccess.tsx
rafijv: Done

CandelR (approved):

PR #737, opened 2023-07-11
[PB-592] fix/Fast navigation in the menu shows a wrong message

lvargas4:
- Updated views rerender method to dismiss previous toast notifications.

This can be handled for singular cases of toasts using an id, although I thought this would be a better approach since the different toasts would be belonging to other views. Let me know if you'd rather make this more specific.


CandelR (changes requested):
This does not solve the problem of the error notification appearing when navigating quickly between views, it only hides if you keep navigating the notifications that have appeared, the error notification still appears. 
Better to focus on finding out what causes that error notification to appear and fix that problem.  👀

lvargas4:
Seems like on the hasMoreDriveFolders/hasMoreDriveFiles validation missed the case where both flags where false,, that caused that itemsPromise where empty and had no a promise to await making the function to fail reading the "result" property as "items" where undefined

CandelR (approved):

PR #739, opened 2023-07-17
[PB-670] fix/wrong euro sign on plan modal

miguelsw:
Issue:
- Add € in current plan

Details:
- I have added the euro symbol to the current plan, replicating it as done in the selected plan.

Last Update:
- I am using a function from money service to get the currency symbol


src/app/core/views/Preferences/tabs/Plans/PlanSelector.tsx
```diff
@@ -351,10 +351,7 @@ const ChangePlanDialog = ({
           <p className="text-2xl font-medium text-primary">{currentPlanSizeString}</p>
           {subscription?.type === 'subscription' ? (
             <div>
-              <span className="text-base font-medium">
-                {subscription.currency}
-                {displayAmount(currentAmountMonthly)}
-              </span>
+              <span className="text-base font-medium">{`€${displayAmount(currentAmountMonthly)}`}</span>
```
CandelR: Try not to hardcode the currency, use the incoming currency and make an object or function to parse to the desired character. 
If you have seen it somewhere else it is likely that it has slipped in, if you can change where you have seen that it was perfect. 👌

CandelR (changes requested):

src/app/core/views/Preferences/tabs/Plans/PlanSelector.tsx
miguelsw: @CandelR I found the money service to use a function to get the currency symbol, thanks for the advice!

CandelR (approved):

PR #741, opened 2023-07-18
Feat/WT-325-global-search

aboredvaro:


src/app/core/components/Navbar/Navbar.tsx
```diff
@@ -31,68 +28,321 @@ interface NavbarProps {
...
+    let item: number | null = null;
+    if (e.key === 'Escape') {
+      setPreventBlur(false);
+    } else if (e.key === 'ArrowUp') {
+      e.preventDefault();
+      setSelectedResult((current) => (current > 0 ? current - 1 : 0));
+      item = selectedResult > 0 ? selectedResult - 1 : 0;
+    } else if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      setSelectedResult((current) => (current < searchResult.length - 1 ? current + 1 : searchResult.length - 1));
```
CandelR: Perhaps we could extract the `searchResult.length - 1` to a const named `lastItemIndex` or something similar to make the code more readable, as it repeats a lot in the conditions

src/app/core/components/Navbar/Navbar.tsx
```diff
@@ -31,68 +28,321 @@ interface NavbarProps {
...
+    },
+    {
+      id: 'video',
+      Icon: iconService.getItemIcon(false, 'mp4'),
+      name: translate('general.searchBar.filters.video') as string,
+    },
+    {
+      id: 'audio',
+      Icon: iconService.getItemIcon(false, 'mp3'),
+      name: translate('general.searchBar.filters.audio') as string,
```
CandelR: Why does each function `translate` the casting to string? I don't think it is necessary since the `translate` function returns a string

src/app/drive/services/new-storage.service.ts
```diff
@@ -0,0 +1,14 @@
+import { DriveFileData } from '@internxt/sdk/dist/drive/storage/types';
+import { SdkFactory } from '../../core/factory/sdk';
+
+export async function searchItemsByName(name: string): Promise<DriveFileData[]> {
```
CandelR, line 4: Why has this been added but not used?

CandelR (approved):

PR #745, opened 2023-07-19
[_]: fix/icons for files

xabg2:
Show the correct icon for each file depending on its extension and also allows the user to preview webp images.

### DIFF
--- a/src/app/drive/components/FileViewer/FileViewer.tsx
+++ b/src/app/drive/components/FileViewer/FileViewer.tsx
@@ -17,6 +17,7 @@ import { RootState } from 'app/store';
 import { uiActions } from 'app/store/slices/ui';
 import { setItemsToMove, storageActions } from '../../../store/slices/storage';
 import { isLargeFile } from 'app/core/services/media.service';
+import errorService from 'app/core/services/error.service';
 
 interface FileViewerProps {
   file: DriveFileData;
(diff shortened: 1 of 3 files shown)


src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -203,10 +203,10 @@ const FileViewer = ({
       downloader(new AbortController())
         .then((blob) => {
           setBlob(blob);
-          setIsErrorWhileDownloading(false);
         })
         .catch((err) => {
-          // TODO
+          const error = err as Error;
+          console.error('[DOWNLOAD FILE/ERROR]: ', error.stack || error.message || error);
```
CandelR, line 211: Add `errorService.reportError` function to report the errors to sentry

src/app/drive/types/file-types.ts
```diff
@@ -245,9 +246,9 @@ const excludeUnsupportedVideoExtensions: string[] = [
   'f4b',
 ];
 
-const previewVideoExtensionsGroup: string[] = Object.values(videoExtensions)
-  .flatMap((extensions) => extensions.flat())
-  .filter((extension) => !excludeUnsupportedVideoExtensions.includes(extension));
+const previewVideoExtensionsGroup: string[] = Object.keys(videoExtensions).filter(
+  (extension) => !excludeUnsupportedVideoExtensions.includes(extension),
+);
```
CandelR, line 251: Why only use keys instead of all values of videoExtensions?

xabg2: Because I think here it is filtered by keys, so when we try to search for the mp4 extension for example, it tells us that it cannot be displayed because the key is mpeg-4: [here](https://github.com/internxt/drive-web/blob/aeb2e772457455e50009d308a985ddb0947101b4/src/app/drive/services/file-extension.service.ts)

CandelR (approved):

PR #753, opened 2023-07-25
[PB-530, PB-728]: feature/encrypt owners key with invited user public key

CandelR:
- Added ability to encrypt the owner's encryption key with the invited user public key

⚠️ Need to merge first:
- https://github.com/internxt/sdk/pull/150
- https://github.com/internxt/drive-server-wip/pull/185


src/app/crypto/services/pgp.service.ts
```diff
@@ -1,21 +1,59 @@
...
 }
+
+export const encryptMessageWithPublicKey = async ({
+  message,
+  publicKeyInBase64,
+}: {
+  message: string;
+  publicKeyInBase64: string;
+}): Promise<openpgp.WebStream<string>> => {
+  const publicKeyArmored = Buffer.from(publicKeyInBase64, 'base64').toString('ascii');
```
sg-gs: Why ascii? Why not utf8? Why not using base64 directly?

CandelR: It doesn't need to be ascii, it slipped through. 
I'm going to add unit tests for this service while I'm at it. 👌

src/app/crypto/services/pgp.service.ts
```diff
@@ -1,21 +1,58 @@
-export async function getOpenpgp(): Promise<typeof import('openpgp')> {
-  return import('openpgp');
-}
+import * as openpgp from 'openpgp';
```
sg-gs: The lazy loading of openpgp is due to the size of the library in terms of SEO / Web app speed, I would leave that as it is

src/app/crypto/services/pgp.service.ts
```diff
@@ -1,21 +1,58 @@
...
+};
+
+export const decryptMessageWithPrivateKey = async ({
+  encryptedMessage,
+  privateKeyInBase64,
+}: {
+  encryptedMessage: openpgp.WebStream<string>;
+  privateKeyInBase64: string;
+}): Promise<openpgp.MaybeStream<openpgp.Data> & openpgp.WebStream<Uint8Array>> => {
+  const privateKeyArmored = Buffer.from(privateKeyInBase64, 'base64').toString('ascii');
```
sg-gs: Do not use `ascii`, it is `utf8`, which is the default format when doing a `toString()` with no params

src/app/crypto/services/utilspgp.ts
```diff
@@ -9,74 +8,70 @@ export async function isValidBase64(key: string): Promise<boolean> {
 }
 
 export async function isValid(key: string): Promise<boolean> {
-  const openpgp = await getOpenpgp();
-  const keyResult = await openpgp.key.readArmored(key);
-
-  return !keyResult.err;
+  try {
+    await openpgp.readKey({ armoredKey: key });
```
sg-gs, line 14: Does this throw if the key is invalid?

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -61,12 +67,23 @@ const ShareInviteDialog = (props: ShareInviteDialog): JSX.Element => {
...
+      }
+      return userToInvite;
+    });
+    setUsersToInvite(newUserToInvite);
   };
 
   const onInvite = () => {
-    // Ivite added users
+    usersToInvite.forEach((user) => {
+      const userRoleId = props.roles.find((role) => role.role === user.userRole)?.id;
```
sg-gs: Shouldn't this be `user.role`?

sg-gs: Shouldn't this be `role` directly?

sg-gs (changes requested):

src/app/crypto/services/utilspgp.ts
CandelR: Yes, if it is invalid it thows an error, then return false.

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
CandelR: No, it is the format of the backend `roles` object. `role.role` is the name of the role

sg-gs: Okay, try to abstract your data structures from the backend, we should not depend on how the backend chooses to expose things, or any change there will break the entire frontend. Instead of that, I recommend a layer in-between which may assert that the expected data comes from the backend and transforms that to the preferable data structure. That also will make testing easier & faster

CandelR: Totally agree, I'm going to extract the call to get the roles from the ShareDialog view and add the layer already in this PR and so it is already prepared, as it is not definitive I was putting together the calls to extract it all at once once once we had them all.

test/unit/services/pgp.service.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      publicKeyInBase64,
+    });
+
+    expect(encryptedMessage).toBeDefined();
+  });
+
+  it('should generate keys, encrypt and decrypt a message successfully', async () => {
+    // Step 1: Generate new keys
+    const keys = await generateNewKeys();
+    expect(keys).toHaveProperty('privateKeyArmored');
```
sg-gs: Hey, as you have asserted this on the other test case, you could remove these assertions or put that at the end of the test, so we prepare first, then act, then assert

sg-gs (approved):

sg-gs (approved):

PR #756, opened 2023-07-28
[PB-532] feature/Ability to decrypt the owner's encryption key with the invited user private key

CandelR:
- Added ability to decrypt the owner's encryption key with the invited user private key

⚠️ Do not review until this PR is merged:
- https://github.com/internxt/drive-web/pull/733


src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -45,8 +47,48 @@ function copyShareLink(type: string, code: string, token: string) {
...
+//TODO: TEMPORARY TYPE, COMPLETE WHILE ADVANCING IN AFS
+export type SharedLinkItemType = {
+  id: string;
+  folderId: string;
+  ownerId: string;
+  sharedWith: string;
+  encryptionKey: string;
+  createdAt: string;
+  updatedAt: string;
+  owner_id: string;
```
sg-gs: Let's avoid using camelCase so we are not using these fields from the beginning

src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -45,8 +47,48 @@ function copyShareLink(type: string, code: string, token: string) {
...
+  shared_with: string;
```
sg-gs: Here too

sg-gs (changes requested):

PR #759, opened 2023-07-31
[PB-750]: fix/remove unused code

xabg2:
Unused code in the LogIn component. It was an unused payment code since we changed the flow when the user tries to purchase a plan from the website.


xabg2:
If you can, just give me approval and I will recheck the flow to check that everything works fine. @CandelR

CandelR (approved):

PR #762, opened 2023-08-01
[PB-700]: feat/Reset account

rafijv:
Changes:
- Add reset account without backup key


src/app/auth/services/auth.service.ts
```diff
@@ -205,6 +205,24 @@ const updateCredentialsWithToken = async (
...
+  const authClient = SdkFactory.getNewApiInstance().createAuthClient();
+
+  return authClient.resetAccountWithToken(
+    token,
+    encryptedHashedNewPassword,
+    encryptedHashedNewPasswordSalt,
+    encryptedNewMnemonic,
+  );
+};
+
```
CandelR, line 225: Don't new public and private keys have to be generated with the account reset?

rafijv: No

CandelR (approved):

PR #765, opened 2023-08-01
[PB-516, PB-3058]: feat/send private key encrypted to server wip

miguelsw:
Details:
- We need to send the private key to drive-server-wip instead of sending it to drive-server
- The endpoint used is /password
Updates:
- Added new client methods in the factory to send the private key encrypted to drive-server-wip
- Added field `encryptVersion` to the payload, the backend requires it

⚠️ One change in one method on the SDK is needed, link here:
https://github.com/internxt/sdk/pull/159


src/app/auth/services/auth.service.ts
```diff
@@ -217,7 +222,7 @@ export const changePassword = async (newPassword: string, currentPassword: strin
   const privateKey = Buffer.from(user.privateKey, 'base64').toString();
   const privateKeyEncrypted = aes.encrypt(privateKey, newPassword, getAesInitFromEnv());
 
-  const usersClient = SdkFactory.getInstance().createUsersClient();
+  const usersClient = SdkFactory.getNewApiInstance().createNewUsersClient();
```
CandelR, line 299: Does the new key now have to be sent to the server-wip and not to the server? Wouldn't this break any functionality?

miguelsw: @CandelR In reality, it seems that a migration is being done because Carlos' task of this PR ->https://github.com/internxt/drive-server-wip/pull/175 , which is related to this, the enpoint was created in server-wip, I tested and is working well if we merge this change in sdk repository -> https://github.com/internxt/sdk/pull/159

CandelR (approved):
Waiiting to merge SDK PR

sg-gs:
What about this one @CandelR, is this still needed?

CandelR:
> What about this one @CandelR, is this still needed?

@sg-gs  If we want to migrate the endpoint from server to server-wip I understand that yes. I will talk to Andres about this to see what we do, as he can give me more context on whether it is necessary to migrate the enpoint or not

CandelR:
Endpoint not working properly, blocked until new change password enpoint returns the new generated tokens.

CandelR (approved):
A correction had to be made in drive-sever with the token generation date and drive-server-wip with the return of tokens when changing the password. 
It is now ready for QA

sg-gs (approved):

PR #766, opened 2023-08-01
[PB-760] fix/hot keys control + a is not working

miguelsw:
- Fixed keyboard shortcut "control + a" to select all items


src/app/shared/components/List/index.tsx
```diff
@@ -178,7 +178,7 @@ ListProps<T, F>): JSX.Element {
     'ctrl+a, cmd+a',
     (e) => {
       e.preventDefault();
-      handleKeyPress(selectAllItems);
+      handleKeyPress(selectAllItems());
```
CandelR, line 181: This is not correct, it works because you are executing `selectAllItems` directly, to execute correctly `handleKeyPress` you must call `handleKeyPress` function.
Change `handleKeyPress(selectAllItems());` to `handleKeyPress(selectAllItems)();`

CandelR (approved):

src/app/shared/components/List/index.tsx
CandelR: This is not correct, it works because you are executing `selectAllItems` directly, to execute correctly `handleKeyPress` you must call `handleKeyPress` returned function.
Change `handleKeyPress(selectAllItems())`; to `handleKeyPress(selectAllItems)();`

CandelR (changes requested):

src/app/shared/components/List/index.tsx
miguelsw: It works that way, thank you for telling me the correct way to use that function.

CandelR (approved):

PR #772, opened 2023-08-08
[PB-826]  bugfix/Share multiple photos not working

CandelR:
- Fixed zip generation when download multiple shared fotos

### DIFF
--- a/src/app/share/views/ShareView/SharePhotosView.tsx
+++ b/src/app/share/views/ShareView/SharePhotosView.tsx
@@ -16,12 +16,10 @@ import { GetPhotoShareResponse, PhotoId } from '@internxt/sdk/dist/photos';
 import { SdkFactory } from '../../../core/factory/sdk';
 import network from 'app/network';
 import downloadService from '../../../drive/services/download.service';
-import JSZip from 'jszip';
-import { Readable } from 'stream';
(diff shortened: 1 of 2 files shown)


src/app/share/views/ShareView/SharePhotosView.tsx
```diff
@@ -139,8 +139,37 @@ const SharePhotosView = (props: SharePhotosProps): JSX.Element => {
...
+            const totalLength = chunks.reduce((total, chunk) => total + chunk.length, 0);
+            const concatenatedUint8Array = new Uint8Array(totalLength);
+
+            let offset = 0;
+            for (const chunk of chunks) {
+              concatenatedUint8Array.set(chunk, offset);
+              offset += chunk.length;
+            }
+
+            zip.file(photoName, concatenatedUint8Array, { compression: 'DEFLATE' });
```
sg-gs: Why buffering everything on memory and not using the stream on the fly? This is going to screw the user's device from >1GB

CandelR: It doesn't seem to work with streams. I'll try to investigate a bit more to see what exactly is going on.

sg-gs (approved):

PR #785, opened 2023-08-22
[PB-896]: Feat/Rename shared files

larry-internxt:
- Added resource token for renaming files functionality
- Managed to get file renaming working
- Restructured interfaces for better organization
- Renamed token from xInvitedToken to xResourcesToken for better understanding
- Added needed second token for actions, called nextResourcesToken
- Fixed sharedLink view bugs
- Added refresh name on list after renaming it correctly

### DIFF
--- a/src/app/share/types/index.ts
+++ b/src/app/share/types/index.ts
@@ -1,0 +1,47 @@
+import { SharedFiles, SharedFolders } from '@internxt/sdk/dist/drive/share/types';
+
+export type AdvancedSharedLink = SharedFolders &
+  SharedFiles & {
+    isFolder: boolean;
+    isRootLink: boolean;
+  };
+
+export type OrderBy = { field: 'views' | 'createdAt'; direction: 'ASC' | 'DESC' } | undefined;
+
+export type SharedLinkItemType = {
+  id: string;
+  folderId: string;
+  ownerId: string;
+  sharedWith: string;
+  encryptionKey: string;
+  createdAt: string;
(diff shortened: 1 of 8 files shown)


src/app/share/types/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type SharedLinkItemType = {
+  id: string;
+  folderId: string;
+  ownerId: string;
+  sharedWith: string;
+  encryptionKey: string;
+  createdAt: string;
+  updatedAt: string;
+  owner_id: string;
```
sg-gs: Try to avoid using this fields, if they arrive, as we should just use camelCase

src/app/share/types/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    removedAt: string | null;
+  };
+  owner: {
+    uuid: string;
+    email: string;
+    name: string;
+    lastname: string;
+    avatar: string | null;
+  };
+  file: any;
```
sg-gs: Type this

sg-gs (changes requested):

src/app/share/types/index.ts
larry-disabled: hey this entire type was created by Ramon but I think it is not ended and it is not being used as it was planned. I've decided to remove it, so when Ramon comes back we will discuss it to see if it is really necessary or not. We can always bring it back in that case

sg-gs (approved):

PR #812, opened 2023-09-06
[PB-861] fix/inject react-tooltip css and decreased delay to 400ms

apsantiso:
If a version before the 5.13.0 is used, you need to import the css file manually so react-tooltip works as expected. We are using the 5.8.3 version and this step was missing, making the tooltip look glitchy.

image
source > https://react-tooltip.com/docs/getting-started

- Imported css file from react-tooltip library.
- Added noArrow prop. This arrow was not showing before because of the missing css file, this is needed to maintain the same style.
- Decreased delay to show the tooltip to 400ms.

https://github.com/internxt/drive-web/assets/143480783/937821d6-5b2a-4f09-98fc-d5e528e8b59f

### DIFF
--- a/src/app/shared/components/Tooltip/Tooltip.tsx
+++ b/src/app/shared/components/Tooltip/Tooltip.tsx
@@ -6,13 +6,14 @@ interface TooltipProps {
   delayShow?: number;
 }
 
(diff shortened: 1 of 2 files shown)


sg-gs (approved):
LGTM

PR #813, opened 2023-09-06
[_]: feat/hiding or displaying options depending on the role

rafijv:
- Reload list after manage users access
- Temporarily hide general access public option
- Hide general access options for users without permissions
- Hide invite button to viewers
- Temporarily Hide copy link button

### DIFF
--- a/src/app/drive/components/ShareDialog/ShareDialog.tsx
+++ b/src/app/drive/components/ShareDialog/ShareDialog.tsx
@@ -370,7 +370,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
                     </div>
                   </Button>
                 )}
-                {(currentUserFolderRole === 'owner' || currentUserFolderRole === 'editor') && (
+                {currentUserFolderRole != 'viewer' && (
                   <Button variant="secondary" onClick={onInviteUser}>
                     <UserPlus size={24} />
                     <span>{translate('modals.shareModal.list.invite')}</span>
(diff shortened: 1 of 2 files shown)


src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -108,6 +108,18 @@ export default function SharedView(): JSX.Element {
     }
   }, []);
 
+  useEffect(() => {
+    if (!isShowInvitationsOpen) fetchRootItems();
+  }, [isShowInvitationsOpen]);
```
CandelR: I think that this has been removed in the accept/decline invitations task.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -133,6 +134,11 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
     return () => clearTimeout(timer);
   }, [accessRequests]);
 
+  useEffect(() => {
+    const currentInvitedUser = invitedUsers.find((user) => user.email === props.user.email);
+    setCurrentUserFolderRole(currentInvitedUser?.roleName);
+  }, [invitedUsers]);
```
CandelR, line 140: When is no user invited, at the moment it not appear any user on the list. This causes that the owner couldn't invite users with this logic.

CandelR (changes requested):

src/app/share/views/SharedLinksView/SharedView.tsx
rafijv: Removed, was an error merging branches

src/app/drive/components/ShareDialog/ShareDialog.tsx
rafijv: Fixed

CandelR (approved):

PR #818, opened 2023-09-08
[PB-692]: fix/update subscription correctly

xabg2:
**ISSUE:**
When the user who has a paid plan upgrades his subscription to another that has a coupon code, the user pays the full price without a discount.

**HOW TO SOLVE IT:**
Now, we allow to update the subscription with a coupon code, if it exists so that only a part of it is charged.


sg-gs (approved):

sg-gs:
@masterprog-cmd says: _**Missing:** This PR must be merged before: https://github.com/internxt/sdk/pull/167_

PR #827, opened 2023-09-12
[_]: feat/add 3DS request

xabg2:
Ask the bank for approval when required when changing subscription (3DSecure problem)


xabg2:
We must update the SDK before deployment.

PR #828, opened 2023-09-13
[_]: Added redirection to accept or decline sharing folder invitation from link

CandelR:
- Accept or decline sharing folder invitation from URL

⚠️These SDK changes are needed
-  https://github.com/internxt/sdk/pull/171


src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -24,10 +24,17 @@ import TextInput from '../TextInput/TextInput';
...
 import { referralsThunks } from 'app/store/slices/referrals';
 import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
+import shareService from '../../../share/services/share.service';
+import notificationsService, { ToastType } from '../../../notifications/services/notifications.service';
 
 export default function LogIn(): JSX.Element {
   const { translate } = useTranslationContext();
   const urlParams = new URLSearchParams(window.location.search);
+
+  const sharingId = urlParams.get('sharingId');
```
sg-gs, line 34: What happens if some param is wrong / has the wrong format / is missing?

sg-gs (changes requested):

src/app/auth/components/LogIn/LogIn.tsx
CandelR: Nothing, it will not work. If some param is wrong it will show a error message informing that has not been possible to accept/decline the invitation to the shared folder

sg-gs (approved):

PR #830, opened 2023-09-14
[PB-954]: feat/allow users share files privately

rafijv:
- Allow users share files
- Add list of users with access to a file
- Allow to remove role on a file
- Allow owner to stop sharing a file
- Add files to the shared list


sg-gs (approved):
LGTM

PR #837, opened 2023-09-19
[PB-960]: fix/hide pending invitations button

rafijv:
- Display manage access action for files in shared view
- Hide invitation button if no invitations
- Fix reload shared view after manage access

### DIFF
--- a/src/app/share/views/SharedLinksView/SharedView.tsx
+++ b/src/app/share/views/SharedLinksView/SharedView.tsx
@@ -121,7 +121,14 @@ export default function SharedView(): JSX.Element {
 
   useEffect(() => {
     if (!currentFolderId && !isShareDialogOpen && !folderUUID) {
-      setTimeout(fetchRootFolders, 200);
+      setTimeout(() => {
+        setPage[0];
+        setShareItems([]);
+        setHasMoreRootFolders(true);
+        setHasMoreFolders(true);
+        setHasMoreItems(true);
+        fetchRootFolders();
+      }, 200);
     } else if (currentFolderId && !isShareDialogOpen && !folderUUID) {
       setTimeout(fetchFolders, 200);
     }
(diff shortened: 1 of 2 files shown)


src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -121,7 +121,14 @@ export default function SharedView(): JSX.Element {
 
   useEffect(() => {
     if (!currentFolderId && !isShareDialogOpen && !folderUUID) {
-      setTimeout(fetchRootFolders, 200);
+      setTimeout(() => {
+        setPage[0];
```
CandelR: The function is not being executed here. `setPage[0]` -> `setPage(0);`
In `onShowInvitationsModalClose` is the same error, change it in this PR please.

src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -121,7 +121,14 @@ export default function SharedView(): JSX.Element {
...
+        setShareItems([]);
+        setHasMoreRootFolders(true);
+        setHasMoreFolders(true);
+        setHasMoreItems(true);
+        fetchRootFolders();
```
CandelR: If all of these it is for reset the Shared view state, add it to a function and reuse it in all places that is need it.

CandelR (changes requested):

src/app/share/views/SharedLinksView/SharedView.tsx
rafijv: Done

src/app/share/views/SharedLinksView/SharedView.tsx
rafijv: Done

CandelR (approved):

PR #854, opened 2023-09-21
[PB-958]: fix/change shared item overlay icon

rafijv:
- Change shared item overlay icon


aboredvaro:
@CandelR @rafijv Guys I think we went back in time in the Drive list, looks like something happened during the merge and we took back the old Drive list ("Type" column should not appear, it should like very similar to the Shared list. Also the space between the file icon and the file name its from the old one)

CandelR (approved):

PR #862, opened 2023-09-22
[PB-980]: Fix/display photos usage

xabg2:
Display the photos usage in Settings (Usage section).

### DIFF
--- a/src/app/drive/services/usage.service.ts
+++ b/src/app/drive/services/usage.service.ts
@@ -25,7 +25,7 @@ async function getUsageDetails(): Promise<UsageDetailsProps> {
   const storageClient = SdkFactory.getInstance().createStorageClient();
   const photosClient = await SdkFactory.getInstance().createPhotosClient();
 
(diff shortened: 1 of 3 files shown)


src/app/drive/services/usage.service.ts
```diff
@@ -10,9 +10,7 @@ export async function fetchUsage(): Promise<UsageResponse> {
     photosClient.photos.getUsage(),
   ]);
 
-  driveUsage.total += photosUsage;
-
-  return driveUsage;
+  return { ...driveUsage, photos: photosUsage };
```
CandelR, line 13: This way not all the space used will be shown in the main Drive view as the space used for photos is not added up.
How should it be displayed where it was already displayed?

CandelR: This way not all the space used will be shown in the main Drive view as the space used for photos is not added up.
How should it be displayed where it was already displayed?

CandelR (changes requested):

CandelR (approved):

PR #864, opened 2023-09-25
[PB-1013]: Fix/enable download of files from sharings Root list

apsantiso:
1. Get owner credentials from response to download files and folders instead of extra request.

needs:

https://github.com/internxt/drive-server-wip/pull/214

### DIFF
--- a/src/app/share/views/SharedLinksView/SharedView.tsx
+++ b/src/app/share/views/SharedLinksView/SharedView.tsx
@@ -185,7 +185,6 @@ export default function SharedView(): JSX.Element {
         shareItem.isFolder = true;
         shareItem.isRootLink = true;
         shareItem.name = getItemPlainName(shareItem as unknown as DriveItemData);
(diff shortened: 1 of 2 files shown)


src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -633,8 +633,8 @@ export default function SharedView(): JSX.Element {
     try {
       if (shareItem.isRootLink) {
         const { credentials } = await shareService.getSharedFolderContent(
-          // folderUUID
-          shareItem.uuid,
+          // folderUUID = uuid. If it's a file, it comes as folderUuid.
+          shareItem.isFolder ? shareItem.uuid : shareItem.folderUuid,
```
sg-gs: We should refactor this rather than patching it with this misleading naming @Ap1402

CandelR (approved):

PR #865, opened 2023-09-25
[PB-1010]: bugfix/public sharing

CandelR:
- Added `encryptionCode` field to use always same code to decrypt public shared item
Need to merge sdk:
- https://github.com/internxt/sdk/pull/174

### DIFF
--- a/src/app/share/services/share.service.ts
+++ b/src/app/share/services/share.service.ts
@@ -376,21 +376,25 @@ export const getPublicShareLink = async (uuid: string, itemType: 'folder' | 'fil
   const encryptedMnemonic = aes.encrypt(mnemonic, code);
 
   try {
-    const publicSharingItemData = await shareService.createPublicSharingItem({
+    const publicSharingItemData = await createPublicSharingItem({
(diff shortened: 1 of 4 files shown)


src/app/share/services/share.service.ts
```diff
@@ -376,21 +376,25 @@ export const getPublicShareLink = async (uuid: string, itemType: 'folder' | 'fil
   const encryptedMnemonic = aes.encrypt(mnemonic, code);
 
   try {
-    const publicSharingItemData = await shareService.createPublicSharingItem({
+    const publicSharingItemData = await createPublicSharingItem({
       encryptionAlgorithm: 'inxt-v2',
       encryptionKey: encryptedMnemonic,
       itemType,
       itemId: uuid,
+      encryptedCode: code,
```
sg-gs: The code needs to be sent encrypted with the mnemonic. Please check how we were doing it on the old sharing and you will find the line where the code is encrypted with the mnemonic previously to be sent. @CandelR

sg-gs (changes requested):

sg-gs (approved):

PR #870, opened 2023-09-27
[PB-982] bugfix/display notification when max space used

CandelR:
- Display max space used dialog when backend returns max space used error
- Added new design for show max space used alert

### DIFF
--- a/src/app/network/UploadManager.ts
+++ b/src/app/network/UploadManager.ts
@@ -8,11 +8,15 @@ import errorService from '../core/services/error.service';
 import { ConnectionLostError } from './requests';
 import { t } from 'i18next';
 import analyticsService from '../analytics/services/analytics.service';
+import { uiActions } from '../store/slices/ui';
+import { Dispatch } from '@reduxjs/toolkit';
 
 const TWENTY_MEGABYTES = 20 * 1024 * 1024;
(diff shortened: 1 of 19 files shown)


src/app/network/UploadManager.ts
```diff
@@ -175,6 +179,8 @@ class UploadManager {
             next(null, driveFileDataWithNameParsed);
           })
           .catch((err) => {
+            console.log({ err });
```
sg-gs: Remove this `console.log`

src/app/network/UploadManager.ts
```diff
@@ -43,11 +46,12 @@ type UploadManagerFileParams = {
 
 export const uploadFileWithManager = (
   files: UploadManagerFileParams[],
+  dispatch: Dispatch,
```
sg-gs: We are coupling the upload manager seriously with other unrelated things. It would be better to add callbacks / event emitters to specific events, there is no need to generate more debt here.

sg-gs (changes requested):

src/app/network/UploadManager.ts
CandelR: You are right. I didn't think at the time to put a callback, I change it.

sg-gs (approved):

PR #873, opened 2023-09-29
[_]: Migrate Cypress to TS mode for drive web

PixoDev:
This PR adds support for Cypress using TS, it adds:

- A tsconfig.json file with the needed configuration
- Refactors for commands that can not benefit from TS suggestions
- Moves away from npm, uses yarn instead


xabg2 (approved):
We have to move the staticData.json to an .env and add this file to .gitignore.

PR #879, opened 2023-10-04
[PB-901]: fix/display rename folder menu option only to owners

rafijv:
- Edit function to identify owner of an item


CandelR (changes requested):
It has conflicts.
I didn't know there was a task and I added that fix in another task I did, if you want to compare and leave what you think is best.

rafijv:
> It has conflicts. I didn't know there was a task and I added that fix in another task I did, if you want to compare and leave what you think is best.

Your solution is fine, I remove the PR

PR #885, opened 2023-10-10
[_]: webpack upgrade

apsantiso:


CandelR:
Closed because we already merge the task https://github.com/internxt/drive-web/pull/887 that contains this changes

PR #890, opened 2023-10-18
[PB-798]: feat/update shared layout styles

rafijv:
Updated shared layout styles, icons and copy


CandelR (approved):
Good job with sonar cloud 🚀

PR #892, opened 2023-10-20
[PB-995] Feat/enable invite users public sharings

apsantiso:
Depends on:

1. https://github.com/internxt/sdk/pull/177
2. https://github.com/internxt/drive-server-wip/pull/218

Changes:

1. Sharing type is now fetched from the backend when a shareDialog is open.
2. Users can be invited to public shared items.
3. If an **_invited_** user clicks on copy link, it fetches the public entry from the backend and uses the owner's mnemonic to decrypt the code.
4. When the change access mode dropdown is changed, it lets the backend know it.


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -259,7 +268,11 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
     }
 
     if (itemToShare?.item.uuid) {
-      await shareService.getPublicShareLink(itemToShare?.item.uuid, itemToShare.item.isFolder ? 'folder' : 'file');
+      await shareService.getPublicShareLink(
+        itemToShare?.item.uuid,
+        itemToShare.item.isFolder ? 'folder' : 'file',
+        (itemToShare?.item as unknown as AdvancedSharedItem)?.encryptionKey,
```
CandelR: If it is not necessary, we will avoid this type of casting. 
Add the type to the sdk since you have made changes.

src/app/i18n/locales/en.json
```diff
@@ -656,7 +656,8 @@
       },
       "errors": {
         "updatingRole": "The role could not be updated, please try again later.",
-        "copy-to-clipboard": "Error sharing item, try again later."
+        "copy-to-clipboard": "Error sharing item, try again later.",
+        "update-sharing-acces": "The sharing mode could not be updated, please try again later"
```
CandelR: Add this translation to other languages

CandelR (commented):
Review the both code smells and it will be done, good job 🚀

src/app/drive/components/ShareDialog/ShareDialog.tsx
apsantiso: Hey This does not come from the SDk. I think is more like a type problem as typescript does not know if this is driveItem or advanced Share Item. Added a typeguard function as TS recommends and a small change in the type to prevent this. 

I might be wrong tho, sorry if so.

CandelR (approved):
Approved but not merge pending QA. @Ap1402

PR #896, opened 2023-10-23
[PB-748]: Feat/update tailwindcss

xabg2:
Update tailwindCSS version to the latest one to take advantage of colors and tokens for dark mode.


xabg2:
I have to check the UI (preview link - Vercel). But I think that's all.

PR #898, opened 2023-10-25
[PB-1087] fix:get plainname as name for globalsearch

apsantiso:
- Fixed downloading of an item in global search with its name instead of undefined


CandelR (approved):

CandelR (approved):
Approved but not merged until it passes QA @Ap1402

PR #899, opened 2023-10-25
[PB-1063] feature/add web workers to upload

CandelR:
- Added web workers to upload files


rafijv (approved):

sg-gs (changes requested):
Hey @CandelR,
The use of the worker is mixed with what the promotion is. You should try not to couple things, there is no need to generate that debt:

- If in the multipart upload you need to change the method that does the upload, it is preferable to use a callback, since it allows the case without Workers / with Workers. On the other hand, any nesting that you have to do to use the worker is somewhere else and not mixed with the logic of the upload.
- You anchor the upload directly to the worker usage, which could not be the case

It is more a question of the SRP and that each function has just one responsibility, but this functions to do a lot of things now. If we want to keep the code clear we need to split things. For example, you can inject how the data transfer is performed using a callback, so if the method is the current uploadBlob one or it is the worker-backed one, this codebase is not dependent on that and the callback hell is pilling up on these methods would not be there, at least

sg-gs (approved):

PR #902, opened 2023-10-26
[PB-916]: feat/add track events

rafijv:
Track events added:
- Shared items
- File preview
- Backup key download
- Password recovery


src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -162,17 +174,23 @@ const FileViewer = ({
   useEffect(() => {
     setIsPreviewAvailable(true);
     const largeFile = isLargeFile(file?.size);
-
+    trackFilePreviewClicked(trackFilePreviewProperties);
```
CandelR, line 175: I think that this place for these events is not the best, it is triggered every time the file is displayed or changed and not when the button is clicked to show the file preview.

rafijv: > I think that this place for these events is not the best, it is triggered every time the file is displayed or changed and not when the button is clicked to show the file preview.

I have been testing it and it only executes when it corresponds to the event requirements.

CandelR: Perhaps I have misunderstood. Doesn't it have to be launched when the preview button is pressed? This useEffect is executed when the value of show or file changes, even if it coincides with the click of the button, there may be a case in which this component is used in another part of the web that does not execute in that case. 
If this is not the case, maybe I'm wrong.

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -89,6 +95,12 @@ const FileViewer = ({
   const isEditNameDialogOpen = useAppSelector((state: RootState) => state.ui.isEditFolderNameDialog);
   const isShareItemSettingsDialogOpen = useAppSelector((state) => state.ui.isShareItemDialogOpenInPreviewView);
 
+  const trackFilePreviewProperties: TrackingPlan.FilePreviewProperties = {
+    file_size: file.size,
+    file_extension: file.type,
+    preview_id: file.uuid,
```
CandelR: This is crashing public shared file view.

src/app/drive/components/FileViewer/FileViewer.tsx
rafijv: This event is created to detect possible extensions that do not have preview and in case it is necessary to implement it. I understand that it must be executed whenever a file is previewed regardless of how it is accessed.

src/app/drive/components/FileViewer/FileViewer.tsx
rafijv: Fixed

CandelR (approved):
Approved! Waiting the QA approve to merge it @rafijv

PR #907, opened 2023-11-07
[_]: fix/impact

xabg2:
Tracks are now sent directly to the Impact API bypassing RudderStack.


src/app/analytics/services/analytics.service.ts
```diff
@@ -148,23 +155,23 @@ export function signInAttempted(email: string, error: string | Error): void {
...
+
+  source !== 'direct' &&
+    axios
+      .post(IMPACT_API, {
+        anonymousId: anonymousID,
+        timestamp: dayjs().format('YYYY-MM-DDTHH:mm:ss.sssZ'),
+        messageId: '',
+        userId: uuid,
+        type: 'track',
+        event: 'USer Signup',
```
sg-gs: It should be 'User Signup', carefully with the first S, is in uppercase

src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -138,12 +138,6 @@ export default function LogIn(): JSX.Element {
         window.rudderanalytics.identify(user.uuid, { email: user.email, uuid: user.uuid });
         window.rudderanalytics.track('User Signin', { email: user.email });
 
-        // analyticsService.identify(user, user.email);
```
sg-gs, line 141: Why are we removing this?

sg-gs (changes requested):

src/app/auth/components/LogIn/LogIn.tsx
xabg2: It was commented. Now, I guess we use:
`window.rudderanalytics.identify(user.uuid, { email: user.email, uuid: user.uuid });
        window.rudderanalytics.track('User Signin', { email: user.email });` (line 138/139)

sg-gs (approved):

PR #911, opened 2023-11-08
[PB-1085 , PB-1084]:feature/handle not registered users invitation

apsantiso:
Requires:  https://github.com/internxt/sdk/pull/180


src/app/auth/components/SignUp/useSignUp.ts
```diff
@@ -110,5 +132,54 @@ export function useSignUp(registerSource: 'activate' | 'appsumo', referrer?: str
...
 
-  return { updateInfo, doRegister };
+  const doRegisterPreCreatedUser = async (email: string, password: string, invitationId: string, captcha: string) => {
+    const authClient = SdkFactory.getNewApiInstance().createAuthClient();
+
+    const registerDetails = await generateRegisterDetails(email, password, captcha);
+
+    const data = await authClient.registerPreCreatedUser({ ...registerDetails, invitationId });
+    const { token } = data;
+    const user: UserSettings = data.user as unknown as UserSettings;
```
CandelR: Avoid using "as unknown" as much as possible as we may cause unexpected errors if something of the type being matched does not match.

src/app/auth/services/user.service.ts
```diff
@@ -22,6 +22,11 @@ const inviteAFriend = (email: string): Promise<void> => {
   return usersClient.sendInvitation(email);
 };
 
+const preCreateUser = (email: string): Promise<{ publicKey: string; user: { email: string; uuid: string } }> => {
```
CandelR: The return type of this is in the sdk? If it is use it please

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -82,185 +97,214 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
...
+  const getUserPublicKey = async (email: string): Promise<string> => {
+    let publicKey = '';
+    try {
+      const publicKeyResponse = await userService.getPublicKeyByEmail(email);
+      publicKey = publicKeyResponse.publicKey;
+    } catch (error) {
+      if (error instanceof Error) {
+        const errorBody = JSON.parse(error.message);
+        if (errorBody.statusCode !== 404) {
+          throw error;
```
CandelR: If the error is different of 404 the propagation will make the app crash.
Handle correctly all the possible errors and add the report service.

src/app/store/slices/sharedLinks/index.ts
```diff
@@ -202,12 +204,21 @@ const shareItemWithUser = createAsyncThunk<string | void, ShareFileWithUserPaylo
...
+
+      if (payload.isNewUser) {
+        const prCreatedUserResponse = await userService.preCreateUser(payload.sharedWith);
+        publicKey = prCreatedUserResponse.publicKey;
+      }
 
       const encryptedMnemonic = await encryptMessageWithPublicKey({
         message: mnemonic,
-        publicKeyInBase64: publicKey,
+        publicKeyInBase64: publicKey ?? '',
```
CandelR: I think that public key always has to have value. Isn't it?

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -82,185 +97,214 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
...
-      usersToInvite.forEach((user) => {
-        const userRoleId = props.roles.find((role) => role.name === user.userRole)?.id;
-        if (!userRoleId) return;
+    await Promise.all(sharingPromises);
+    trackRestrictedShared(trackingRestrictedSharedProperties);
+  };
+
+  //TODO: EXTRACT THIS LOGIC OUT OF THE DIALOG
+  const onInvite = async ({ inviteNewUsers = false }) => {
+    const preCreateUsers = inviteNewUsers;
```
CandelR: If you have to rename a boolean variable that is only used here, maybe the name given to it is not the best one. Perhaps renaming it to something that fits better would be a better option than assigning it to another const with another name.

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -82,185 +97,214 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
...
+    const usersList = [...usersToInvite];
+    let isThereAnyNewUser = newUsersExists;
```
CandelR, line 155: Same here, maybe we should rename the name of the state.

src/app/i18n/locales/en.json
```diff
@@ -744,6 +744,10 @@
         "successSentInvitation": "Successful sharing with {{email}}",
         "acceptedSuccessfully": "Shared folder invitation successfully accepted",
         "declinedSuccessfully": "Shared folder invitation successfully declined",
+        "inviteNewUsersBody": "People without Internxt account will be sent a signup invitation email",
+        "inviteNewUsersTitle": "Invite new Users?",
+        "accept": "Accept",
+        "cancel": "Cancel",
```
CandelR, line 755: Please add the translation into other languages

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
```diff
@@ -0,0 +1,306 @@
...
+      errorService.reportError(err);
+      const castedError = errorService.castError(err);
+      setSignupError(castedError.message);
+    } finally {
+      setShowError(true);
+    }
+  };
+
+  if (invitationValidation.isLoading) {
+    return <></>;
```
CandelR, line 200: Any reason not to add a charging element, such as a spinner or any kind of element that indicates that it is loading?

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
```diff
@@ -0,0 +1,306 @@
...
+        label:
+          result.reason === 'NOT_COMPLEX_ENOUGH'
+            ? 'Password is not complex enough'
+            : 'Password has to be at least 8 characters long',
+      });
+    } else if (result.strength === 'medium') {
+      setPasswordState({ tag: 'warning', label: 'Password is weak' });
+    } else {
+      setPasswordState({ tag: 'success', label: 'Password is strong' });
+    }
```
CandelR, line 148: Could you add a TODO in order to remember extract this? It is repeated in a few components.

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
```diff
@@ -0,0 +1,306 @@
...
+      await dispatch(userThunks.initializeUserThunk());
+      dispatch(productsThunks.initializeThunk());
+      dispatch(planThunks.initializeThunk());
+
+      dispatch(referralsThunks.initializeThunk());
+
+      window.rudderanalytics.identify(xUser.uuid, { email, uuid: xUser.uuid });
+      window.rudderanalytics.track('User Signup', { email });
+
+      return navigationService.push(AppView.Shared);
```
CandelR, line 188: Just to verify the flow. After successfull sign up it has to redirect the user to Shared view?

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
```diff
@@ -0,0 +1,306 @@
...
+              <label className="space-y-0.5">
+                <span>{translate('auth.email')}</span>
+                <TextInput
+                  placeholder={translate('auth.email')}
+                  label="email"
+                  type="email"
+                  disabled={hasEmailParam}
+                  register={register}
+                  required={true}
+                  minLength={{ value: 1, message: 'Email must not be empty' }}
```
CandelR: Extract this message to a translation

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
apsantiso: Yes

src/app/auth/components/SignUp/useSignUp.ts
apsantiso: This flow is almost 100% exactly to the register flow and view, types and other things returned from the SDK are also the same. It is the same flow so I just copied it and extracted the `registrationDetails` so we can use that function in doRegister too. 

I'm going to try to remove the unknown, but probably it is there for a reason as the same type is used in doRegister.

src/app/share/views/SharedGuestSignUp/ShareGuestSingUpView.tsx
apsantiso: I created a thread in slack mentioning we needed this like 2 weeks ago, and one of the possible options discussed was to show an empty page while this verification is being made as the backend returns a quick response. 

If you want another component to be shown, please go ahead and let me know.

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
apsantiso: No, this `let` is going to change in this function flow and it is essencial to cover the two possible invitation cases we have, I prefer to not set any state here  to prevent any rerendering while this is being executed (Maybe now that we are using React 18 the possibility of rendering problems is nonexistent tho)

```
   if (usersList.length === 0 && isValidEmail(email)) {
      const publicKey = await getUserPublicKey(email);
      if (!publicKey && !preCreateUsers) {
        isThereAnyNewUser = true;
      }
      usersList.push({
        email,
        userRole,
        isNewUser: !publicKey,
        publicKey,
      });
    }
```

src/app/auth/components/SignUp/useSignUp.ts
apsantiso: Done. It seems that we can fix the same thing in the normal signup by just changing the type from the SDK. When you register an user, the type returned is Omit<UserSettings, "bucket"> & { referralCode: string} so that's the cause of the "unknown" there as well. 

The backend is returning the bucketId so there is no need for any "Omit" there, by just removing that omit, we can cast easily this.

CandelR (approved):
Release the sdk so that the preview can be tested 👌 
Good job @Ap1402 🚀

CandelR (changes requested):
The redirection after clicking on the link in the email does not seem to work @Ap1402

CandelR (approved):
Approved, not merged, waiting to change the environment variable in server-wip to check that the email is sent correctly

PR #922, opened 2023-11-16
[PB-1044]: feat/view item details

xabg2:
Users can now see the metadata of their files/folders. 

The data the users can see are:
* Name
* Size
* Type
* Updated
* Modified
* Uploaded by
* Shared
* Location


sg-gs (approved):
Added my review as a crypto-related file is affected @CandelR. Still review it yourself carefully.

src/app/drive/components/DriveExplorer/DriveExplorerList/DriveExplorerList.tsx
```diff
@@ -202,6 +202,19 @@ const DriveExplorerList: React.FC<DriveExplorerListProps> = memo((props) => {
     [dispatch, uiActions],
   );
 
+  const getDetails = useCallback(
+    (item: DriveItemData) => {
+      const itemDetails: DriveItemDetails = {
+        ...item,
+        isShared: (item.sharings && item.sharings.length > 0) ?? false,
```
CandelR: if items is not > 0 it returns `false` already, is not necessary the nullish operator:
`!!(item.sharings && item.sharings.length)`
Perhaps the name of the function could be `showDetails`, as it is to show the modal details of the item

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+          setIsLoading(false);
+        })
+        .catch((err) => {
+          const error = errorService.castError(err);
+          errorService.reportError(error);
+        });
+    }
+  }, [item, isOpen]);
+
+  function onClose() {
```
CandelR: Have you used traditional functions instead of arrow functions for some reason?
If not, try to use arrow functions to maintain consistency in the project code

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+ * - Shared
+ * - Size (only for files)
+ * - Type (only for files)
+ * - Uploaded
+ * - Modified
+ * - Uploaded by
+ * - Location
+ *  */
+
+const ItemDetailsDialog = ({ onSharedFolderClicked }: { onSharedFolderClicked?: (item: any) => void }) => {
```
CandelR: Avoid the use of `any` as far as possible

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const user = localStorageService.getUser();
+  const { onNameClicked } = useDriveItemActions(item as DriveItemData);
+  const isFolder = item?.isFolder;
+
+  useEffect(() => {
+    if (isOpen && item && user) {
+      setIsLoading(true);
+      const isShared = item.isShared ? translate('actions.yes') : translate('actions.no');
+      const uploaded = formateDate(item.createdAt);
+      const modified = formateDate(item.updatedAt);
```
CandelR, line 93: `uploadedDate` and `modifiedDate` names gives more context

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
```diff
@@ -0,0 +1,198 @@
...
+      getDetailsData(item, isShared, uploaded, modified, user.email)
+        .then((details) => {
+          setItemProps(details);
+          setIsLoading(false);
+        })
+        .catch((err) => {
+          const error = errorService.castError(err);
+          errorService.reportError(error);
+        });
```
CandelR, line 102: If something during `getDetailsData` execution fails, this catch never be executed because already exists a try catch block inside the function. 
Just to keep in mind, the behaviour if something goes wrong will be that the detail block is not displayed, is that behaviour correct?

src/app/share/views/SharedLinksView/SharedLinksView.tsx
```diff
@@ -211,6 +211,18 @@ export default function SharedLinksView(): JSX.Element {
...
+    (item) => {
+      const itemDetails = {
+        ...item,
+        isShared: true,
+      };
+      dispatch(uiActions.setItemDetailsItem(itemDetails));
+      dispatch(uiActions.setIsItemDetailsDialogOpen(true));
+    },
+    [dispatch, uiActions],
+  );
```
CandelR: This function is repeated with some change, maybe it would be a good idea to try to extract it.

src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -1040,3 +1068,21 @@ export default function SharedView(): JSX.Element {
...
+    state.ui.isShareDialogOpen ||
+    state.ui.isSurveyDialogOpen ||
+    state.ui.isEditFolderNameDialog ||
+    state.ui.isFileViewerOpen ||
+    state.ui.isMoveItemsDialogOpen ||
+    state.ui.isCreateFolderDialogOpen ||
+    state.ui.isNameCollisionDialogOpen ||
+    state.ui.isReachedPlanLimitDialogOpen ||
+    state.ui.isItemDetailsDialogOpen,
+}))(SharedView);
```
CandelR, line 1088: 👌 🚀

src/app/shared/components/Breadcrumbs/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -178,6 +177,16 @@ const BreadcrumbsItem = (props: BreadcrumbsItemProps): JSX.Element => {
     dispatch(storageThunks.downloadItemsThunk(currentFolder));
   };
 
+  const onDetailsItemButtonClicked = async () => {
+    const itemDetails: DriveItemDetails = {
+      ...currentFolder[0],
+      isShared: true,
```
CandelR: Why this `isShared` prop is always `true`?

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
xabg2: I use uploaded and modified names because I use the object keys for the render as well.
Link to Figma to get more context: [Item Details](https://www.figma.com/file/62hpKnJYlxUN77HUECXAlF/Drive-Web?type=design&node-id=8113-370&mode=design&t=H1QAcVnEY3aa5eLj-11)

Maybe I need to leave a comment explaining it?

src/app/shared/components/Breadcrumbs/BreadcrumbsItem/BreadcrumbsItem.tsx
xabg2: My mistake. It should be false for BreadcrubsItems since we don't have the dropdown in Shared View.

src/app/drive/components/ItemDetailsDialog/ItemDetailsDialog.tsx
xabg2: I use `any` instead of `item as unknown as AdvancedSharedItem` when passing the item through the function, since the item we pass does not have the same attributes as the object expected in the function.

CandelR (approved):
Good job @masterprog-cmd 🥳 
Waiting to merge it when it passes QA.

PR #924, opened 2023-11-20
[PB-1267] - feature/No Private access view

CandelR:
- Added flow to redirect users when they access a private folder without permission


src/app/share/views/SharedLinksView/SharedView.tsx
```diff
@@ -493,16 +495,18 @@ export default function SharedView(): JSX.Element {
...
         })
         .finally(() => {
-          const currentURL = history.location.pathname;
-          const nuevaURL = currentURL.replace(/folderuuid=valor&?/, '');
-          history.replace(nuevaURL);
+          if (hasTriggeredError && folderUUID) {
+            navigationService.push(AppView.RequestAccess, { folderuuid: folderUUID });
+          } else {
+            const currentURL = history.location.pathname;
+            const nuevaURL = currentURL.replace(/folderuuid=valor&?/, '');
```
xabg2: newURL could be better @CandelR

src/app/drive/views/RequestAccess/RequestAccess.tsx
```diff
@@ -7,9 +7,14 @@ import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
 import Button from 'app/shared/components/Button/Button';
 import bigLogo from 'assets/icons/big-logo.svg';
 import './RequestAccess.scss';
+import { logOut } from '../../../auth/services/auth.service';
```
xabg2, line 11: Maybe we can extract this: `const alertUsersTextareaMaxLenght = 950;` as `const TEXTAREA_MAX_LENGTH = 950;` or something like that here (outside the function).

src/app/share/views/SharedLinksView/SharedView.tsx
CandelR: Wow, totally right :)

xabg2 (approved):

PR #926, opened 2023-11-22
[PB-472]: feat/navigate files and folders with url

rafijv:
- The Owner of the file or folder can navigate directly with the url
- Navigate in drive using browsing history with forward and backward between files/folders
- The breadcrumb is updated correctly when navigating forward and backward
- The Url is updated correctly when navigating


src/App.tsx
```diff
@@ -101,9 +101,9 @@ class App extends Component<AppProps> {
 
       console.log(castedError.message);
```
CandelR: Change this console.log to report the error with the `errorService`

src/app/drive/components/DriveExplorer/DriveExplorerItem/hooks/useDriveItemActions.tsx
```diff
@@ -189,11 +192,18 @@ const useDriveItemActions = (item: DriveItemData): DriveItemActions => {
   };
 
   const onItemDoubleClicked = (): void => {
+    const pathnameSplit = pathname.split('/');
+    const isRecentsView = pathnameSplit[2] === 'recents';
+
     if (item.isFolder) {
-      dispatch(storageThunks.goToFolderThunk({ name: item.name, id: item.id }));
+      history.push(`/app/folder/${item.uuid}`);
```
CandelR: Extracts navigation URLS for standardisation and ease of maintenance.

src/app/drive/components/DriveExplorer/DriveExplorerItem/hooks/useDriveItemActions.tsx
```diff
@@ -189,11 +192,18 @@ const useDriveItemActions = (item: DriveItemData): DriveItemActions => {
...
     } else {
-      dispatch(uiActions.setIsFileViewerOpen(true));
-      dispatch(uiActions.setFileViewerItem(item));
+      if (isRecentsView) {
+        dispatch(uiActions.setIsFileViewerOpen(true));
+        dispatch(uiActions.setFileViewerItem(item));
+      } else {
+        history.push(`/app/file/${item.uuid}`);
```
CandelR: Extract this also

src/App.tsx
```diff
@@ -120,88 +120,97 @@ class App extends Component<AppProps> {
...
-        window.rudderanalytics.page(PATH_NAMES[window.location.pathname]);
-        serverPage(PATH_NAMES[window.location.pathname]).catch(() => {
-          // NO OP
-        });
-      }
-    }
+  const onCloseFileViewer = () => {
+    const pathname = navigationService.history.location.pathname.split('/');
+    const isRecentsView = pathname[2] === 'recents';
+    const isSharedView = pathname[2] === 'shared';
```
CandelR: I think would be better to make a function that checks if it is the path that you want to check. In that way we can do this check in every part of the project using that function.

src/App.tsx
```diff
@@ -120,88 +120,97 @@ class App extends Component<AppProps> {
...
+
+          <Toaster position="bottom-center" />
+
+          <NewsletterDialog isOpen={isNewsletterDialogOpen} />
+          {isSurveyDialogOpen && <SurveyDialog isOpen={isSurveyDialogOpen} />}
+
+          {isFileViewerOpen && fileViewerItem && (
+            <FileViewerWrapper
+              file={fileViewerItem}
+              onClose={() => onCloseFileViewer()}
```
CandelR: nit: This way will create an anonymous function in every render. If it is no necessary, would be better in performance only pass the function `onClose={onCloseFileViewer}`

src/app/core/config/app.json
```diff
@@ -120,6 +120,13 @@
       "auth": true,
       "hideSearch": true
     },
+    {
+      "id": "drive-items",
+      "layout": "header-and-sidenav",
+      "path": "/app/:folderType?/:id?",
```
CandelR: it is `folderType` or `itemType`?

src/app/drive/views/DriveView/DriveView.tsx
```diff
@@ -21,28 +24,64 @@ export interface DriveViewProps {
...
+      errorService.reportError(error);
+    }
+  };
+
+  const goFile = async (folderUuid) => {
+    const storageClient = SdkFactory.getNewApiInstance().createNewStorageClient();
+    const [responsePromise] = storageClient.getFile(folderUuid);
+
+    try {
+      const fileMeta = await responsePromise;
```
CandelR: Add this into a service

src/App.tsx

src/app/drive/components/DriveExplorer/DriveExplorerItem/hooks/useDriveItemActions.tsx
rafijv: Done

src/app/drive/components/DriveExplorer/DriveExplorerItem/hooks/useDriveItemActions.tsx
rafijv: Done

src/App.tsx

src/App.tsx

src/app/core/config/app.json
rafijv: Done

src/app/drive/views/DriveView/DriveView.tsx

src/App.tsx
```diff
@@ -47,40 +47,40 @@ interface AppProps {
...
-
-    if ((token && skipSignupIfLoggedIn) || (token && navigationService.history.location.pathname !== '/new')) {
-      /**
-       * In case we receive a valid redirectUrl param, we return to that URL with the current token
-       */
-      const redirectUrl = authService.getRedirectUrl(params, token);
-
-      if (redirectUrl) {
-        window.location.replace(redirectUrl);
-        return;
```
CandelR, line 68: Removing this return it will execute the code after the redirect if I'm not mistaken.

src/app/core/services/navigation.service.ts
```diff
@@ -44,6 +41,22 @@ const navigationService = {
...
+  },
+  pushFile(uuid: string | undefined): void {
+    instance.push(`/file/${uuid}`);
+  },
+  isCurrentPath(path: string): boolean {
+    const pathname = navigationService.history.location.pathname.split('/');
+    const currentPath = pathname[1];
+    return currentPath === path;
+  },
+  getUuid(): string {
```
CandelR: This could return `undefined` also in case it is used and `pathname[2]` has no value

src/app/drive/services/file.service/index.ts
```diff
@@ -94,12 +94,20 @@ async function fetchDeleted(): Promise<DriveFileData[]> {
   });
 }
 
+export function getFile(uuid: string) {
```
CandelR: Remember to add the return type

src/App.tsx
rafijv: This block of code has not been removed, it has just been moved but github shows it as removed and below as new code.

src/app/core/services/navigation.service.ts
rafijv: That way of using DriveView.tsx
```
    const itemUuid = navigationService.getUuid();

    if (isFolderView && itemUuid) {
```

src/app/drive/services/file.service/index.ts
rafijv: Done

src/App.tsx
CandelR: But the early `return;` after the redirection it is not in the code as far I can see. Could you point me where it is?

src/app/core/services/navigation.service.ts
CandelR: That's an external implementation of the function that does a check. If someone uses the method in the future without checking it, it is possible to get a `undefined`.

rafijv: Done

CandelR (approved):

PR #942, opened 2023-11-28
[MKT-205]: feat/christmas campaign banner

xabg2:
The banner for the christmas campaign has been added.


xabg2:
⚠️  Merge on Dec 11.
**Missing:**
* Background image.

src/app/banners/BlackFridayBanner.tsx
```diff
@@ -31,8 +32,9 @@ const BlackFridayBanner = ({ showBanner, onClose }: { showBanner: boolean; onClo
          absolute bottom-0 left-0 right-0 top-0 z-10 bg-black bg-opacity-40`}
     >
       <div
+        style={{ backgroundImage: `url(${ChristmasBG})`, backgroundSize: 'cover', backgroundPosition: 'center' }}
```
CandelR, line 35: If this gonna be a generic banner change the name to Banner instead of BlackFridayBanner, so every time we have to reuse it we know this is the banner element. 
The same with translations.

CandelR (changes requested):

src/app/banners/BlackFridayBanner.tsx
xabg2: It is just the banner for christmas/black-friday offers. We have another banner for lifetime and so on because we can reuse it without creating it again.

CandelR (approved):

PR #934, opened 2023-11-26
[PB-968]: fix/rename item in preview view

xabg2:
Extract all dropdown/item actions to a custom hook. Some functions name has also been fixed.

Now, TopBarActions uses the same dropdown as the item uses in the Drive Explorer.

PR #943, opened 2023-11-28
[PB-1278]: fix/remove duplicated edit name dialog

xabg2:
One of the item renaming dialogs has been removed and now, we use only one dialog for all views.

### DIFF
--- a/src/app/drive/components/FileViewer/FileViewerWrapper.tsx
+++ b/src/app/drive/components/FileViewer/FileViewerWrapper.tsx
@@ -80,7 +80,7 @@ const FileViewerWrapper = ({ file, onClose, showPreview }: FileViewerWrapperProp
     onRenameItemButtonClicked,
     onRestoreItemButtonClicked,
     onDeletePermanentlyButtonClicked,
-  } = useDriveItemActions(file);
+  } = useDriveItemActions(currentFile);
 
   const isCurrentUserViewer = useCallback(() => {
     return currentUserRole === UserRoles.Reader;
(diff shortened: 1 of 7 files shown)


xabg2:
This PR is part of this one: https://github.com/internxt/drive-web/pull/934

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -67,7 +67,7 @@ const FileViewerWrapper = ({ file, onClose, showPreview }: FileViewerWrapperProp
   const recentsActions = pathId === 'recents';
   const sharedActions = pathId === 'shared';
   const trashActions = pathId === 'trash';
-  const isSharedItem = file.sharings && file.sharings?.length > 0;
+  const isSharedItem = sharedActions;
```
CandelR: Why don't you rename the other const instead of referencing it with another name? It's the same value, maybe that means that the name it has is not the best.

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -155,7 +154,11 @@ const FileViewerWrapper = ({ file, onClose, showPreview }: FileViewerWrapperProp
   useEffect(() => {
     setBlob(null);
     if (dirtyName) {
-      setCurrentFile?.(currentItemsFolder?.find((item) => item.name === dirtyName) as DriveFileData);
+      // setCurrentFile?.(currentItemsFolder?.find((item) => item.name === dirtyName) as DriveFileData);
```
CandelR: You forgot to delete this code

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
xabg2: That's right.

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
xabg2: 😅

CandelR (approved):

PR #944, opened 2023-11-29
[PB-1348]: fix/hide list header

xabg2:
Remove "Name", "Owner", "Size" and "Create" for empty state in Shared View.


src/app/shared/components/List/index.tsx
```diff
@@ -211,47 +211,47 @@ ListProps<T, F>): JSX.Element {
...
+              <div className="flex h-full flex-row items-center justify-between pr-4">
+                <BaseCheckbox
+                  checked={selectedItems.length > 0}
+                  indeterminate={items.length > selectedItems.length && selectedItems.length > 0}
+                  onClick={onTopSelectionCheckboxClick}
+                />
+              </div>
+
+              {header.map((column) => (
+                <div
```
CandelR: I think this is not the best way to hide the header in case you want to show only the empty state. 
It makes the code more cluttered and harder to distinguish between the header and the body of the list. 
If you want to use the same ternary to show the empty or not, I think it is a better alternative to remove the Header and Body to separate components (not necessarily in another file, they can be kept in this one) and put them together there, so each element remains separate and makes maintenance easier.

If you can think of another way that you consider good, that's fine by me.

src/app/shared/components/List/index.tsx
```diff
@@ -283,14 +283,14 @@ ListProps<T, F>): JSX.Element {
...
             </InfiniteScroll>
-          </>
-        ) : (
-          <>{loader}</>
-        )}
-
-        {/* Click outside of the list to unselect all items */}
-        {items.length > 0 && <div className="h-full w-full py-6" onClick={unselectAllItems} />}
-      </div>
+            {items.length > 0 && <div className="h-full w-full py-6" onClick={unselectAllItems} />}
```
CandelR: What is the reason for moving this? Don't forget to move the comment that refers to it.

CandelR (changes requested):

src/app/shared/components/List/index.tsx
xabg2: We can separate the header and body into different components, I agree with you.

CandelR: Drive, Recents and Trash are already using the same view component, the `DriveExplorer` one that uses this list.

src/app/shared/components/List/index.tsx
```diff
@@ -210,48 +258,23 @@ ListProps<T, F>): JSX.Element {
...
+              selectedItems={selectedItems}
+              onTopSelectionCheckboxClick={onTopSelectionCheckboxClick}
+              items={items}
+              header={header}
+              orderBy={orderBy}
+              onOrderableColumnClicked={onOrderableColumnClicked}
+              menu={menu}
+              isVerticalScrollbarVisible={isVerticalScrollbarVisible}
+            />
+
```
CandelR: In this way the header it will be scrolled. It has to stay fixed.
I think that we have to add the conditional to the Header component outside the Body

CandelR (changes requested):

CandelR (approved):

PR #956, opened 2023-12-11
[PB-1372]: fix/user can't rename an item by pressing "R" in preview mode.

xabg2:
User can't rename an item by pressing "R" in preview mode.


src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -144,6 +144,14 @@ const FileViewerWrapper = ({ file, onClose, showPreview }: FileViewerWrapperProp
     });
   };
 
+  const renameItemFromKeyboard = () => {
+    if (isSharedView) {
+      if (!isCurrentUserViewer()) {
```
CandelR: Unify this conditional

CandelR (approved):

PR #959, opened 2023-12-11
[PB-1210]: fix/context menu options aren't completely displayed

xabg2:
The context menu has been fixed. 

For now the context menu opens downwards until it does not fit, then we show it upwards. With this we try to prevent as much as possible that the context menu is cut off.

### DIFF
--- a/src/app/shared/components/List/ListItem.tsx
+++ b/src/app/shared/components/List/ListItem.tsx
@@ -205,14 +205,14 @@ export default function ListItem<T extends { id: string }>({
 
             function handleOpenPosition() {
               const element = menuButtonRef.current;
-              const contextMenu = menuItemsRef.current;
+              const contextMenu = menuItemsRef?.current?.offsetHeight || 300;
               if (!element) return;
               if (!contextMenu) return;
 
               const { bottom } = element.getBoundingClientRect();
               const windowHeight = window.innerHeight;
 
-              const isHalfway = bottom > contextMenu.offsetHeight / 2 + windowHeight / 2.3;
+              const isHalfway = bottom + contextMenu > windowHeight;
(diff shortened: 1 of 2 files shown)


src/app/shared/components/List/ListItem.tsx
```diff
@@ -199,12 +205,14 @@ export default function ListItem<T extends { id: string }>({
 
             function handleOpenPosition() {
               const element = menuButtonRef.current;
+              const contextMenu = menuItemsRef?.current?.offsetHeight || 300;
```
CandelR: this const has the `contextMenuHeight`, isn't it?

src/app/shared/components/List/ListItem.tsx
```diff
@@ -199,12 +205,14 @@ export default function ListItem<T extends { id: string }>({
...
               if (!element) return;
+              if (!contextMenu) return;
 
               const { bottom } = element.getBoundingClientRect();
               const windowHeight = window.innerHeight;
 
-              const isHalfway = bottom > windowHeight / 2;
+              const isHalfway = bottom + contextMenu > windowHeight;
               setIsHalfwayDown(isHalfway);
```
CandelR: I think this no longer represents whether it is below the middle of the screen or not, better to update the state name as well to avoid confusion.
Now it seems that it always opens downwards as long as it fits, no?

CandelR (changes requested):
As the main "bug" in the task has not been fixed due to technical limitations, please update the description more specifically as to what the changes made change.

src/app/shared/components/List/ListItem.tsx
xabg2: Yep, I can update the name to a better one.

CandelR (approved):

PR #962, opened 2023-12-13
[PB-1423]: fix/restore items from trash

xabg2:
When the user restores an item from the recycle garbage can, the recycle garbage can appears empty even if the user has more items in the recycle garbage can.

Now, only the restored items are deleted and the other items still appear.

Also, the move button has been disabled when the user tries to move the item to the same folder where it is located.

### DIFF
--- a/src/app/drive/components/DriveExplorer/DriveExplorer.tsx
+++ b/src/app/drive/components/DriveExplorer/DriveExplorer.tsx
@@ -575,7 +575,8 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
       <MoveItemsDialog
         items={[...items]}
         onItemsMoved={() => {
-          !isTrash && dispatch(fetchSortedFolderContentThunk(currentFolderId));
+          dispatch(fetchSortedFolderContentThunk(currentFolderId));
+          onItemsMoved?.();
         }}
         isTrash={isTrash}
       />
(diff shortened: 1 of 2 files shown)


src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -163,11 +165,9 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
         await restoreItemsFromTrash(itemsToMove, destinationFolderId, translate as TFunction, props.isTrash);
       }
 
-      props.onItemsMoved?.();
```
CandelR: Don't remove this, this will cause it not to execute what is being passed through this prop.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -575,8 +575,7 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
       <MoveItemsDialog
         items={[...items]}
         onItemsMoved={() => {
-          dispatch(fetchSortedFolderContentThunk(currentFolderId));
-          onItemsMoved?.();
+          !isTrash && dispatch(fetchSortedFolderContentThunk(currentFolderId));
```
CandelR: I would say that this does not affect trash items even if you are in trash, in fact, it is necessary to update Drive so that when the user comes back they will find the folder updated.

CandelR (changes requested):

CandelR (approved):

PR #964, opened 2023-12-14
fix: added loading state to shareInviteDialog invite button

apsantiso:


src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -288,7 +293,7 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
               <BaseCheckbox checked={notifyUser} />
               <p className="ml-2 text-base font-medium">{translate('modals.shareModal.invite.notifyUsers')}</p>
             </div>
-            <Button variant="primary" onClick={onInvite} disabled={isInviteButtonDisabled}>
+            <Button variant="primary" onClick={onInvite} disabled={isInviteButtonDisabled} loading={isAnyInviteLoading}>
```
CandelR, line 296: I think we should deactivate the button while the invitation is being processed.

CandelR (changes requested):

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
apsantiso: It disables if "loading" or "disabled" is true
image

CandelR (approved):

PR #970, opened 2023-12-18
[PB-1377] [PB-1381] Feat/password protected sharings

apsantiso:
Requirements are already merged. Feel free to merge if needed.

Requires:

https://github.com/internxt/sdk/pull/182 (merged)

Backend changes:
https://github.com/internxt/drive-server-wip/pull/247 (merged)

Remember to run migrations to add the "encryptedPassword" field to your local backend if you are testing locally.


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -136,7 +159,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
     if (roles.length === 0) dispatch(sharedThunks.getSharedFolderRoles());
 
     if (roles.length > 0) loadShareInfo();
```
apsantiso: Sharing info is not being loaded when you access to the modal from Drive. This fixes that, let me know if there is a better way.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -175,31 +198,33 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
   }, [itemToShare, roles]);
 
   const loadShareInfo = async () => {
+    if (!itemToShare?.item) return;
+
+    const isItemNotSharedYet = !isAdvanchedShareItem(itemToShare?.item) && !itemToShare.item.sharings?.length;
+    if (isItemNotSharedYet) return;
+
```
apsantiso: Avoids unnecesary calls to the endpoint that returns the sharing data and sharing type. There is nothing to check here as the item is not shared yet.

Types here are kind of problematic here, they don't align with the current backend response and are heavily nested. So I used the typeguard `isAdvanchedShareItem` to be able to access to "item.sharings". 

Current item fetched from backend:

`{
    "id": 2013,
     ..... , 
    "shares": [],
    "sharings": []
}`

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -308,6 +333,51 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
+      setOpenPasswordDisableDialog(true);
+    }
+  }, [isPasswordProtected]);
+
+  const onSavePublicSharePassword = useCallback(
+    async (passwordValue: string) => {
+      try {
+        let sharingInfo = sharingMeta;
+
+        if (!sharingInfo) {
```
apsantiso: If users click on "protect with password" (and saves a password) and the item is not being shared yet, we create one public sharing automatically for them.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -175,31 +198,33 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
+    if (isItemNotSharedYet) return;
+
     setIsLoading(true);
     // Change object type of itemToShare to AdvancedSharedItem
     let shareAccessMode: AccessMode = 'public';
     let sharingType = 'public';
+    let isAlreadyPasswordProtected = false;
 
-    if (props.isDriveItem) {
-      sharingType = (itemToShare?.item as DriveItemData & { sharings: { type: string; id: string }[] }).sharings?.[0]
```
apsantiso, line 184: We dont really need this check. The` sharingType` returned by the backend in the drive section is not accurate anymore since we implemented invites on public sharings.

We should always use the getSharingType endpoint to get the sharing info.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -175,20 +198,26 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
+
+    if (!isItemNotSharedYet) {
       try {
         const sharingData = await shareService.getSharingType(itemId, itemType);
         sharingType = sharingData.type;
+        isAlreadyPasswordProtected = sharingData.encryptedPassword !== null;
+        setSharingMeta(sharingData);
       } catch (error) {
         errorService.reportError(error);
       }
```
apsantiso, line 223: This endpoint only returns data if the item is shared, so to prevent not esential requests to the backend, lets check first if the item was shared.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -275,11 +303,14 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
       const encryptionKey = isAdvanchedShareItem(itemToShare.item) ? itemToShare?.item?.encryptionKey : undefined;
-      await shareService.getPublicShareLink(
+      const sharingInfo = await shareService.getPublicShareLink(
         itemToShare?.item.uuid,
         itemToShare.item.isFolder ? 'folder' : 'file',
         encryptionKey,
       );
+      if (sharingInfo) {
+        setSharingMeta(sharingInfo);
+      }
```
apsantiso, line 313: I know this is a bit of a workaround, but we already send a request to copy the link, so let's use the same request to get the sharingInfo sent by the backend.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -175,20 +198,26 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
-      sharingType = (itemToShare?.item as DriveItemData & { sharings: { type: string; id: string }[] }).sharings?.[0]
-        ?.type;
-    } else {
-      const itemType = itemToShare?.item.isFolder ? 'folder' : 'file';
-      const itemId = itemToShare?.item.uuid ?? '';
+    const itemType = itemToShare?.item.isFolder ? 'folder' : 'file';
+    const itemId = itemToShare?.item.uuid ?? '';
+
+    const isItemNotSharedYet =
+      !isAdvanchedShareItem(itemToShare?.item) && !itemToShare.item.sharings?.length && !sharingMeta;
```
CandelR, line 213: Drive items not have `encryptionKey`. Will it work correctly in the Drive section? Has the property been added to return it in the Drive items to server-wip in this task?

src/app/share/components/SharePasswordInputDialog/SharePasswordInputDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <Modal maxWidth="max-w-sm" className="space-y-5 p-5" isOpen={isOpen} onClose={onClose}>
+      <p className="text-2xl font-medium">
+        {!isAlreadyProtected
+          ? translate('modals.shareModal.protectSharingModal.createPasswordTitle')
+          : translate('modals.shareModal.protectSharingModal.editPasswordTitle')}
+      </p>
+      <Input
+        onChange={(value) => {
+          if (value.length <= 50) {
```
CandelR: Add the number to a constant

src/app/share/views/ShareView/ShareFileView.tsx
```diff
@@ -133,8 +134,10 @@ export default function ShareFileView(props: ShareViewProps): JSX.Element {
           name: res.item.plainName,
         });
       })
-      .catch((err) => {
+      .catch(async (err) => {
         if (err.message === 'Forbidden') {
+          const itemData = await shareService.getPublicSharedItemInfo(sharingId);
+          setItemData(itemData);
```
CandelR: Perhaps it is better to extract the logic that makes the call to a function, since the call can give some kind of error and it is convenient to add a try catch as well.

src/app/share/views/ShareView/ShareFolderView.tsx
```diff
@@ -121,8 +119,10 @@ export default function ShareFolderView(props: ShareViewProps): JSX.Element {
       .then((folderSize) => {
         setSize(folderSize);
       })
-      .catch((err) => {
+      .catch(async (err) => {
         if (err.message === 'Forbidden') {
+          const itemData = await shareService.getPublicSharedItemInfo(sharingId);
+          setItemData(itemData);
```
CandelR: Same here, maybe it is better to extract this to a function to control possible errors.

CandelR (changes requested):

src/app/drive/components/ShareDialog/ShareDialog.tsx
apsantiso: To access to itemToShare.item.sharings without Typescript complaining, I needed to narrow the type. So, this is the flow of this condition:

1. If `isAdvanchedShareItem` is false, it means it is not shared yet. This also means that the item comes from the **Drive** section.
2. So, we have a ! before the previous condition, now we check ` itemToShare.item.sharings?.length` (we know this comes from Drive already), if is other than 0, then the item was shared. 
3. If no, we check `SharingMeta` , which should be populated already by `loadShareInfo`. If this is also false, then the item has not been shared.

apsantiso: Drive was my main section for testing as we had a few unhandled cases when the modal is open from that section. Hope this fixes unhandled cases we have on production while also saves us a few requests to the backend haha

CandelR: ah excuse me, the conditional is checking if it fulfils the three conditions to affirm that it is not shared. I had read it wrong 🙏

CandelR (approved):

CandelR:
@apsantiso approved! Remains to resolve conflicts 🚀

PR #993, opened 2024-01-11
[PB-1526]: feat/details actions in Grid view

xabg2:
The details action has been added to Grid view (context menu).

**FIXES:**
* Weird behavior of the Move elements dialog. Sometimes the button is disabled when it is not supposed to.
* Navigation from the breadcrumbs when we move an item from the dialog.


src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -144,10 +144,20 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
   const setDriveBreadcrumb = () => {
     const driveBreadcrumbPath = [...currentNamePaths, { id: itemsToMove[0].id, name: itemsToMove[0].name }];
     dispatch(storageActions.popNamePathUpTo({ id: currentNamePaths[0].id, name: currentNamePaths[0].name }));
-    itemsToMove[0].isFolder &&
+
+    if (itemsToMove.length > 1) {
```
CandelR, line 149: Why if `itemsToMove` length is more than 1 has to do nothing?

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -144,10 +144,20 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
...
+      return;
+    } else {
```
CandelR: Remove the else conditional, it is no necessary with the previous guard clause.

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -42,8 +42,8 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
   const dispatch = useAppDispatch();
   const isOpen = useAppSelector((state: RootState) => state.ui.isMoveItemsDialogOpen);
   const rootFolderID: number = useSelector((state: RootState) => storageSelectors.rootFolderId(state));
-  const itemParentId: number = itemsToMove[0]?.parentId ?? itemsToMove[0]?.folderId;
-  const isDriveAndCurrentFolder = !props.isTrash && itemParentId === currentFolderId;
+  const itemParentId = itemsToMove[0]?.parentId ?? itemsToMove[0]?.folderId;
+  const isDriveAndCurrentFolder = !props.isTrash && itemParentId === destinationId;
```
CandelR, line 47: Is changing `currentFolderId` to `destinationId` some kind of fix for something?

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -301,9 +302,11 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
     (item: DriveItemData | AdvancedSharedItem) => {
       if (item.isFolder) {
         dispatch(storageThunks.goToFolderThunk({ name: item.name, id: item.id }));
+        navigationService.pushFolder(item.uuid);
```
CandelR, line 306: Is this added because it is not navigating to the folder? 
If `goToFolderThunk` is not working better add the change inside the thunk, not outside.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -301,9 +302,11 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
...
       } else {
         dispatch(uiActions.setIsFileViewerOpen(true));
         dispatch(uiActions.setFileViewerItem(item as DriveItemData));
+        navigationService.pushFile(item.uuid);
```
CandelR, line 308: It is necessary the previous dispatch in order to trigger the FileViewer?

src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerListItem/DriveExplorerListItem.tsx
```diff
@@ -55,7 +55,7 @@ const DriveExplorerListItem = ({ item }: DriveExplorerItemProps): JSX.Element =>
       <div className="flex min-w-activity grow items-center pr-3">
         {/* ICON */}
         <div className="box-content flex items-center pr-4">
-          <div className="flex h-10 w-10 justify-center drop-shadow-soft">
+          <div className="relative flex h-10 w-10 justify-center drop-shadow-soft">
```
CandelR, line 58: For what is the relative style?

CandelR (changes requested):
If there are more changes than the task title please add it to the description 🙏

src/app/drive/components/DriveExplorer/DriveExplorerItem/DriveExplorerListItem/DriveExplorerListItem.tsx
xabg2: For the user icon in Safari. When a component is absolute, we must add relative to the parent that contains this absolute child, otherwise, the absolute element will get the whole screen.

[Info](https://www.freecodecamp.org/news/css-positioning-position-absolute-and-relative/#:~:text=Absolute%2Dpositioned%20elements%20are%20completely,inside%20a%20statically%20positioned%20parent.).

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
xabg2: Yep, because the FileViewer is placed in App.tsx, so we can use the FileViewerWrapper component for SharedView, no matter if the user is logged in or not. But we can add the wrapper in DriveExplorer like SharedView to open it with props instead of redux.

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
xabg2: Because if the user moves more than 1 item, we redirect the user to the first selected folder. But if this is the expected behavior, I can change it.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
xabg2: I've noticed that we use the `pushFolder` outside the `goToFolderThunk` in some cases, for example in Breadcrumbs. I'm not sure if its better to add `pushFolder` inside of `goToFolderThunk`

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
xabg2: Yep. Now the move items dialog works weird. If you can, check the MoveItemsDialog in prod and then check it in the preview link.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
CandelR: Mmm, we should revise it to standardise in one place the way to navigate through the folders.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
CandelR: I think we are not talking about the same thing. 
In Drive at least, with the new navigation you don't need to use the `dispatch` of those `uiActions` used here, just add the uuid of the file or folder with the navigationService and it opens automatically. 
Check if it works as expected without the previous dispatches:
```
dispatch(uiActions.setIsFileViewerOpen(true));
dispatch(uiActions.setFileViewerItem(item as DriveItemData)); 
```

xabg2: Ah okay. I'll take a look.

xabg2: Hey! I've just checked if it works without the dispatch and you were right 🙌🏽

aboredvaro (approved):

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -141,20 +142,15 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
...
       return;
     }
 
-    driveBreadcrumbPath.forEach((item) => {
-      dispatch(storageActions.pushNamePath({ id: item.id, name: item.name }));
-    });
-
-    if (itemsToMove[0].isFolder) {
-      navigationService.pushFolder(itemsToMove[0].uuid);
+    if (itemsToMove[0].isFolder && currentItemUuid === itemsToMove[0].uuid) {
```
CandelR: abstracts the conditions for a better understanding

CandelR (approved):

src/app/drive/components/DriveItemDropdownActions/DriveItemDropdownActions.tsx
```diff
@@ -1,169 +1,139 @@
...
-          text: translate('drive.dropdown.openPreview'),
+  const {
+    onCopyLinkButtonClicked,
+    onMoveItemButtonClicked,
+    onMoveToTrashButtonClicked,
+    onShowDetailsButtonClicked,
+    onDownloadItemButtonClicked,
+    onLinkSettingsButtonClicked,
+    onRenameItemButtonClicked,
+    onOpenPreviewButtonClicked,
```
SimonGodefroid, line 30: **nit** 
these would be less of verbose mouthfuls with `handleCopyLink` kind of convention.
```tsx
onClick={handleClick}
```

sg-gs (approved):

PR #995, opened 2024-01-15
[PB-1536, PB-1537, PB-1538, PB-1539] Added files and folders pause upload ability and retry file upload

CandelR:
- Added pause and resume to multipart file uploads
- Added pause and resume to folder uploads
- Added retry option to file uploads
- Added retry option to folder uploads


src/app/tasks/types.ts
```diff
@@ -140,18 +144,27 @@ export type TaskData = (
...
+  parentFolderId?: number;
 };
 
 export interface TaskNotification {
   taskId: string;
   action: TaskType;
   status: TaskStatus;
   title: string;
-  item?: DriveItemData | { name: string; type: string; items?: DriveItemData[] } | { id: number; name: string };
+  item?:
```
sg-gs: Please use Types here so it's clear what each payload is and why it's there

src/app/tasks/hooks/useRetry.tsx
```diff
@@ -0,0 +1,79 @@
...
+        dispatch(
+          uploadFolderThunk({
+            root: notification?.folderToUpload?.folder as IRoot,
+            currentFolderId: currentFolderId,
+            options: {
+              taskId,
+            },
+          }),
+        );
+    } else if (item && taskId) {
```
sg-gs: What happens in the case of the else? it's a good idea to handle it, so we do not ignore any failure there.

src/app/tasks/hooks/useRetry.tsx
sg-gs, line 1: So we have a hook here, if the function pureness is respected, we should be able to add a unit test for this without any issue.

src/app/tasks/components/TaskLoggerItem/TaskLoggerItem.tsx
```diff
@@ -72,17 +73,19 @@ const TaskLoggerItem = ({ notification }: TaskLoggerItemProps): JSX.Element => {
...
           <span className={`text-sm ${messageClassName}`}>{notification.subtitle}</span>
         </div>
         <TaskLoggerActions
+          taskId={notification.taskId}
           isHovered={isHovered}
           status={notification.status}
           progress={progress.toString()}
           cancelAction={onCancelButtonClicked}
-          retryAction={isDownloadError ? retryDownload : () => {}}
+          retryAction={isDownloadError ? retryDownload : retryUpload}
```
sg-gs: I would use a function here, as it is not clear why if it is not a download error, why we retry the upload

src/app/tasks/components/TaskLoggerActions/TaskLoggerActions.tsx
```diff
@@ -19,7 +22,28 @@ type TaskLoggerActionsProps = {
...
+const pauseUpload = (id: string) => {
+  localStorageService.setUploadState(id, TaskStatus.Paused);
+  tasksService.updateTask({
+    taskId: id,
+    merge: { status: TaskStatus.Paused },
+  });
+};
+
+const resumeUpload = (id: string) => {
+  localStorageService.setUploadState(id, TaskStatus.InProcess);
```
sg-gs: I recommend using the local database rather than the local storage, also, if we can hide which is the implementation under a cache service, the better. 

If we do not do that, what is going to happen if the local storage suddenly is not supported, is blocked or we want to switch to another persistence layer? We would need to change each local storage usage in the code, instead of having it centralized. I think the Repository pattern is the right fit here

src/app/network/upload.ts
```diff
@@ -127,15 +131,22 @@ export function uploadFile(bucketId: string, params: IUploadParams): Promise<str
...
+    uploadPromise = facade.uploadMultipart(
+      bucketId,
+      params.mnemonic,
+      file,
+      {
+        uploadingCallback: params.progressCallback,
+        abortController: uploadAbortController,
+        parts: Math.ceil(params.filesize / partSize),
+      },
+      params?.continueUploadOptions,
```
sg-gs: Why the continue upload is not inside the params body like the uploading callback or the abortController?

src/app/network/UploadManager.ts
```diff
@@ -110,16 +113,33 @@ class UploadManager {
...
 
+        if (!existsRelatedTask)
+          tasksService.updateTask({
+            taskId: taskId,
+            merge: {
+              status: TaskStatus.InProcess,
+            },
+          });
+
+        const uploadStatus = localStorageService.getUploadState(this.options?.relatedTaskId ?? taskId);
```
sg-gs: The upload manager should not know anything about the localStorage, this is a very coupled implementation. I would need to decouple this first if I want to test only the manager behavior

src/app/drive/services/worker.service/upload.service.ts
```diff
@@ -0,0 +1,29 @@
+import { TaskStatus } from '../../../tasks/types';
+
+async function waitForContiueUploadSignal(taskId: string): Promise<void> {
+  return new Promise((resolve) => {
+    const checkProgress = async () => {
```
sg-gs: Please extract the `checkProgress` function definition from the Promise callback, this code complexity is a headache

src/app/drive/services/network.service/index.ts
```diff
@@ -99,16 +113,22 @@ export class Network {
           console.log('[MAIN_THREAD]: Message received from worker', msg);
           if (msg.data.progress) {
             params.progressCallback(msg.data.progress, msg.data.uploadedBytes, msg.data.totalBytes);
-          } else if (msg.data.result === 'success') {
+          } else if (msg.data.result === WORKER_MESSAGE_STATES.SUCCESS) {
```
sg-gs: `switch` is your friend here!

sg-gs (changes requested):
Summary @CandelR: 
- We need to decouple what the UI needs from the ops we run to transfer data
- We need to reduce the complexity of this code
- We need to avoid to couple Redux, Thunks and co from the pure logic, including hooks

src/app/drive/services/worker.service/uploadWorkerUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/**
+ * Waits for a signal to continue the upload for the specified task.
+ *
+ * @param {string} taskId - The identifier of the task for which the upload continuation signal is awaited.
+ *
+ * @returns {Promise<void>} - A promise that resolves once the upload continuation signal is received.
+ */
+async function waitForContiueUploadSignal(taskId: string): Promise<void> {
```
sg-gs: `async` keyword is not needed here

src/app/tasks/hooks/useRetry.ts
```diff
@@ -0,0 +1,84 @@
...
+}: RetryDownloadArgs): RetryDownload => {
+  const retryDownload = useCallback(() => {
+    const { item, taskId } = notification;
+    const isZipAndMultipleItems = item && 'items' in item && item?.items && item?.type === 'zip';
+
+    if (isZipAndMultipleItems) {
+      downloadItemsAsZip(item.items as DriveItemData[], taskId);
+    } else if (item && taskId) {
+      downloadItems(item as DriveItemData, taskId);
+    }
```
sg-gs, line 32: and the `else`?

src/app/repositories/DatabaseUploadRepository.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    return DatabaseUploadRepository.instance;
+  }
+
+  async setUploadState(id: string, status: TaskStatus): Promise<void> {
+    await this.db.put(DatabaseCollection.UploadItemStatus, id, status);
+  }
+
+  async getUploadState(id: string): Promise<TaskStatus | undefined> {
+    return await this.db.get(DatabaseCollection.UploadItemStatus, id);
```
sg-gs: `return` directly and remove `await` and `async`

sg-gs (approved):
Please correct the comments I have left

PR #999, opened 2024-01-16
[PB-1401, PB-1403] add blocked account view

apsantiso:
Dependencies:

https://github.com/internxt/drive-server-wip/pull/250
https://github.com/internxt/sdk/pull/185

New dependencies for Http Error code catching on login:

https://github.com/internxt/sdk/pull/186

If this is not implemented, then we would need to compare the message string instead of the code to know if the user has been blocked.


src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -172,12 +180,14 @@ export default function LogIn(): JSX.Element {
 
       if (castedError.message.includes('not activated') && auth.isValidEmail(email)) {
         navigationService.history.push(`/activate/${email}`);
-      } else {
-        // analyticsService.signInAttempted(email, castedError);
       }
 
       setLoginError([castedError.message]);
       setShowErrors(true);
+      if (castedError.message.includes('Your account has been blocked for security reasons. Please reach out to us')) {
```
CandelR: This does not seem to be a good way to check the error, at the slightest change in the message it will stop working, and it will not take into account that a change in the message may break something
Use the status that returns the call or another signal.

src/app/auth/services/auth.service.ts
```diff
@@ -41,7 +41,9 @@ export async function logOut(loginParams?: Record<string, string>): Promise<void
   await databaseService.clear();
   localStorageService.clear();
   RealtimeService.getInstance().stop();
-  navigationService.push(AppView.Login, loginParams);
+  if (navigationService.history.location.pathname !== '/blocked-account') {
```
CandelR: Use the value of the corresponding `AppView` enum to check the `pathname` to avoid errors if the pathname of that view ever changes.

src/app/auth/views/BlockedAccountView/BlockedAccountView.tsx
```diff
@@ -0,0 +1,154 @@
...
+    () =>
+      enableResendButton ? (
+        <button className="ml-2 cursor-pointer text-primary" onClick={resendAccountUnblockEmail}>
+          {translate('blockedAccount.resend')}
+        </button>
+      ) : (
+        <span className="ml-2 font-medium">&nbsp;{countDown}</span>
+      ),
+    [countDown, enableResendButton, resendAccountUnblockEmail],
+  );
```
CandelR: There is not much point in this useCallback here, if you want to memoise the component so that it is not rendered again if it is not necessary use useMemo.

src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -172,12 +180,14 @@ export default function LogIn(): JSX.Element {
...
+        await sendUnblockAccountEmail(email);
+        navigationService.push(AppView.BlockedAccount, { email: email });
```
CandelR: I have seen that the view `BlockedAcccount` requires a token, where do you get it from in this case if the navigation is direct after making the unlock request?

CandelR (changes requested):

CandelR (changes requested):
The sdk version needs to be updated to use the new unlocking features.

src/app/auth/components/LogIn/LogIn.tsx
apsantiso: It is optional. 

1. Exists: show a loader and send the token to the backend to unlock the account linked to it.  If the request is successfull then go to /login, otherwise show expired view. The backend will get all the needed info from the jwt so the front is agnostic to this.
2. No token: show the view and let the user to resend the email after 30 seconds.

CandelR (approved):

PR #1005, opened 2024-01-22
[PB-1372, PB-1373, PB-1595]: Fix/shortcuts for Shared and File Viewer

xabg2:
**FIX:**
* Remove duplicated Language for free accounts.
* Allow the user to rename item using shortcuts in File Viewer by pressing R.
* Allow the user to move the item to trash using shortcuts by pressing ESC.

**More:**
* FileViewerWrapper has been refactored.


src/app/drive/components/DriveExplorer/DriveExplorerList/DriveExplorerList.tsx
```diff
@@ -429,7 +429,10 @@ const DriveExplorerList: React.FC<DriveExplorerListProps> = memo((props) => {
                 })
               : contextMenuDriveNotSharedLink({
                   shareLink: openLinkSettings,
-                  openPreview: openPreview,
+                  openPreview: () => {
+                    console.log(props.selectedItems[0]);
+                    openPreview(props.selectedItems[0]);
```
PixoDev: you better check for that [0] item, it always starts like this, and few weeks later you end debugging a release on a friday @masterprog-cmd

src/app/drive/components/FileViewer/FileViewer.tsx
```diff
@@ -39,7 +39,10 @@ interface FileViewerProps {
   totalFolderIndex?: number;
   fileIndex?: number;
   dropdownItems?: TopBarActionsMenu;
-  renameItemFromKeyboard?: () => void;
+  keyboardShortcuts?: {
+    renameItemFromKeyboard: ((item) => void) | undefined;
```
PixoDev, line 43: rather than defining each handler like this, why dont you create something like a `KeyboardShortcut` object that looks something like this:

```
type KeyboardShortcut {
    key: string;
    startListening: () => void;
    stopListening: () => void;
}
```

You can then create a class that encapsulates all that logic, and control how it behaves from each component, in terms of props, you just pass an array of `KeyboardShortcut`

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -70,108 +63,28 @@ const FileViewerWrapper = ({
...
-    onMoveItemButtonClicked,
-    onCopyLinkButtonClicked,
-    onShowDetailsButtonClicked,
-    onLinkSettingsButtonClicked,
-    onMoveToTrashButtonClicked,
-    onRenameItemButtonClicked,
-    onRestoreItemButtonClicked,
-    onDeletePermanentlyButtonClicked,
-  } = useDriveItemActions(currentFile);
+  const driveItemActions = useDriveItemActions(currentFile);
```
PixoDev, line 72: not sure if you want to keep this global, maybe you can scope it down using a local context for each wrapper?

src/app/drive/components/FileViewer/utils/fileViewerWrapperUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+        onClose();
+      },
+    });
+  };
+
+  const sharedActionsMenu = (): ListItemMenu<AdvancedSharedItem> => {
+    return contextMenuDriveItemSharedAFS({
+      openShareAccessSettings: isOwner ? onLinkSettingsButtonClicked : undefined,
+      copyLink: onCopyLinkButtonClicked,
+      deleteLink: () => undefined,
```
PixoDev: this looks like code smell, why is this a NOOP?

src/app/drive/components/FileViewer/utils/fileViewerWrapperUtils.ts
```diff
@@ -0,0 +1,222 @@
...
+    },
+    abort: () => {
+      abortController.abort();
+    },
+  };
+}
+
+export const useKeyboardShortcuts = ({
+  sharedKeyboardShortcuts,
+  driveItemActions,
```
PixoDev, line 196: why is this coupled to driveItemActions? for me this looks like should be "product independent" as is not business logic

src/app/drive/components/FileViewer/utils/fileViewerWrapperUtils.ts
xabg2: Because we don't use delete link anymore since we added the new Shared. But yeah, I think we have to remove this option.

src/app/drive/components/DriveExplorer/DriveExplorerList/DriveExplorerList.tsx
xabg2: Yep, I just was checking something, I forgot to remove all😅

src/app/drive/components/FileViewer/utils/fileViewerWrapperUtils.ts
```diff
@@ -0,0 +1,220 @@
...
+      return trashActionsMenu();
+    } else {
+      return driveActionsMenu();
+    }
+  };
+
+  return contextMenuActions();
+};
+
+export function getFileContentManager(currentFile, downloadFile, handleFileThumbnail) {
```
CandelR: nit: Personally I prefer to export all the functions at the end of the file, it is easier to see what this file exports

src/app/drive/components/FileViewer/utils/fileViewerWrapperUtils.ts
```diff
@@ -0,0 +1,220 @@
...
+export type TopBarActionsMenu = ListItemMenu<DriveItemData> | ListItemMenu<AdvancedSharedItem>;
+
+type PathProps = 'drive' | 'trash' | 'shared' | 'recents';
+
+interface FileViewerShortcuts {
+  renameItemFromKeyboard: ((item) => void) | undefined;
+  removeItemFromKeyboard: ((item) => void) | undefined;
+}
+
+export const TopDropdownBarActionsMenu = ({
```
CandelR: This is not a component if I am not mistaken, by convention the functions use `camelCase` and components `PascalCase`

CandelR (approved):

PR #1007, opened 2024-01-29
[PB-1525] feat/persist language in login and signup

rafijv:
When the user arrives at drive web from the website the selected language persists.


src/App.tsx
```diff
@@ -58,8 +59,19 @@ const App = (props: AppProps): JSX.Element => {
...
 
+  const changeLanguage = () => {
+    const websiteLanguage = getCookie('LOCALE');
+
+    if (websiteLanguage && websiteLanguage === 'zh') {
+      i18next.changeLanguage('cn');
+    } else if (websiteLanguage && websiteLanguage != 'zh') {
+      i18next.changeLanguage(websiteLanguage);
+    }
+  };
```
CandelR: Put this function in the service, so that everything related to translations is kept there

CandelR (changes requested):

src/App.tsx

CandelR (approved):

PR #1013, opened 2024-02-01
[PB-701] feat/display folder size on shared landing page

rafijv:
Added folder size in shared folder landig page


src/app/drive/services/size.service.ts
```diff
@@ -7,7 +7,7 @@ export function bytesToString(
   decimals = 1,
   hideSizeString = false,
 ): string {
-  if (size > 0) {
+  if (size >= 0) {
     return PrettySize(size, removeSpace, useSingleChar, decimals, hideSizeString);
   } else {
     return '';
```
CandelR: This change affects many parts of the app by making it appear that there are 0 bytes instead of nothing at all. Does this have to be the case?
Now, for example, in the list of folders in Drive it appears that they occupy 0 bytes, which is not the case.

src/app/share/views/ShareView/ShareFolderView.tsx
```diff
@@ -182,6 +174,17 @@ export default function ShareFolderView(props: ShareViewProps): JSX.Element {
     return confirmationMessage; // WebKit, Chrome <34
   };
 
+  const getFolderSize = async (folderId) => {
```
CandelR: add the type to the function arguments

src/app/share/views/ShareView/ShareFolderView.tsx
```diff
@@ -182,6 +174,17 @@ export default function ShareFolderView(props: ShareViewProps): JSX.Element {
...
+    try {
+      const folderSize = await shareService.getSharedFolderSize(folderId);
```
CandelR: nit: rename this, as it is not the size of the folder, it is an object containing the size

CandelR (changes requested):

src/app/share/views/ShareView/ShareFolderView.tsx
rafijv: Done

src/app/share/views/ShareView/ShareFolderView.tsx
rafijv: Done

src/app/drive/services/size.service.ts
rafijv: I've left it as it was

CandelR (approved):

PR #1016, opened 2024-02-05
[PB-701] feat/display folder size on shared landing page

rafijv:
Add spinner to folder size


src/app/share/views/ShareView/ShareFolderView.tsx
```diff
@@ -271,7 +271,11 @@ export default function ShareFolderView(props: ShareViewProps): JSX.Element {
               <abbr className="w-screen max-w-prose break-words px-10 text-xl sm:w-full" title={info?.item?.plainName}>
                 {info?.item?.plainName}
               </abbr>
-              <span className="text-gray-60">{folderSize || '0MB'}</span>
+              {folderSize === null ? (
```
CandelR: If the call fails the spinner will be displayed inifitly. I think adding something to let the user know that the call has failed will be more user friendly and won't give the feeling that it's broken.

CandelR (changes requested):

src/app/share/views/ShareView/ShareFolderView.tsx
rafijv: In case the call fails, the loader will be hidden and nothing will be displayed, it is the same behaviour as before.

CandelR (approved):
@rafijv Approved, but before merge solve the sonar cloud issue

PR #1028, opened 2024-02-13
[PB-1639] refactor/Refactor Shared view component

CandelR:
- Refactored SharedView

### DIFF
--- a/src/app/drive/components/ShareDialog/ShareDialog.tsx
+++ b/src/app/drive/components/ShareDialog/ShareDialog.tsx
@@ -153,7 +153,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
   const userList = useRef<HTMLDivElement>(null);
   const userOptions = useRef<HTMLButtonElement>(null);
 
-  const itemOwnerEmail = props?.isDriveItem ? '' : (itemToShare?.item as AdvancedSharedItem).user?.email;
+  const itemOwnerEmail = props?.isDriveItem ? '' : (itemToShare?.item as AdvancedSharedItem)?.user?.email;
   const isUserOwner = (!!itemOwnerEmail && itemOwnerEmail === props?.user?.email) || !!props?.isDriveItem;
 
   const closeSelectedUserPopover = () => setSelectedUserListIndex(null);
--- a/src/app/drive/components/ShareDialog/ShareDialog.tsx
+++ b/src/app/drive/components/ShareDialog/ShareDialog.tsx
@@ -93,6 +93,7 @@ type ShareDialogProps = {
   user: UserSettings;
   isDriveItem?: boolean;
   onShareItem?: () => void;
+  onCloseDialog?: () => void;
 };
 
 const isAdvanchedShareItem = (item: DriveItemData | AdvancedSharedItem): item is AdvancedSharedItem => {
@@ -129,6 +130,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
   const isOpen = useAppSelector((state: RootState) => state.ui.isShareDialogOpen);
   const isToastNotificationOpen = useAppSelector((state: RootState) => state.ui.isToastNotificationOpen);
   const itemToShare = useAppSelector((state) => state.storage.itemToShare);
+  const { onCloseDialog } = props;
 
   const [roles, setRoles] = useState<Role[]>([]);
   const [inviteDialogRoles, setInviteDialogRoles] = useState<Role[]>([]);
@@ -168,6 +170,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
     setView('general');
     setIsPasswordProtected(false);
     setSharingMeta(null);
+    onCloseDialog?.();
   };
 
   useEffect(() => {
--- a/src/app/drive/components/ShareDialog/ShareDialog.tsx
+++ b/src/app/drive/components/ShareDialog/ShareDialog.tsx
@@ -43,6 +43,7 @@ import { Tooltip } from 'react-tooltip';
 import { DELAY_SHOW_MS } from 'app/shared/components/Tooltip/Tooltip';
 import StopSharingItemDialog from '../StopSharingItemDialog/StopSharingItemDialog';
 import { MAX_SHARED_NAME_LENGTH } from 'app/share/views/SharedLinksView/SharedView';
+import { isUserItemOwner } from '../../../share/views/SharedLinksView/sharedViewUtils';
 
 type AccessMode = 'public' | 'restricted';
 type UserRole = 'owner' | 'editor' | 'reader';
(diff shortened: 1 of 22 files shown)


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -93,6 +93,7 @@ type ShareDialogProps = {
   user: UserSettings;
   isDriveItem?: boolean;
   onShareItem?: () => void;
+  onCloseDialog?: () => void;
 };
 
 const isAdvanchedShareItem = (item: DriveItemData | AdvancedSharedItem): item is AdvancedSharedItem => {
```
PixoDev, line 100: Im a bit confused by this, a `DriveItemData` can be a shared item, but a shared item is an `AdvancedSharedItem` which is not the same as a `DriveItemData`, but could be 😅

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -129,6 +130,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
   const isOpen = useAppSelector((state: RootState) => state.ui.isShareDialogOpen);
   const isToastNotificationOpen = useAppSelector((state: RootState) => state.ui.isToastNotificationOpen);
   const itemToShare = useAppSelector((state) => state.storage.itemToShare);
+  const { onCloseDialog } = props;
```
PixoDev, line 134: Just an idea for this, cause you can abstract this kind of things easily by thinking about this, this way:

A dialog is just a wrapper.

Meaning you don't have a `ShareDialog`, you have let's say a `ShareConfigView`, which you can fit inside multiple components, one of the components that you might use this in, is a let's call it `Dialog`, so you can do this:

```
<Dialog onClose={props.onClose}>
    <ShareConfig />
</Dialog>
```

And by doing this you:

- Removed a lot of props from a lot of components if you use this pattern
- Now your dialogs are an abstract UI box where you can fit other components
- You can now reuse the `ShareConfig` in other parts of the app if needed

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -151,7 +153,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
   const userList = useRef<HTMLDivElement>(null);
   const userOptions = useRef<HTMLButtonElement>(null);
 
-  const itemOwnerEmail = props?.isDriveItem ? '' : (itemToShare?.item as AdvancedSharedItem).user?.email;
+  const itemOwnerEmail = props?.isDriveItem ? '' : (itemToShare?.item as AdvancedSharedItem)?.user?.email;
   const isUserOwner = (!!itemOwnerEmail && itemOwnerEmail === props?.user?.email) || !!props?.isDriveItem;
```
PixoDev: `AdvancedShareUtils.isUserOwner()` might by  a better way to handle this, so you have your logic centralized in one place, and every component that uses this, uses the same centralized logic

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -168,6 +170,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
     setView('general');
     setIsPasswordProtected(false);
     setSharingMeta(null);
+    onCloseDialog?.();
```
PixoDev, line 176: This is related to what I said above, this prop here feels weird, specially if it could be undefined, meaning that it could be not in a dialog

src/app/share/services/redirections.service.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Handles access to a private shared folder, fetching its content and navigating accordingly.
+ *
+ * @param {Object} options - Options object containing parameters.
+ * @param {string} options.folderUUID - The UUID of the folder to access.
+ * @param {History} options.history - The history object for navigation.
+ * @param {(sharedFolder: AdvancedSharedItem) => void} options.navigateToFolder - Function to navigate to the accessed folder.
+ * @param {() => void} options.onError - Function to be executed if an error occurs during the process.
+ */
+const handlePrivateSharedFolderAccess = ({
```
PixoDev: I think this function is hiding a lot of logic, maybe splitting this into multiple pieces will help to maintain + understand it.

`getPrivateSharedFolderAccess` +  `redirectBySharedFolderAccess` maybe?

src/app/share/services/redirections.service.ts
```diff
@@ -0,0 +1,65 @@
...
+        const shareItem = { plainName: (item as any).name, uuid: folderUUID, isFolder: true };
+        navigateToFolder(shareItem as AdvancedSharedItem);
+      })
+      .catch((error) => {
+        if (error.status === 403) {
+          statusError = 403;
+          navigationService.push(AppView.RequestAccess, { folderuuid: folderUUID });
+          return;
+        } else if (error.status === 404) {
+          notificationsService.show({ text: t('shared.errors.folderNotExists'), type: ToastType.Error });
```
PixoDev: you don't want to mix UI with logic, is better to return this result, and handle the UI stuff in an UI component.
Otherwise, a service can start showing notifications based on multiple criteria.

I know is tempting to do this, but you are just hidding errors by doing this, cause is never going to reach the consumer layer (a component in this case).

src/app/share/views/SharedLinksView/components/SharedItemList.tsx
```diff
@@ -0,0 +1,243 @@
...
+    <div key="3" className="h-4 w-40 rounded bg-gray-5" />
+  </div>,
+  <div key="4" className="h-4 w-20 rounded bg-gray-5" />,
+  <div key="5" className="h-4 w-24 rounded bg-gray-5" />,
+  <div key="6" className="h-4 w-20 rounded bg-gray-5" />,
+];
+
+type SharedListItem = {
+  onClickItem: (item) => void;
+  keyBoardShortcutActions: {
```
PixoDev, line 28: `onKeyboardShortcutPress(pressedKey: KeyboardShortcutEnum)` is going to kill 2 props here

src/app/share/views/SharedLinksView/components/SharedItemList.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const itemComposition = getSharedListItemComposition({
+    onItemDoubleClicked,
+    onNameClicked,
+    user,
+    getOwnerAvatarSrc,
+    checkIfIsItemSelected,
+  });
+
+  return (
+    <List<any, 'updatedAt' | 'createdAt' | 'ownerId' | 'fileSize'>
```
PixoDev: why any?

src/app/drive/components/ShareDialog/ShareDialog.tsx
PixoDev: An example: https://www.radix-ui.com/primitives/docs/components/alert-dialog

src/app/drive/components/ShareDialog/ShareDialog.tsx
CandelR: In principle an `AdvancedSharedItem` is like `DriveItemData` with some extra fields related to the new file sharing, I'll check it in case I introduce some changes in the types.

src/app/drive/components/ShareDialog/ShareDialog.tsx
CandelR: This prop is more related to an action that would be executed when closing a dialog, there is no need for an action to be executed when closing same dialog in different views.

src/app/share/views/SharedLinksView/components/SharedItemList.tsx
CandelR: It was already like that and I guess I forgot to change it. I'll check it!

src/app/share/views/SharedLinksView/components/SharedItemList.tsx
CandelR: After reviewing it, it would be necessary to change the implementation of the list, since it is not that any key is supported, but only specific keys that are supported by the generic list. We would have to review the other implementations of the list, if it's ok with you I'll leave it for another time, so as not to lose the focus of this refactor.

rafijv (approved):

PR #1037, opened 2024-02-26
[PB-1638] feat/refactor Breadcrumbs component

rafijv:
- Refactor breadcrumbs


src/app/drive/views/DriveView/DriveView.tsx
```diff
@@ -81,47 +80,14 @@ const DriveView = (props: DriveViewProps) => {
...
-
   return (
     <>
       <Helmet>
         <title>{title}</title>
         <link rel="canonical" href={`${process.env.REACT_APP_HOSTNAME}`} />
       </Helmet>
       <DriveExplorer
-        title={<Breadcrumbs items={breadcrumbItems()} rootBreadcrumbItemDataCy="driveViewRootBreadcrumb" />}
+        title={<BreadcrumbsDriveView namePath={namePath} dispatch={dispatch} />}
```
CandelR: Could use directly the dispatch inside the BreacbrumsDriveView container instead of passing by props

src/app/shared/components/Breadcrumbs/BreadcrumbsMenu/BreadcrumbsMenu.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const { translate } = useTranslationContext();
+  const dispatch = useAppDispatch();
+  const allItems = useAppSelector((state) => state.storage.levels);
+  const namePath = useAppSelector((state) => state.storage.namePath);
+  const currentBreadcrumb = namePath[namePath.length - 1];
+  const { breadcrumbDirtyName } = useDriveItemStoreProps();
+  const currentDevice = useAppSelector((state) => state.backups.currentDevice);
+  const path = getAppConfig().views.find((view) => view.path === location.pathname);
+  const pathId = path?.id;
+  const isSharedView = pathId === 'shared';
```
CandelR: I would do the same as you have done with the other containers in this Menu, I would pass the menu to the BreadcbrumbsItem by props instead of importing it directly, so in each container you can customize the BreadcrumbMenu and let the breadcrumb container of each view dictate the behaviour of the menu and the breadcrumbs.
That way the component will be simpler and will have less logic that belongs to each view where it is implemented.

CandelR (changes requested):

src/app/shared/components/Breadcrumbs/BreadcrumbsMenu/BreadcrumbsMenu.tsx
```diff
@@ -0,0 +1,283 @@
...
+      isShared: !!currentFolder[0].sharings?.length,
+      view: 'Drive',
+    };
+    dispatch(uiActions.setItemDetailsItem(itemDetails));
+    dispatch(uiActions.setIsItemDetailsDialogOpen(true));
+  };
+
+  const onCopyLinkButtonClicked = () => {
+    const item = currentFolder[0];
+    dispatch(sharedThunks.getSharedLinkThunk({ item }));
```
CandelR: Removing the old part of shared links I have seen that this is the get public link deprecated, as you have to modify the component if you can change it to the new way of getting the public link it would be great!

rafijv: Ok

src/app/drive/views/DriveView/DriveView.tsx

src/app/shared/components/Breadcrumbs/BreadcrumbsMenu/BreadcrumbsMenu.tsx
rafijv: Done

src/app/shared/components/Breadcrumbs/BreadcrumbsMenu/BreadcrumbsMenu.tsx
rafijv: Done

CandelR (approved):

PR #1052, opened 2024-03-06
[PB-1763]: fix/moving folder breacks breadcrumbs navigation

rafijv:


src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
```diff
@@ -122,13 +124,13 @@ const MoveItemsDialog = (props: MoveItemsDialogProps): JSX.Element => {
...
-    const driveBreadcrumbPath = [...currentNamePaths, { id: itemsToMove[0].id, name: itemsToMove[0].name }];
-    dispatch(storageActions.popNamePathUpTo({ id: currentNamePaths[0].id, name: currentNamePaths[0].name }));
-    itemsToMove[0].isFolder &&
-      driveBreadcrumbPath.forEach((item) => {
-        dispatch(storageActions.pushNamePath({ id: item.id, name: item.name }));
-      });
+  const setDriveBreadcrumb = async (itemsToMove) => {
+    const breadcrumbsList: FolderAncestor[] = await newStorageService.getFolderAncestors(itemsToMove[0].uuid);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore:next-line
```
CandelR, line 130: What typescript error is giving this change that needs to be ignored?

CandelR (changes requested):

src/app/drive/components/MoveItemsDialog/MoveItemsDialog.tsx
rafijv: It seems to be a problem with the version of typescript we use that gives error with toReversed()

PR #1056, opened 2024-03-11
[PB-1810]: feature/Add new preferences dialog

rafijv:


src/app/preferences/index.tsx
```diff
@@ -0,0 +1,107 @@
+import { useEffect, useState } from 'react';
```
CandelR: Don't use `index.tsx` to name a component file, I don't find it intuitive to do a search for the component itself. 
If you think it's better for some other reason, we can discuss it. 
In the 3 PRs I have in draft I have created a structure (the name of the parent folder is temporary until we remove the old settings), if you are ok we can use that structure, if not add one you are ok with and I'll adapt.

src/app/core/services/navigation.service.ts
```diff
@@ -57,6 +57,12 @@ const navigationService = {
     const currentUuid = pathname[2];
     return currentUuid;
   },
+  openPreferencesDialog(section: string, subsection: string) {
```
CandelR: Change the param to an object, it will be more understandable to use.

src/app/store/slices/ui/index.ts
```diff
@@ -123,6 +125,9 @@ export const uiSlice = createSlice({
     setIsSurveyDialogOpen: (state: UISliceState, action: PayloadAction<boolean>) => {
       state.isSurveyDialogOpen = action.payload;
     },
+    setIsPreferencesDialogOpen: (state: UISliceState, action: PayloadAction<boolean>) => {
+      state.isPreferencesDialogOpen = action.payload;
+    },
```
CandelR, line 130: If it is not necessary don't add the state to open the modal to redux, if it is only going to be managed from App.tsx you can use a local state. Using redux for that is overkill.

src/App.tsx
```diff
@@ -38,24 +38,53 @@ import { FolderPath } from 'app/drive/types';
...
+
+  useEffect(() => {
+    if (isOpenPreferencesDialog) {
+      dispatch(uiActions.setIsPreferencesDialogOpen(true));
+      setHaveParamsChanged(false);
+    } else {
+      dispatch(uiActions.setIsPreferencesDialogOpen(false));
+      setHaveParamsChanged(false);
+    }
+  }, [haveParamsChanged]);
```
CandelR: Perhaps it would be a good idea to extract this to group the code that manages the modal state of Preferences

src/app/preferences/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      );
+      changeSection(sectionParams[0].section, subsectionParams[0].subsection);
+      setHaveParamsChanged(false);
+    } else {
+      dispatch(uiActions.setIsPreferencesDialogOpen(false));
+      setHaveParamsChanged(false);
+    }
+  }, [haveParamsChanged]);
+
+  const changeSection = (onSection, OnSubsection) => {
```
CandelR: add the args types

src/app/preferences/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const selectedSection = SECTIONS.filter((section) => section.section === onSection);
+    const selectedSubsection = selectedSection.filter((subsection) => subsection.subsection === OnSubsection);
+    setActiveSection(selectedSubsection);
+  };
+
+  const goSection = (section, subsection) => {
```
CandelR: same here, add the types

src/app/preferences/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    changeSection(section, subsection);
+  };
+
+  const onClose = () => {
+    dispatch(uiActions.setIsPreferencesDialogOpen(false));
+    navigationService.closePreferencesDialog();
+  };
+
+  return (
+    <Modal maxWidth="w-full max-w-4xl" className="m-0 flex" isOpen={true} onClose={() => onClose()}>
```
CandelR: to display the modal could pass the `isOpen` prop from the parent

src/app/preferences/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+      </section>
+      {/* SIDEBAR MENU */}
+
+      {/* ACTIVE SECTION */}
+      <section className="relative w-full">
+        <button className="absolute right-0 z-50 m-4 flex w-auto" onClick={() => onClose()}>
+          <X size={22} />
+        </button>
+        <h2>{activeSection[0].title}</h2>
+      </section>
```
CandelR: I have checked in the preview and it seems to me that this does not have the same styles as in the design. I also think that the modal has to have a fixed size.

CandelR (changes requested):

src/app/preferences/index.tsx
rafijv: For the size of the modal, it is the one specified in the description of the task.
https://inxt.atlassian.net/browse/PB-1810
The styles I have taken from figma, anyway when we integrate the rest of the sections I can check it.

src/app/preferences/index.tsx
rafijv: Done

src/app/preferences/index.tsx
rafijv: Done

src/app/preferences/index.tsx
rafijv: Done

src/App.tsx

src/app/store/slices/ui/index.ts
rafijv: Needed for other components

src/app/core/services/navigation.service.ts
rafijv: Done

src/app/preferences/index.tsx
rafijv: I agree with what you, done

CandelR (approved):

PR #1060, opened 2024-03-18
[_]: feat/dynamic-upload-concurrency-depending-on-ram-chrome

sg-gs:
**Context**
We have received complaints about crashes on the upload of thousands of files, because with the time, it exhaustes some devices resources using Chrome. After some research, the RAM could be the cause of this issue as we use concurrency to use all the available bandwidth on the upload to make it as fast as possible. 

**Changes**
- The UploadManager modifies the upload queue concurrency on Chrome depending on the RAM usage

### DIFF
--- a/src/app/network/UploadManager.ts
+++ b/src/app/network/UploadManager.ts
@@ -103,18 +103,10 @@ class UploadManager {
       if (this.abortController?.signal.aborted ?? fileData.abortController?.signal.aborted) return;
 
       if (window.performance && (window.performance as any).memory) {
-        const memory = (window.performance as unknown as { 
-          memory: {
-            totalJSHeapSize: number, 
-            usedJSHeapSize: number, 
-            jsHeapSizeLimit: number 
-          }
-        }).memory;
+        const memory = window.performance.memory;
 
-        if (memory.jsHeapSizeLimit !== null && memory.usedJSHeapSize !== null) {
+        if (memory && memory.jsHeapSizeLimit !== null && memory.usedJSHeapSize !== null) {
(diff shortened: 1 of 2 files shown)


src/app/network/UploadManager.ts
```diff
@@ -101,6 +102,41 @@ class UploadManager {
...
+          memory: {
+            totalJSHeapSize: number, 
+            usedJSHeapSize: number, 
+            jsHeapSizeLimit: number 
+          }
+        }).memory;
+
+        if (memory.jsHeapSizeLimit !== null && memory.usedJSHeapSize !== null) {
+          const memoryUsagePercentage = memory.usedJSHeapSize / memory.jsHeapSizeLimit;
+          console.log({...memory, memoryUsagePercentage});
```
CandelR: forget to remove this console.log

src/app/network/UploadManager.ts
```diff
@@ -101,6 +102,41 @@ class UploadManager {
     (fileData, next: (err: Error | null, res?: DriveFileData) => void) => {
       if (this.abortController?.signal.aborted ?? fileData.abortController?.signal.aborted) return;
 
+      if (window.performance && (window.performance as any).memory) {
+        const memory = (window.performance as unknown as {
```
CandelR: To avoid having to make these castings, we can overwrite the Performance interface by adding this to the globals.d.ts file:
```
interface Performance extends Performance {
  memory?: {
    jsHeapSizeLimit: number;
    totalJSHeapSize: number;
    usedJSHeapSize: number;
  };
}
```

This way we already have the typing added in case we want to use this API in any other part of the application.

CandelR (changes requested):

CandelR (approved):

PR #1062, opened 2024-03-20
[PB-1816]: feature/add payment method

rafijv:
- Add Workspace billing overview
- Add Workspace billing payment method


src/app/newSettings/components/BillingOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const local = localStorageService.get('i18nextLng') ?? navigator.language.split('-')[0];
+  const isFreeSuscription = plan.subscription?.type === 'free';
+
+  const subscriptionData: { amountInterval: string; interval: 'monthly' | 'yearly'; renewDate: string } | undefined =
+    getSubscriptionData({ userSubscription: plan.subscription, plan, local });
+  const nextBillingDate = getNextBillingDate(plan.subscription);
+  const [integerPart, decimalPart] = subscriptionData?.amountInterval?.split('.') ?? [];
+
+  return (
+    <section className="flex flex-row">
```
CandelR, line 25: Maybe you wanted to use the Section element that is already implemented?

src/app/newSettings/components/BillingOverview.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { RootState } from 'app/store';
+import { PlanState } from 'app/store/slices/plan';
+import { getSubscriptionData, getNextBillingDate } from '../utils/suscriptionUtils';
+import localStorageService from 'app/core/services/local-storage.service';
+import { bytesToString } from '../../drive/services/size.service';
+
+import Card from 'app/shared/components/Card';
+
+const BillingOverview = () => {
+  const plan = useSelector<RootState, PlanState>((state) => state.plan);
```
CandelR: Avoid attaching components to redux states, instead of adding it here, pass it through props.
Same with other props, if it something that is not related directly with the representation of a value in the component, try to pass it through props. 
Let's try to think as if the new components are going to go into the new component library, so it will be easier to separate what has to go into the props and what can go into the component.

src/app/newSettings/components/BillingPaymentMethodCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+        if (data.card) {
+          setState({ tag: 'ready', card: data.card as PaymentMethod['card'] });
+        } else if ('type' in data) {
+          setState({ tag: 'ready', type: data.type });
+        } else {
+          setState({ tag: 'empty' });
+        }
+      })
+      .catch(() => setState({ tag: 'empty' }));
+  }, []);
```
CandelR: Same as before, I guess with this component is the one that was in the old Settings. You can create a container that implements this call and so we take it out of the component and it is ready to move it to the library.

CandelR (changes requested):

src/app/newSettings/components/BillingOverview.tsx
rafijv: In this case I don't need that component, I could use the empty fragments of react or a div but I prefer to give it some semantics with the tags.

src/app/newSettings/components/BillingOverview.tsx
rafijv: This is already being used in other components, such as Overview, I understand that this will change when the new api calls are implemented and it is better to use this for now rather than mock data.

src/app/newSettings/components/BillingPaymentMethodCard.tsx
rafijv: Done

src/app/newSettings/components/BillingOverview.tsx
CandelR: Maybe in a `container` or `view`, if so, please, move it out of the `components` folder to its corresponding directory.

rafijv: Finally I have moved this logic to the view

CandelR (approved):

PR #1070, opened 2024-04-03
[_]: feature/master to staging

CandelR:
-  Merge last master changes to staging

**Changes:**
- Fixed bug that caused files/folders to fail to load when navigating directly to a folder's url
- Center the "Terms and conditions"/"Need help" labels in Auth view for mobile.
- The UploadManager modifies the upload queue concurrency on Chrome depending on the RAM usage
- New actions added, these actions would:
  - build and publish a Docker image with every push to the branch
  - add preview label when the PR is ready_for_review (this triggers the app creation)
  - clean up the created app whe the PR is closed
  - build and publish a Docker image when the PR is merged
  - enable slash commands to check and update the created environment
- World cloud security banner has been added.
- Fixed bug that moved the onboarding file upload button

PR #1071, opened 2024-04-03
[PB-1948] bugfix/downloading zip error for staging

CandelR:
- Fixed bug that caused the last page of the folder or file not to be included in the zip file
- Added tests

⚠️ Note:
- Wait for the merger of this PR https://github.com/internxt/drive-web/pull/1070 for review


rafijv (approved):

src/app/drive/services/foldersZip.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+async function addAllFoldersToZip(
+  currentAbsolutePath: string,
+  iterator: Iterator<DriveFolderData>,
+  zip: FlatFolderZip,
+): Promise<DriveFolderData[]> {
+  const path = currentAbsolutePath;
+  const allFolders: DriveFolderData[] = [];
+
+  const addFolderToZip = async (folder: DriveFolderData) => {
```
sg-gs: `async` is not needed if you are not / do not need to await `zip.addFolder`

sg-gs (approved):

PR #1088, opened 2024-04-16
[PB-1990]: feature/add tabs component

rafijv:
- Add tabs component


src/app/newSettings/Sections/Workspace/Members/components/ActivityTab.tsx
```diff
@@ -0,0 +1,7 @@
+import { ActivityTabProps } from '../../../../types';
+
+const ActivityTab = ({ user, activity }: ActivityTabProps): JSX.Element => {
+  return <div>ACTIVITY TAB</div>;
```
CandelR, line 4: Do the names of the tabs need translations or do they always have a fixed name in English?

rafijv: This is only temporary just to display something, here I have to create the whole Activity view.

CandelR (approved):

PR #1101, opened 2024-04-23
[MKT-315]: feat/star wars campaign

xabg2:
For the Star Wars campaign, we've added:
* New banner
* New exclusive theme for users who bought a subscription/lifetime plan with the Star Wars coupon.


masterprog-cmd:
⚠️  We have to update the SDK version before merge.

CandelR (approved):

PR #1140, opened 2024-06-03
[BR-188]: fix/update file preview metadata

xabg2:
- Updating metadata in the file preview when navigating with the arrow keys.
- Lifetime prices without decimals.

### DIFF
--- a/src/app/core/services/navigation.service.ts
+++ b/src/app/core/services/navigation.service.ts
@@ -4,6 +4,7 @@ import queryString from 'query-string';
 import { PATH_NAMES, serverPage } from '../../analytics/services/analytics.service';
 import { AppView } from '../types';
 import configService from './config.service';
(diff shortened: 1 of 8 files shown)


src/app/core/services/navigation.service.ts
```diff
@@ -57,6 +57,12 @@ const navigationService = {
     const lastSegment = pathname[pathname.length - 1];
     return lastSegment;
   },
+  replaceState(uuid: string | undefined): void {
+    const pathname = navigationService.history.location.pathname.split('/');
+    pathname[pathname.length - 1] = uuid ?? '';
```
CandelR: Add a try catch in order to handle if the pathname lenght was 0 or other unexpected case to avoid app crashes

CandelR (approved):

PR #1142, opened 2024-06-04
[_]: bugfix/Resolved duplicate route ids

CandelR:
- Added manage of duplicated routes


src/app/routes/paths.json
```diff
@@ -234,13 +234,13 @@
...
     {
-      "id": "share-photos-token",
+      "id": "share-photos-token-2",
       "layout": "share",
       "path": "/d/sh/photos/:token([a-z0-9]{24})/:code",
       "exact": false
     },
     {
-      "id": "share-photos-token",
+      "id": "share-photos-token-3",
```
sg-gs: I would remove all the /photos/ as those shares no longer exist

src/app/routes/paths.json
```diff
@@ -210,13 +210,13 @@
       "exact": false
     },
     {
-      "id": "share-folder-token",
+      "id": "share-folder-token-2",
```
sg-gs, line 207: We do not need /d/sh, we need /d/ and /sh/ (2 different)

sg-gs (changes requested):

src/app/routes/paths.json
CandelR: Don't we leave it to support any photos links that may still exist?

sg-gs (approved):

PR #1145, opened 2024-06-06
[_]: feat/custom context app name for tracks

xabg2:
'- Modifying the name of the app context that is sent to rudder with an event to filter the events with that name.


CandelR (approved):

sg-gs:
No longer needed @CandelR @masterprog-cmd

PR #1151, opened 2024-06-12
[PB-2115] feat: edit payment method for B2B and individual subscription

evillalba94:
#### Explanation
To implement this, a new endpoint was created that must receive the `paymentMethodId` and the `subscriptionType` (`business` or `individual`) on which the payment method must be updated.

It is important that when creating the payment method, a type field is created in the metadata to identify which subscription it was created for. Available options: `business` and `individual`. 

Once the payment method has been created, proceed to send it to the new endpoint.

Note: The SDK update must be taken into account for the consumption of the endpoint.
PR Sdk: https://github.com/internxt/sdk/pull/206

#### Endpoint 
URL: `/subscriptions/update-payment-method`
Method: POST
Body
```
    {
        paymentMethodId: { type: 'string' },
        subscriptionType: { type: 'string', enum: ['business', 'individual'] },
    }
```


evillalba94:
The adaptation was modified so this PR is not necessary.

PR #1164, opened 2024-06-25
[PB-2141] feature/design b2b review

rafijv:
Design adjustments to the new preference modal


src/app/newSettings/PreferencesDialog.tsx
```diff
@@ -53,50 +52,53 @@ const PreferencesDialog = (props: PreferencesDialogProps) => {
...
 
   return (
     <Modal
       maxWidth="max-w-4xl"
-      className="dark:dbg-surface m-0 flex h-640 shadow-sm"
+      className="m-0 flex h-640 overflow-hidden shadow-sm"
       isOpen={isPreferencesDialogOpen}
-      onClose={() => onClose()}
+      onClose={() => onClosePreferences()}
+      // preventClosing={isToastNotificationOpen}
```
CandelR: Does this have to work or do we remove the line?

src/app/newSettings/PreferencesDialog.tsx
```diff
@@ -53,50 +52,53 @@ const PreferencesDialog = (props: PreferencesDialogProps) => {
...
+        )}
+        {activeSection?.section === 'workspace' && activeSection?.subsection === 'overview' && (
+          <OverviewSection onClosePreferences={() => onClosePreferences()} />
+        )}
+        {activeSection?.section === 'workspace' && activeSection?.subsection === 'members' && (
+          <MembersSection onClosePreferences={() => onClosePreferences()} />
+        )}
         {activeSection?.section === 'workspace' && activeSection?.subsection === 'billing' && (
-          <BillingWorkspaceSection />
+          <BillingWorkspaceSection onClosePreferences={() => onClosePreferences()} />
```
CandelR: no need to add in a anonymous function, could pass the reference directly. That way is better for the performance, otherwise the anonymous function is destroyed and created for each render

src/app/newSettings/PreferencesDialog.tsx
rafijv: Removed

src/app/newSettings/PreferencesDialog.tsx
rafijv: Done

PR #1182, opened 2024-07-12
[_] fix/workspace testing

rafijv:
- fix fetch workspaces after accepting an invitation
- fix workspace breadcrumbs
- fix invite members input

### DIFF
--- a/src/app/shared/components/Breadcrumbs/Containers/BreadcrumbsDriveView.tsx
+++ b/src/app/shared/components/Breadcrumbs/Containers/BreadcrumbsDriveView.tsx
@@ -2,7 +2,8 @@ import navigationService from 'app/core/services/navigation.service';
 import { AppView } from 'app/core/types';
 import { FolderPath } from 'app/drive/types';
 import Breadcrumbs from 'app/shared/components/Breadcrumbs/Breadcrumbs';
-import { useAppDispatch } from 'app/store/hooks';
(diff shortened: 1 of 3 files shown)


src/app/shared/components/Breadcrumbs/Containers/BreadcrumbsDriveView.tsx
```diff
@@ -16,12 +17,16 @@ interface BreadcrumbsDriveViewProps {
...
 
   const breadcrumbDriveViewItems = (): BreadcrumbItemData[] => {
     const items: BreadcrumbItemData[] = [];
 
     if (namePath.length > 0) {
-      const firstPath = namePath[0];
+      const breadcrumbNamePath = [...namePath];
+      const parsedBreadcrumbNamePath =
+        selectedWorkspace && breadcrumbNamePath.length > 1 ? breadcrumbNamePath.slice(1) : breadcrumbNamePath;
+      const firstPath = parsedBreadcrumbNamePath[0];
```
CandelR, line 29: From what I see here you are removing the first element, haven't you found out why?
So it appears with the name Drive in the breadcrumb?

rafijv: The first element when calling the ancestors always corresponds to personal drive, in the case of workspaces I select the second one, which is the root of the workspace.
In this way Drive in the breadcrumbs is always the root folder of drive or workspace depending on what we have selected

PR #1186, opened 2024-07-18
[PB-2341] feature/workspaces url navigation

rafijv:
- add url navigation in workspaces

### DIFF
--- a/src/app/drive/components/FileViewer/FileViewerWrapper.tsx
+++ b/src/app/drive/components/FileViewer/FileViewerWrapper.tsx
@@ -82,7 +82,6 @@ const FileViewerWrapper = ({
 
   useEffect(() => {
     setBlob(null);
-    navigationService.replaceState(currentFile?.uuid);
     dispatch(uiActions.setFileViewerItem(currentFile));
     if (currentFile && !updateProgress && !isDownloadStarted) {
       setIsDownloadStarted(true);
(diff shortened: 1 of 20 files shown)


src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -82,7 +82,6 @@ const FileViewerWrapper = ({
 
   useEffect(() => {
     setBlob(null);
-    navigationService.replaceState(currentFile?.uuid);
```
CandelR, line 85: why has been removed?

CandelR (approved):

CandelR:
@rafijv review sonarcloud in order to not introduce new issues please 🙏

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
rafijv: Because when the file viewer was opened the url was changed again and that is already done correctly at the time of opening the file.
It was something that had been done before, I have been testing it and trying to understand why to do it twice but I have not found any reason.

CandelR: Perfect, thanks for the explanation 🚀

CandelR:
@rafijv Good job! 🚀

PR #1189, opened 2024-07-19
[PB-2341] feature/workspaces url navigationv

rafijv:
- fix navigation from workspace to personal space

### DIFF
--- a/src/app/drive/views/DriveView/DriveView.tsx
+++ b/src/app/drive/views/DriveView/DriveView.tsx
@@ -57,7 +57,8 @@ const DriveView = (props: DriveViewProps) => {
     if (isFolderView && itemUuid && workspaceUuid && !isSelectedWorkspace) {
       setWorkspaceWithUrl(workspaceUuid);
     } else if (isFolderView && itemUuid && !workspaceUuid) {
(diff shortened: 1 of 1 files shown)


src/app/drive/views/DriveView/DriveView.tsx
```diff
@@ -81,6 +85,14 @@ const DriveView = (props: DriveViewProps) => {
     }
   };
 
+  const setPersonalWithUrl = (itemUuid) => {
+    dispatch(workspacesActions.setCredentials(null));
+    dispatch(workspacesActions.setSelectedWorkspace(null));
+    localStorageService.set(STORAGE_KEYS.WORKSPACE_CREDENTIALS, 'null');
+    localStorageService.set(STORAGE_KEYS.B2B_WORKSPACE, 'null');
+    isFolderView ? goFolder(itemUuid) : showFile(itemUuid);
```
CandelR: I think it will be better to keep the navigation function out of this function that sets the workspace so that the function focuses the behaviour on one thing only

CandelR (changes requested):

src/app/drive/views/DriveView/DriveView.tsx

CandelR (approved):

PR #1203, opened 2024-07-25
[_] feature/add loader while switching workspace

rafijv:
- display spinner while switching workspace


CandelR (changes requested):
@rafijv  I have asked for a design to make this look better and have a better UX

src/app/store/slices/workspaces/workspacesStore.ts
```diff
@@ -97,6 +100,7 @@ const fetchCredentials = createAsyncThunk<void, undefined, { state: RootState }>
       dispatch(workspacesActions.setCredentials(credentials));
       localStorageService.set(STORAGE_KEYS.WORKSPACE_CREDENTIALS, JSON.stringify(credentials));
     }
+    dispatch(workspacesActions.setIsLoadingCredentials(false));
```
CandelR, line 103: we have to manage when this fails so that the loader doesn't get stuck in infinity.

rafijv: Done

rafijv:
> @rafijv I have asked for a design to make this look better and have a better UX

Done

CandelR (approved):

PR #1216, opened 2024-07-31
[PB-2372] feature/manage teams

rafijv:
- add create teams
- add list teams


src/app/i18n/locales/zh.json
```diff
@@ -1540,7 +1540,18 @@
         }
       },
       "teams": {
-        "title": "团队"
+        "title": "團隊",
```
CandelR, line 1543: these ch translations are exactly the same as those of tw. Is this correct?

src/app/newSettings/Sections/Workspace/Teams/TeamsSection.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const [createTeamDialogOpen, setCreateTeamDialogOpen] = useState(false);
+  const [newTeamName, setNewTeamName] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    getTeams();
+  }, []);
+
+  const getTeams = async () => {
+    if (selectedWorkspace) {
```
CandelR, line 29: I think it would be a better experience if we added a loader while fetching the teams

src/app/newSettings/Sections/Workspace/Teams/components/CreateTeamDialog.tsx
```diff
@@ -0,0 +1,51 @@
...
+  createTeam,
+}) => {
+  const { translate } = useTranslationContext();
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} width="w-96">
+      <h2 className="mb-5 text-2xl font-medium text-gray-100">
+        {translate('preferences.workspace.teams.createTeamDialog.title')}
+      </h2>
+      <Input
+        label={translate('preferences.workspace.teams.createTeamDialog.title')}
```
CandelR, line 30: Here is also the title of the modal, it has to show something like "Team name", right?

src/app/newSettings/Sections/Workspace/Teams/components/TeamsList.tsx
```diff
@@ -0,0 +1,54 @@
...
+        {teams.map((team) => {
+          const isLastTeam = teams.length === teams.indexOf(team) + 1;
+          return (
+            <div
+              key={team.team.id}
+              className={`grid h-12 cursor-pointer grid-cols-3 border-x border-b border-gray-10 bg-surface py-2 text-base font-medium hover:bg-gray-5 ${
+                isLastTeam && 'rounded-b-xl'
+              }`}
+            >
+              <div className="col-span-2 flex items-center pl-5">{team.team.name}</div>
```
CandelR, line 44: The default team name arrives `null`, is this a known problem?
We should check with someone in the backend to make sure it has a default name.
I also think something like a hyphen should be added in the case of no name so that it doesn't give the impression that it is not being displayed.

rafijv: I have already left it in a comment for backend, I think it should not be in this list since it is really the wokspace.
I keep it in consideration to solve it.

src/app/newSettings/Sections/Workspace/Teams/components/CreateTeamDialog.tsx
rafijv: If I understand you correctly, that is shown in the team detail, which is the next thing I'm working on, it will be included in the next PR.

src/app/newSettings/Sections/Workspace/Teams/TeamsSection.tsx
rafijv: I realized while doing what I am doing now and will be included in the next PR.

src/app/i18n/locales/zh.json
rafijv: I've changed the tw translation

rafijv:
If it seems okay to you, approve this PR while I clarify the Teams listing with the backend. I'll have the other thing ready for the next PR.

rafijv:
I am closing this PR and creating another one that addresses the issues discussed in this one and adds more functionality

PR #1228, opened 2024-08-13
[PB-2513] fix/clear workspace on logout

rafijv:
- add reset WorkspaceState

### DIFF
--- a/src/app/store/slices/user/index.ts
+++ b/src/app/store/slices/user/index.ts
@@ -19,6 +19,7 @@ import { sessionSelectors } from '../session/session.selectors';
 import { storageActions } from '../storage';
 import { teamActions } from '../team';
 import { uiActions } from '../ui';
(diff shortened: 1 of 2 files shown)


src/app/store/slices/user/index.ts
```diff
@@ -128,6 +129,7 @@ export const logoutThunk = createAsyncThunk<void, void, { state: RootState }>(
     dispatch(storageActions.resetState());
     dispatch(uiActions.resetState());
     dispatch(referralsActions.resetState());
+    dispatch(workspacesActions.resetState());
```
CandelR, line 132: Seeing this comes to my mind, does it also clean the workspace data in the local storage?

rafijv: No, that was already being done but this one was missing.

CandelR (approved):

PR #1233, opened 2024-08-14
[PB-985]: feature/playwright e2e tests

savilashokworks:
## Description

Added setup of playwright and test for login and sign up

## Related Issues

-
## Related Pull Requests

-

## Checklist

- [x] QA Passed

## How Has This Been Tested?

-

## Additional Notes

-


tests/helper/staticData.ts
```diff
@@ -0,0 +1,38 @@
+export const staticData = {
+  //CREDENTIALS
+  email: 'testingb2b@inxt.com',
+  password: 'test123.',
```
CandelR: Maybe we could make part of the email dynamic, so that each time the signup is tested as well

tests/specs/login.setup.ts
```diff
@@ -0,0 +1,18 @@
+import { test as setup } from '@playwright/test';
+import { loginPage } from '../pages/loginPage';
+import { staticData } from '../helper/staticData';
+
+const authFile = './tests/specs/playwright/.auth/user.json';
```
CandelR: Have you tried to automate the login for each test so you don't have to have the user data in a file?  
This may limit some days to do tests, because apart from putting sensitive data of the user in question, it will be filled with test data when uploading files or folders and perform certain actions.
This together with the dynamic email and the sign up, you will be able to test each run with an independent user.
Let me know if I miss anything 😄

tests/locators/driveLocators.ts
```diff
@@ -0,0 +1,3 @@
+export const driveLocators = {
+  driveTitle: '[title="Drive"]',
+};
```
CandelR: I don't see the need to include the type in the name of each type of file, as they are already in the directory, it is already understood that they are of that type. 
For example, the locators, `driveLocators.ts` -> `drive.ts` since they are already in `locators/drive.ts`. :

CandelR (commented):
Good work @savilashokworks. This is looking good.

tests/specs/playwright/.auth/credentials.json
```diff
@@ -0,0 +1 @@
+{"email":"Art.Goyette77@yahoo.com","password":"whSneqH707MPwbp"}
```
CandelR: Be careful when exposing sensitive data such as username/password. The user has to be dynamic in every test.
With cypress we wrote the generated user and password dynamically in a json so that after registering the user, the other tests could use the login and re-run the test

tests/helper/staticData.ts
savilashokworks: Ramon!! Yes, it makes perfect sense and I should've done that way! Great, I'll include it in my next commit ;)

tests/locators/driveLocators.ts
savilashokworks: Great!! Thanks a lot for that correction!

tests/specs/login.setup.ts
savilashokworks: Let me know if I'm right. You want to log in before each test (no matter if the tests include "log in" tests or not) ❓ 

In that case, I think I still need the login.setup.ts so as to create the user, log in and save the cookies (that's what it's meant to do). Then I can just ignore the cookies stored and log in in a before hook. I would still need to figure out how to pass the credentials into the tests without revealing data, is that right?

But the setup.ts speeds the whole testing up by using the cookies instead of logging in "manually", that's the benefit of it, but if you're ok with logging in before each test, it can be done, definitely. Unless you want to get rid of the setup.ts D:

tests/pages/drivePage.ts
```diff
@@ -64,36 +93,82 @@ export class DrivePage {
...
     expect(buttonText).toEqual('Cancel');
     await cancelButton.click();
     return this.folderCreationModal;
   }
   async rightClickOnBody() {
-    await this.driveTitle.waitFor({ state: 'visible' });
+    await expect(this.driveTitle).toBeVisible({ timeout: 10000 });
     await this.page.locator('body').click({ button: 'right' });
+    await this.page.waitForTimeout(500);
+    //await expect(this.rightClickOnBodyModal).toBeVisible()
```
CandelR: forgot to remove this commeted line.

tests/pages/loginPage.ts
```diff
@@ -65,35 +69,43 @@ export class loginPage {
...
     const createAccountText = await this.createAccount.textContent();
+    expect(dontHaveAccountText).toEqual("Don't have an account?");
+    expect(createAccountText).toEqual('Create account');
+    await expect(this.createAccount).toBeEnabled();
     await this.createAccount.click();
-    await this.createAccounTitle.waitFor({ state: 'visible', timeout: 3000 });
-    const createAccountTitle = await this.createAccounTitle.textContent();
+    await expect(this.createAccounTitle).toBeVisible();
+    const createAccountHeader = this.createAccounTitle.textContent();
+    //console.log(createAccountHeader)
```
CandelR: forgot to remove this commented log.

playwright.config.ts
```diff
@@ -27,7 +29,7 @@ export default defineConfig({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: 'https://staging.drive.internxt.com/login',
+    baseURL: 'https://drive.internxt.com/login',
```
CandelR: You need to run the test with the current branch in a local environment, not pointing to the production frontend, the backend could be the production one but not the frontend. 
If you point to the production frontend you are not testing the actual changes.

tests/helper/staticData.ts
```diff
@@ -7,7 +7,7 @@ export const staticData = {
   //URLS
   passwordRecoveryLinkURL: 'https://drive.internxt.com/recovery-link/',
   signUpURL: 'https://drive.internxt.com/new',
-  driveURL: 'https://staging.drive.internxt.com/',
+  driveURL: 'https://drive.internxt.com/',
```
CandelR: Better add a constant with the baseURL and reuse it everywhere, otherwise it's going to be a hassle if you have to add any changes.

tests/specs/login.setup.ts
```diff
@@ -36,10 +36,11 @@ setup('Creating new user and logging in', async ({ browser }) => {
 
   const loginpage = new loginPage(newPage);
 
-  await newPage.goto('https://staging.drive.internxt.com/login');
+  await newPage.goto('https://drive.internxt.com/login');
   await expect(newPage).toHaveURL(/.*login/);
 
-  const endpointPromise = newPage.waitForResponse('https://drive.internxt.com/api/access');
+  //const endpointPromise = newPage.waitForResponse('https://drive.internxt.com/api/access');
+  const endpointPromise = newPage.waitForResponse('https://gateway.internxt.com/drive-old/access');
```
CandelR: Add all the urls to constants

tests/specs/login.setup.ts
```diff
@@ -36,10 +36,11 @@ setup('Creating new user and logging in', async ({ browser }) => {
 
   const loginpage = new loginPage(newPage);
 
-  await newPage.goto('https://staging.drive.internxt.com/login');
+  await newPage.goto('https://drive.internxt.com/login');
```
CandelR: same here, use a constnat for urls

tests/specs/login.setup.ts
```diff
@@ -19,9 +19,9 @@ setup('Creating new user and logging in', async ({ browser }) => {
   const userCredentials = { email, password };
   fs.writeFileSync(credentialsFile, JSON.stringify(userCredentials));
 
-  console.log(`User data, new user: ${email}`);
+  console.log(`User data, new user: ${email}, password: ${password}`);
 
-  await page.goto('https://staging.drive.internxt.com/new');
+  await page.goto('https://drive.internxt.com/new');
```
CandelR: remember to use the constnat for urls and point to local environment

CandelR (changes requested):

CandelR:
@QAZebass What was the status of the tests? To see if we can add it to Master to do more e2e tests

Jona-Internxt (approved):

CandelR (approved):

Issue #1243, opened 2024-08-23
Localization error for folder not found URL


claell:
https://drive.internxt.com/not-found?itemType=folder

![grafik](https://github.com/user-attachments/assets/f996c62d-c02c-4816-9296-7ced63f40693)

I am in Germany, but the error message is displayed in Dutch.

claell:
Same for files, BTW: https://drive.internxt.com/not-found?itemType=file

claell:
Issue 1 is from here: https://github.com/internxt/drive-web/blob/7e883f7234bffc4bc4d5124abeb3d71167f24f8d/src/app/i18n/locales/de.json#L1399
Issue 2 is from https://github.com/internxt/drive-web/blob/7e883f7234bffc4bc4d5124abeb3d71167f24f8d/src/app/i18n/locales/de.json#L1402

claell:
Confirmed this is fixed in production for me. Thanks again!

PR #1250, opened 2024-09-02
[PB-2715] fix/check de translation

rafijv:
- fix item not found de translation


CandelR (approved):

claell:
Nice, thanks for fixing this so quickly, @rafijv @CandelR!

Linking #1243 which is fixed by this.

rafijv:
@claell Thanks for report!

PR #1252, opened 2024-09-04
[_]: fix/one time payments using coupons with 100% OFF

xabg2:
Fixing one-time payments when using a 100% OFF coupon.


xabg2:
⚠️ This PR needs this SDK version before merging:
https://github.com/internxt/sdk/pull/241

**Update:**
The SDK version has been updated.

CandelR (approved):

PR #1257, opened 2024-09-05
[MKT-448]: feat/b2b integrated checkout (logic)

xabg2:
Handling all logic to allow the user to purchase a b2b plan:
* Redirect the user to our integrated checkout instead of the Stripe Checkout
* How many seats the user want for his b2b subscription
* Updating b2b subscription from int checkout if needed
* Optional fields to fill it with the company data (company name and tax ID)


masterprog-cmd:
⚠️ Do not merge this PR before updating the SDK.
Related PR: https://github.com/internxt/sdk/pull/242

src/app/payment/utils/getProductAmount.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (couponCodeData?.amountOff) {
+    finalAmount = (amount - couponCodeData.amountOff / 100) * users;
+  } else if (couponCodeData?.percentOff) {
+    const discount = 100 - couponCodeData.percentOff;
+    finalAmount = ((amount * discount) / 100) * users;
+  } else {
+    finalAmount = amount * users;
+  }
+
+  // Formatear la cantidad para que elimine .00 si no es necesario
```
CandelR: Leave comments in English

src/app/payment/components/checkout/SelectSeatsComponent.tsx
```diff
@@ -1,66 +1,99 @@
...
+  minSeats,
   seats,
-  onUsersChange,
+  onSeatsChange,
 }: SelectUsersComponentProps): JSX.Element => {
+  const { translate } = useTranslationContext();
+  const [totalUsers, setTotalUsers] = useState<number>(seats);
+  const [showButton, setShowButton] = useState<boolean>(false);
+
+  const onInputChanges = () => {
```
CandelR: it seems that the name of the function does not reflect the behaviour of the function

src/app/newSettings/Sections/Account/Plans/PlansSection.tsx
```diff
@@ -211,37 +211,22 @@ const PlansSection = ({ changeSection, onClosePreferences }: PlansSectionProps)
...
-        onClosePreferences();
-        navigationService.push(AppView.Checkout, {
-          planId: priceId,
-          currency: currency,
-        });
-      } else {
-        await handleCheckoutSession({ priceId, currency, userEmail: user.email, mode });
-      }
-
+    if (!isCurrentPlanTypeSubscription || isLifetimeIntervalSelected) {
```
CandelR, line 234: Isn't this check redundant?
If it is not subscription, isn't it always lifetime?

src/app/payment/components/checkout/SelectSeatsComponent.tsx
```diff
@@ -1,66 +1,99 @@
...
+            e.preventDefault();
+            onSeatsChange(seats + 1);
+            setTotalUsers(seats + 1);
+          }}
+          className="flex h-full flex-col items-center justify-center rounded-r-lg px-4 hover:bg-gray-10"
+        >
+          +
+        </button>
+      </div>
+      {showButton ? (
```
CandelR: I think it would be better to change the name of the conditional constant to something not so generic, like: showApplyButton, or something that identifies which button is going to be shown

src/app/payment/store/checkoutReducer.ts
```diff
@@ -6,6 +6,12 @@ export const initialStateForCheckout: State = {
   promoCodeName: '',
   avatarBlob: null,
   isPaying: false,
+  isCheckoutReadyToRender: false,
+  isUpdateSubscriptionDialogOpen: false,
+  isUpdatingSubscription: false,
+  isUpsellSwitchActivated: false,
+  country: '',
+  prices: [],
```
CandelR, line 14: why can it have a price list instead of a price?

src/app/payment/views/IntegratedCheckoutView/CheckoutViewWrapper.tsx
```diff
@@ -82,10 +80,9 @@ export interface CheckoutViewManager {
 }
 
 const ONE_YEAR_IN_MONTHS = 12;
-
 const IS_PRODUCTION = envService.isProduction();
-
 const RETURN_URL_DOMAIN = IS_PRODUCTION ? process.env.REACT_APP_HOSTNAME : 'http://localhost:3000';
+const STATUS_CODE_ERROR = [409, 422];
```
CandelR: Perhaps it would be better to create an object, so that the key that identifies the reason for the status code, I think, would be clearer. 
Then you can extract the values from the object and use the list for comparison

src/app/newSettings/Sections/Account/Plans/PlansSection.tsx
xabg2: `isCurrentPlanTypeSubscription` checks if the user already has a subscription. If the user has a subscription, we have to check if it is a lifetime plan to redirect the user to our integrated checkout or update the subscription using the dialog.

src/app/payment/store/checkoutReducer.ts
xabg2: Because we need all prices to pass them to the ChangePlanDialog (Idk why we pass all prices instead the one we want to update). I can update it and pass the Price directly, but I think it does not fits in this PR, wdyt? @CandelR

src/app/payment/views/IntegratedCheckoutView/CheckoutViewWrapper.tsx
xabg2: I agree 👍🏽

src/app/newSettings/Sections/Account/Plans/PlansSection.tsx
CandelR: Ah okay! all clear! 💯

src/app/payment/store/checkoutReducer.ts
CandelR: Okay, we'll discuss whether it's better to pass only one price, I don't have enough context right now. 
And if we finally have to change it, we'll change it in another task.

src/app/payment/views/IntegratedCheckoutView/CheckoutViewWrapper.tsx
```diff
@@ -432,39 +448,18 @@ const CheckoutViewWrapper = () => {
...
-    return userData;
-  };
-
   const handlePromoCodeError = (err: unknown, showNotification?: boolean) => {
     const error = err as Error;
-    const errorMessage = error.message.includes('Promotion code with an id')
-      ? error.message
-      : 'Something went wrong, try again later';
+    const statusCode = (err as any).status;
+    let errorMessage = 'Something went wrong, try again later';
```
CandelR, line 454: A more specific message would be better, so that it is known that the promo code is not valid, and it should be added to the translations

CandelR (approved):

PR #1258, opened 2024-09-05
[PB-2691]: feat/remove obsolete teams section

rafijv:
- Removed unused teams code


src/app/store/slices/storage/storage.selectors.ts
```diff
@@ -22,10 +21,7 @@ const storageSelectors = {
     return state.storage.namePath.reduce((t, path) => `${t}${path.name}/`, '/');
   },
   bucket(state: RootState): string {
-    const { team } = state.team;
-    const isTeam: boolean = sessionSelectors.isTeam(state);
-    // TODO: NEED TO CHANGE THIS BY WORKSPACE BUCKET
-    return (isTeam ? team?.bucket : state.user.user?.bucket) ?? '';
+    return state.user.user?.bucket ?? '';
```
CandelR, line 24: Is it not necessary to check if the workspace is selected in order to use the corresponding bucket? I mean for the TODO

rafijv: I think the comment is old thinking in a way how it would work but it is not necessary

CandelR (approved):

PR #1261, opened 2024-09-10
[PB-2458]: fix/display right-click menu

rafijv:
- Fix the height at which the right-click menu is displayed up or down

### DIFF
--- a/src/app/shared/components/List/ListItem.tsx
+++ b/src/app/shared/components/List/ListItem.tsx
@@ -37,6 +37,7 @@ interface ItemProps<T> {
 }
 
 const MENU_BUTTON_HEIGHT = 40;
+const TOP_FIXED_HEIGHT = 270;
 
 // This is used to get the size of the menu item list and adjust its position depending on where you are trying to open it.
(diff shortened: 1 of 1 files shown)


src/app/shared/components/List/ListItem.tsx
```diff
@@ -141,7 +142,7 @@ export default function ListItem<T extends { id: string }>({
     }
 
     if (event.clientY + childHeight > innerHeight) {
-      y = y - childHeight;
+      y = y - childHeight + TOP_FIXED_HEIGHT;
```
CandelR: This solution breaks the behaviour of unfolding upwards when there is no space underneath.
I would look for something more along the lines of not overflowing the limits and if there is no space it shrinks but still scrolls.

CandelR (changes requested):

src/app/shared/components/List/ListItem.tsx
rafijv: The solution applied is to fix the height at which the menu is displayed up or down so that it is not hidden.

CandelR (approved):

PR #1265, opened 2024-09-13
[PB-469]: fix/breadcrumbs and New folder dialog name issues

xabg2:
Fixed bug when renaming a folder from breadcrumbs and navigate that showed the previous folder name.


src/app/drive/views/DriveView/DriveView.tsx
```diff
@@ -104,6 +104,7 @@ const DriveView = (props: DriveViewProps) => {
   const goFolder = async (folderUuid: string, workspacesToken?: string) => {
     try {
       const folderMeta = await newStorageService.getFolderMeta(folderUuid, workspacesToken);
+      dispatch(uiActions.setCurrentEditingBreadcrumbNameDirty(''));
```
CandelR: Check if we can reset the `BreadcrumbNameDirty` in the file when we are using it

xabg2: Solved @CandelR

src/app/shared/components/Breadcrumbs/Containers/BreadcrumbsDriveView.tsx
```diff
@@ -18,6 +21,15 @@ const BreadcrumbsDriveView = (props: BreadcrumbsDriveViewProps) => {
   const { namePath } = props;
   const dispatch = useAppDispatch();
   const { selectedWorkspace } = useAppSelector((state: RootState) => state.workspaces);
+  const currentBreadcrumb = namePath[namePath.length - 1];
+  const { breadcrumbDirtyName } = useDriveItemStoreProps();
+
+  useEffect(() => {
+    if (breadcrumbDirtyName) {
+      getAncestorsAndSetNamePath(currentBreadcrumb.uuid, dispatch);
+      dispatch(uiActions.setCurrentEditingBreadcrumbNameDirty(''));
```
CandelR: If the currentEditingBreadcrumbNameDirty is no longer in use, check if you can remove it and update the breadcrumb during the flow when renaming folders from the breadcrumb.

CandelR (approved):

PR #1276, opened 2024-09-13
[PB-2262]: fix/design workspace issues

xabg2:
Some small issues about the workspaces have been fixed:
* Overview title vertical padding
* Clicking on the overview boxes should redirect users to each subsection
* Member overflowing name
* Member info card spacing
* Option menu states
* Invoice table styles
* The Owner tag has been upgraded to Manager, as the Owner tag should not appear on Teams.


CandelR (approved):
Perfect, you can pass @savilashokworks  the url of the preview so he can check that everything is ok there if it is not necessary to deploy it in staging

Issue #1267, opened 2024-09-13
UI issue for restoring files from drive


claell:
![grafik](https://github.com/user-attachments/assets/3cb1e6f4-2ac1-401f-b0d8-bad93458aa7e)

The right button is expanding too much (possibly due to long localized string).

fvsegarra:
in progress, will be fixed in the next few days
thanks for reporting! :)

claell:
Confirmed fixed with #1280, although that uses a slight workaround and removes the "here" from the string.

PR #1280, opened 2024-09-17
[BR-489]: fix/DE Web restore files from trash text surpasses box width

xabg2:
Fixed the overflow of the restore items from trash button in the move/restore items dialog for the German language.


CandelR (approved):

claell:
@masterprog-cmd thanks for the super quick fix of my issue #1267, which I only saw just now.

PR #1282, opened 2024-09-18
[PB-2375] bugfix/Check duplicated items before uploading on Drive web

CandelR:
- Added new call to check for duplicate items and manage duplicate items.


src/app/store/slices/storage/fileUtils/checkDuplicatedFiles.ts
```diff
@@ -0,0 +1,50 @@
...
+      filesWithoutDuplicates: files,
+    } as DuplicatedFilesResult;
+  }
+
+  const filesToUploadParsedToCheck = files.map((file) => {
+    const { filename, extension } = itemUtils.getFilenameAndExt(file.name);
+    return { plainName: filename, type: extension };
+  });
+
+  const checkDuplicatedFileResponse = await newStorageService.checkDuplicatedFiles(
```
xabg2: Maybe you can extract `{ existentFiles }` directly from here. Or cast it `{ existentFiles: duplicatedFilesResponse }`, but I think existentFiles is enough to understand what this call returns

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -11,10 +11,14 @@ import tasksService from '../../../../tasks/services/tasks.service';
...
 import { planThunks } from '../../plan';
 import workspacesSelectors from '../../workspaces/workspaces.selectors';
+
+import { checkFolderDuplicated } from '../folderUtils/checkFolderDuplicated';
+import { getUniqueFolderName } from '../folderUtils/getUniqueFolderName';
 import { StorageState } from '../storage.model';
 import { deleteItemsThunk } from './deleteItemsThunk';
 import { uploadItemsParallelThunk } from './uploadItemsThunk';
 
+// TODO: REMOVE IROOT from this file, it is in types.ts
```
xabg2, line 21: Can you delete this interface, or is it imported in a pile of files?

xabg2 (approved):

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
CandelR: Right, it is used in a lot of files and I didn't want to increase the size of the changes and make it difficult to review the PR. We will do it in a separate PR so that you can more easily see the changes with respect to this

PR #1286, opened 2024-09-24
Fix typo in en.json

claell:
Just a quick fix that hopefully makes it easier to ship, instead of reporting it through the customer support or other places.


claell:
("centre" is British English, so not entirely wrong. I think, "center" in American English is the preferred way, here, though.)

CandelR:
Hey @claell , thanks for the input.
I am closing this PR as it is more of a preference problem than a bug

PR #1289, opened 2024-09-27
[PB-2980]: fix/show item name in rename item dialog in Recents view

xabg2:
Show the item name in the Rename Item dialog when opened from the Recents section.


src/app/drive/components/EditItemNameDialog/EditItemNameDialog.tsx
```diff
@@ -22,9 +22,10 @@ const EditItemNameDialog: FC<EditItemNameDialogProps> = ({ item, isOpen, resourc
   const [error, setError] = useState('');
   const [isLoading, setIsLoading] = useState(false);
   const [newItemName, setNewItemName] = useState('');
+  const itemName = item?.plainName ?? item?.plain_name ?? '';
```
CandelR: Oops, do we still have any cases where we get the plain name in snake_case? 
Can you check if it is feasible to remove this field using this fix?

CandelR (approved):

PR #1291, opened 2024-09-27
[PB-2666]: Feat/add argon2 and kyber

TamaraFinogina:
PR switches drive-web to Argon2 and adds a post-quantum layer [PB-2666](https://inxt.atlassian.net/browse/PB-2666)

[PB-2666]: https://inxt.atlassian.net/browse/PB-2666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

**Switching to Argon2.**

We now use `hash = PBKDF2(pwd, salt)`. Unfortunately, the parameters are outdated, so we need to upgrade. The idea is to compute Argon2 for all new users and update the hash after login for old ones. 

1.  If no salt is given (new user), then it samples argon_salt, computes `hash = Argon2 ( pwd, argon_salt)` and sends `(hash, 'argon2id$argoin_salt)` to the database
2. if salt starts with `argon2id$`, then it's already migrated user and we compute `hash = Argon2( pwd, argon_salt)`
3. Else, then it's the user that hasn't migrated yet.  We compute `hash = PBKDF2(pwd, salt)`

**NOTE:**  Argon2 follows password [hash competition output format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md):  the output will be `$argon2id$v=19$m=65536,t=2,p=1$gZiV/M1gPc22ElAH/Jh1Hw$CWOrkoo7oJBQ/iyh7uJ0LO2aLEfrHwTWllSAxT0zRno` or the same line but in HEX.

**Add a post-quantum layer.**

Quantum computers will only be a problem for asymmetric encryption, which we use when we send encrypted mnemonics (ecc from OpenPGP in our case). The idea is to use hybrid encryption - ecc and KyberKEM - for sending the mnemonic. Thus, if ecc is broken or Kyber needs different parameters for security, the mnemonic is still safe.

For that, each user should additionally have a Kyber keys pair `(publicKyberKey, privateKyberKey)`. Then, we split mnenomic into two parts (`mnemonic = part1 XOR part2`) and encrypt `par1` with ecc and `part2` with Kyber. 

Encrypt:

```
(ciphertext, secret) await kem.encapsulate(publicKyberKey)
longSecret = await extendSecret(secret, len(mnemonic))
message = mnemonic XOR longSecret
eccCiphertext =  await openpgp.encrypt( message, publicKey)
send `(eccCiphertext, ciphertext)
```

Decrypt: 

```
secret  = await kem.decapsulate(ciphertext, privateKyberKey)
message  =  await openpgp.decrypt(eccCiphertext, privateKey)
longSecret = await extendSecret(secret, len(message))
mnemonic = message XOR longSecret
```

**NOTE 1:**  KyberKEM always outputs a fixed-length secret, but mnemonic can have different sizes. Hence, we extend it with the `extendSecret` function (that is just Blake3 hash).  It can also shrink the secret if needed, but usually mnemonic is bigger than the secret.

**NOTE 2:** During the migration period, we must support both old and new sharing methods. I've added a flag `hybridModeEnabled` to differentiate. But we can remove it and use ciphertext size or a prefix instead.


sg-gs:
@TamaraFinogina Please provide a description

src/app/crypto/services/pgp.service.ts
```diff
@@ -15,53 +19,95 @@ export async function generateNewKeys(): Promise<{
...
   const publicKey = await openpgp.readKey({ armoredKey: publicKeyArmored });
 
+  const publicKyberKey = Buffer.from(publicKyberKeyBase64, 'base64');
+  const { ciphertext, sharedSecret: secret } = await kem.encapsulate(publicKyberKey);
+  const kyberCiphertextStr = Buffer.from(ciphertext).toString('base64');
+  const secretHex = Buffer.from(secret).toString('hex');
+  const messageHex = Buffer.from(message).toString('hex');
+
+  // message should be the same length as secret, which is 256 bits
+  const xoredMessage = XORhex(messageHex, secretHex);
```
CandelR, line 69: Will they always be the same length?
Maybe we should check in case they are not, so that it throws an error.

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -102,17 +103,19 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
...
   };
 
-  const getUserPublicKey = async (email: string): Promise<string> => {
+  const getUserPublicKey = async (email: string): Promise<{ publicKey: string; publicKyberKey: string }> => {
     let publicKey = '';
+    let publicKyberKey = '';
     try {
       const publicKeyResponse = await userService.getPublicKeyByEmail(email);
       publicKey = publicKeyResponse.publicKey;
+      publicKyberKey = publicKeyResponse.publicKyberKey;
```
CandelR: the backend is already prepared to store this new key?

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
```diff
@@ -157,15 +160,16 @@ const ShareInviteDialog = (props: ShareInviteDialogProps): JSX.Element => {
...
-      if (!publicKey && !preCreateUsers) {
+      const { publicKey, publicKyberKey } = await getUserPublicKey(email);
+      if (!publicKey && !preCreateUsers && !publicKyberKey) {
         isThereAnyNewUser = true;
       }
       usersList.push({
         email,
         userRole,
-        isNewUser: !publicKey,
+        isNewUser: !publicKey || !publicKyberKey,
```
CandelR: The user will have both keys always?

test/unit/services/pgp.service.test.ts
```diff
@@ -17,32 +21,55 @@ describe('Encryption and Decryption', () => {
...
     const encryptedMessage = await encryptMessageWithPublicKey({
       message,
       publicKeyInBase64,
+      publicKyberKeyBase64,
     });
 
     expect(encryptedMessage).toBeDefined();
   });
 
+  it('xor should work', async () => {
```
CandelR: It is too generic a test, separate it into different tests in which the description of what it has to achieve is a little more specific

.vscode/settings.json
```diff
@@ -1,11 +1,10 @@
...
\ No newline at end of file
+  "editor.formatOnPaste": true,
+  "editor.formatOnSave": true,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "explicit",
+    "source.fixAll.format": "explicit"
+  },
+  "svg.preview.background": "editor"
+}
```
CandelR: This has to be added to the gitignore, as it is part of the vs code configuration.

package.json
```diff
@@ -80,25 +83,27 @@
...
   "scripts": {
+    "postinstall": "patch-package",
     "preinstall": "node scripts/use-yarn.js",
     "prepare": "husky install",
     "dev": "craco start",
     "start": "craco start",
     "build": "craco build",
     "vercel:install": "yarn run add:npmrc && yarn install",
-    "test:unit": "jest src/ && jest test/unit",
+    "test:unit": "vitest run test/unit && vitest run src/ ",
```
CandelR: Interesting change, with Vitest the problems with global meta variable has gone?

src/app/crypto/services/pgp.service.ts
```diff
@@ -15,53 +19,95 @@ export async function generateNewKeys(): Promise<{
...
+
   return {
     privateKeyArmored: privateKey,
     publicKeyArmored: Buffer.from(publicKey).toString('base64'),
     revocationCertificate: Buffer.from(revocationCertificate).toString('base64'),
+    publicKyberKeyBase64: Buffer.from(publicKyberKey).toString('base64'),
+    privateKyberKeyBase64: Buffer.from(privateKyberKey).toString('base64'),
   };
 }
+export function XORhex(a: string, b: string): string {
```
CandelR, line 39: Maybe it would be good to add jsdoc for this function.

src/app/crypto/services/pgp.service.ts
```diff
@@ -15,53 +19,95 @@ export async function generateNewKeys(): Promise<{
...
+  let res = '',
+    i = a.length,
+    j = b.length;
+  while (i-- > 0 && j-- > 0) res = (parseInt(a.charAt(i), 16) ^ parseInt(b.charAt(j), 16)).toString(16) + res;
```
CandelR, line 46: if the lengths of the chains are different, this will end before they have travelled the longest chain. Is this the desired behaviour?

src/app/crypto/services/pgp.service.ts
TamaraFinogina: Yes, for XOR to work, both inputs must be the same length. Which error should be thrown?

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
TamaraFinogina: Don't think so

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
TamaraFinogina: That's the part I'm still thinking about - migration, when we only will have 1 key out of 2

package.json
TamaraFinogina: meta imports you mean? Yes.

src/app/crypto/services/pgp.service.ts
TamaraFinogina: Inputs should always be of identical length

src/app/crypto/services/pgp.service.ts
CandelR: I think that a custom error in which it is indicated that the length is not equal and that it should be. Provided that it has to be the case of course, that if the length is wrong, it will give problems. 
This way the error is already managed in the place where the function is called, so that there are no inconsistencies and problems with the encryption, before encrypting something wrong, better that it gives an error and warns to be able to see why the length of the strings is different

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
CandelR: So this task is blocked until the backend is ready?

src/app/crypto/services/pgp.service.ts
CandelR: Better to handle the possible undesired case where the length of the strings is not identical in order to save possible problems. 
If you control this by throwing an error as discussed above, this problem would be solved

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
TamaraFinogina: yes

src/app/crypto/services/pgp.service.ts
TamaraFinogina: @CandelR I've applied your suggestions

src/app/crypto/services/pgp.service.ts
```diff
@@ -15,53 +19,112 @@ export async function generateNewKeys(): Promise<{
...
 
   const publicKeyArmored = Buffer.from(publicKeyInBase64, 'base64').toString();
   const publicKey = await openpgp.readKey({ armoredKey: publicKeyArmored });
 
+  const publicKyberKey = Buffer.from(publicKyberKeyBase64, 'base64');
+  const { ciphertext, sharedSecret: secret } = await kem.encapsulate(new Uint8Array(publicKyberKey));
+  const kyberCiphertextStr = Buffer.from(ciphertext).toString('base64');
+  const secretHex = Buffer.from(secret).toString('hex');
+  const messageHex = Buffer.from(message).toString('hex');
+  console.log(`secret length is ${secret.length} and hex len is ${secretHex.length}`);
```
CandelR: Forgot to remove this console.log

CandelR (approved):

src/app/crypto/services/pgp.service.ts
TamaraFinogina: True! Nice catch!

TamaraFinogina: @CandelR Done

patches/@internxt+sdk+1.5.17.patch
sg-gs, line 1: Why is this file here?

sg-gs (approved):

patches/@internxt+sdk+1.5.17.patch
TamaraFinogina: I had to modify sdk, so without a patch this PR won't build

xabg2 (approved):

src/app/auth/services/auth.service.ts
```diff
@@ -150,6 +160,10 @@ export const doLogin = async (
       localStorageService.set('xMnemonic', clearMnemonic);
       localStorageService.set('xNewToken', newToken);
 
+      const argon2Hash = await passToHash({ password });
+
+      authClient.upgradeHash(argon2Hash.hash);
```
sg-gs: Be careful, according to the SDK [PR](https://github.com/internxt/sdk/pull/251/files) I have seen, this is a promise and you should `await` it.

sg-gs (changes requested):

sg-gs:
Hey @xabg2 @CandelR, 
We should review each line more carefully than a typical Pull Request as this could be a breaking change in the way we are modifying the involved cryptography for every customer

sg-gs:
Let's aim to increase the testing coverage to 80% @TamaraFinogina

src/app/auth/components/AuthSideInfo/AuthSideInfo.scss
```diff
@@ -1,5 +1,5 @@
 
-.sideinfo_background {
+.sideinfo-background {
```
TamaraFinogina: due to "stylelint-scss" package upgrade

src/app/auth/components/ChangePassword/ChangePassword.tsx
```diff
@@ -1,19 +1,19 @@
 import { Dispatch, SetStateAction, useState, RefObject, createRef, useEffect } from 'react';
 import { Link } from 'react-router-dom';
-import authService from 'app/auth/services/auth.service';
-import notificationsService, { ToastType } from 'app/notifications/services/notifications.service';
+import authService from '../../../auth/services/auth.service';
```
TamaraFinogina: this and all other path changes are because vitest complains otherwise. Ideally, all have to be changed, I only did as much as I could

src/app/auth/services/auth.service.ts
```diff
@@ -260,7 +283,7 @@ export const changePassword = async (newPassword: string, currentPassword: strin
   const usersClient = SdkFactory.getInstance().createUsersClient();
 
   return usersClient
-    .changePassword(<ChangePasswordPayload>{
+    .changePasswordLegacy(<ChangePasswordPayload>{
```
TamaraFinogina: due to some unrelated update of sdk

src/app/backups/hooks/useBackupsPagination.test.ts
```diff
@@ -10,23 +10,31 @@ import _ from 'lodash';
 import { FetchFolderContentResponse } from '@internxt/sdk/dist/drive/storage/types';
 import { DriveItemData } from '../../drive/types';
 import notificationsService, { ToastType } from '../../notifications/services/notifications.service';
+import { vi, describe, it, expect, beforeEach, Mock } from 'vitest';
 
-jest.mock('../../notifications/services/notifications.service', () => ({
-  show: jest.fn(),
+vi.mock('../../notifications/services/notifications.service', () => ({
```
TamaraFinogina: mock in vitest is different unfortunately

src/app/core/services/socket.service.ts
```diff
@@ -53,7 +53,7 @@ export default class RealtimeService {
     if (!this.socket) {
       throw new Error('Realtime service is not connected');
     }
-    return this.socket.id;
+    return this.socket.id ?? '';
```
TamaraFinogina: this is due to one of the dependencies upgrades

src/app/crypto/services/utils.ts
```diff
@@ -51,16 +145,6 @@ function decryptTextWithKey(encryptedText: string, keyToDecrypt: string): string
   return bytes.toString(CryptoJS.enc.Utf8);
 }
 
-function encryptFilename(filename: string, folderId: number): string {
```
TamaraFinogina: didn't seem to be used anywhere

src/app/network/crypto.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+describe('Test crypto.ts functions', () => {
+  globalThis.Buffer = Buffer;
+  it('encryptFilename should generate a ciphertext', async () => {
+    const mnemonic =
+      'until bonus summer risk chunk oyster census ability frown win pull steel measure employ rigid improve riot remind system earn inch broken chalk clip';
+    const bucketId = 'test busket id';
+    const filename = 'test filename';
+    const result = await encryptFilename(mnemonic, bucketId, filename);
+    expect(result).toBeDefined();
```
TamaraFinogina: dummy test, but helps coverage. There is no corresponding decrypt function as it's a legacy algorithm used in cross-checking different platforms (?)

src/app/network/crypto.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        ),
+      } as unknown as File;
+    }
+    const encryptionKey = crypto.randomBytes(32);
+    const iv = crypto.randomBytes(16);
+    const file = createMockFile('file.txt', 13, 'text/plain');
+    const cipher = crypto.createCipheriv('aes-256-ctr', encryptionKey, iv);
+    const [encryptedFile, hash] = await getEncryptedFile(file, cipher);
+    expect(encryptedFile).toBeDefined();
+    expect(hash).toBeDefined();
```
TamaraFinogina: if someone has real test values, I can add them here and below

src/app/network/crypto.ts
```diff
@@ -66,7 +65,10 @@ export async function encryptFilename(mnemonic: string, bucketId: string, filena
  * @param cipher Cipher used to encrypt the content
  * @returns A readable whose output is the encrypted content of the source stream
  */
-export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Cipher): ReadableStream<Uint8Array> {
+export function encryptReadable(
+  readable: ReadableStream<Uint8Array>,
+  cipher: crypto.Cipher,
```
TamaraFinogina: test complained on exporting Cipher from crypto, this is a workaround

src/app/network/crypto.ts
```diff
@@ -139,51 +145,40 @@ export async function getEncryptedFile(
...
 }
 
 export async function processEveryFileBlobReturnHash(
   chunkedFileReadable: ReadableStream<Uint8Array>,
   onEveryBlob: (blob: Blob) => Promise<void>,
 ): Promise<string> {
   const reader = chunkedFileReadable.getReader();
-  const hasher = new Sha256();
+  const hasher = await getSha256Hasher();
+  hasher.init();
```
TamaraFinogina: all get hasher function of hash-wasm require init()

src/app/network/requests.ts
```diff
@@ -364,7 +373,7 @@ export async function finishUpload(
     index: index.toString('hex'),
     hmac: {
       type: 'sha512',
-      value: generateHMAC([shardMeta], encryptionKey).toString('hex'),
+      value: await generateHMAC([shardMeta], encryptionKey),
```
TamaraFinogina: hash-wasm returns hex directly

src/app/payment/services/payment.service.ts
```diff
@@ -175,7 +175,7 @@ const paymentService = {
   async handlePaymentTeams(priceId: string, quantity: number, mode: StripeSessionMode): Promise<void> {
     const mnemonicTeam = generateMnemonic(256);
     const encMnemonicTeam = await encryptPGP(mnemonicTeam);
-    const codMnemonicTeam = Buffer.from(encMnemonicTeam).toString('base64');
+    const codMnemonicTeam = btoa(encMnemonicTeam as string);
```
TamaraFinogina: I hope it's the same. Update of dependencies ( types/node ?) caused this line to fail build

src/app/network/crypto.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const shardMeta = {
+      challenges_as_str: [],
+      hash: '',
+      index: 0,
+      parity: false,
+      size: 0,
+      tree: [],
+    };
+    const result = await generateHMAC([shardMeta], encryptionKey);
+    const expectedResult = await getHmacSha512(encryptionKey, ['']);
```
sg-gs: HMAC is deterministic, use a hardcoded correct output generated from an input instead of using another function to validate, we are creating a dependency of something that needs to be tested also

src/app/network/crypto.test.ts
sg-gs: Same here, calculate the hash and the output of the encryptedFile for a given input and match the result against a hardcoded output

src/app/network/crypto.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        const chunks = [new Uint8Array([1, 2, 3]), new Uint8Array([4, 5, 6]), new Uint8Array([7, 8, 9])];
+        for (const chunk of chunks) {
+          controller.enqueue(chunk);
+        }
+        controller.close();
+      },
+    });
+
+    const onEveryBlob = async (blob) => {
+      console.log('Received blob:', blob);
```
sg-gs: Remove this console.log

test/unit/services/utils.test.ts
sg-gs, line 1: Great tests @TamaraFinogina, this is the way to do crypto-related work 👍🏽

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const result = await extendSecret(message, 1048);
+    const testResult =
+      'bc3e3d41a1146b069abffad3c0d44860cf664390afce4d9661f7902e7943e085e01c59dab908c04c3342b816941a26d69c2605ebee5ec5291cc55e15b76146e6745f0601156c3596cb75065a9c57f35585a52e1ac70f69131c23d611ce11ee4ab1ec2c009012d236648e77be9295dd0426f29b764d65de58eb7d01dd42248204f45f8e';
+    expect(result).toBe(testResult);
+  });
+});
+
+describe('Test getPBKDF2 with RFC 7914 test vectors', () => {
+  it('getPBKDF2 should pass test 1 for PBKDF2-HMAC-SHA-256 from RFC 7914', async () => {
+    const passord = 'passwd';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 1;
+    const hashLength = 64;
+    const result = await getPBKDF2(passord, salt, iterations, hashLength);
+    const testResult =
+      '55ac046e56e3089fec1691c22544b605f94185216dde0465e68b9d57c20dacbc49ca9cccf179b645991664b39d77ef317c71b845b1e30bd509112041d3a19783';
+    expect(result).toBe(testResult);
+  });
+
+  it('getPBKDF2 should pass test 2 for PBKDF2-HMAC-SHA-256 from RFC 7914', async () => {
+    const passord = 'Password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    }
+    const result = await getRipemd160(message);
+    const testResult = '52783243c1697bdbe16d37f97f68f08325dc1528';
+    expect(result).toBe(testResult);
+  });
+});
+
+describe('Test getArgon2 with test vectors from the reference implementation that won Password Hashing Competition ', () => {
+  it('getArgon2 should pass test 1', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 2;
+    const memorySize = 65536;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '09316115d5cf24ed5a15a31a3ba326e5cf32edc24702987c02b6566f61913cf7';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 2', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 2;
+    const memorySize = 262144;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '78fe1ec91fb3aa5657d72e710854e4c3d9b9198c742f9616c2f085bed95b2e8c';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 3', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 2;
+    const memorySize = 256;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '9dfeb910e80bad0311fee20f9c0e2b12c17987b4cac90c2ef54d5b3021c68bfe';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 4', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 2;
+    const memorySize = 256;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '6d093c501fd5999645e0ea3bf620d7b8be7fd2db59c20d9fff9539da2bf57037';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 5', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 1;
+    const memorySize = 65536;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = 'f6a5adc1ba723dddef9b5ac1d464e180fcd9dffc9d1cbf76cca2fed795d9ca98';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 6', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 4;
+    const memorySize = 65536;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '9025d48e68ef7395cca9079da4c4ec3affb3c8911fe4f86d1a2520856f63172c';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 7', async () => {
+    const passord = 'differentpassword';
```
sg-gs: typo here, password instead of passord

test/unit/services/utils.test.ts
```diff
@@ -0,0 +1,669 @@
...
+    const iterations = 2;
+    const memorySize = 65536;
+    const hashLength = 32;
+    const result = await getArgon2(passord, salt, parallelism, iterations, memorySize, hashLength, 'hex');
+    const testResult = '0b84d652cf6b0c4beaef0dfe278ba6a80df6696281d7e0d2891b817d8c458fde';
+    expect(result).toBe(testResult);
+  });
+
+  it('getArgon2 should pass test 8', async () => {
+    const passord = 'password';
```
sg-gs: typo here, password instead of passord

sg-gs (changes requested):

src/app/payment/services/payment.service.ts
sg-gs: This is deprecated code, you can directly remove this

test/unit/services/utils.test.ts
TamaraFinogina: It's how RFC 7914 defines it.  Don't know why they spelled password like this. 
Screenshot 2024-11-15 at 14 46 27

test/unit/services/utils.test.ts
TamaraFinogina: This one should be fine I think 
image

test/unit/services/pgp.service.test.ts
TamaraFinogina: Done

.vscode/settings.json
TamaraFinogina: Done

src/app/payment/services/payment.service.ts
TamaraFinogina: @sg-gs  This function is used here https://github.com/internxt/drive-web/blob/8bf9912b4c550b98b6282786af6316e7705dd2e2/src/app/store/slices/payment/index.ts#L64 Should I remove try or all export as well?

TamaraFinogina: Done

src/app/network/crypto.test.ts
TamaraFinogina: Done

src/app/network/crypto.test.ts
TamaraFinogina: Done

src/app/network/crypto.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

package.json
```diff
@@ -16,30 +17,29 @@
...
+    "hash-wasm": "^4.11.0",
     "heic2any": "^0.0.4",
     "history": "^4.10.1",
     "https-browserify": "^1.0.0",
-    "i18next": "^22.4.9",
-    "i18next-browser-languagedetector": "^7.2.0",
-    "idb": "^6.1.5",
+    "i18next": "^23.16.4",
+    "i18next-browser-languagedetector": "^8.0.0",
+    "idb": "^8.0.0",
```
CandelR: have checked that this update makes all related work correctly?

src/app/crypto/services/utils.ts
```diff
@@ -1,24 +1,129 @@
...
+  } else if (passObject.salt.startsWith('argon2id$')) {
+    const argonSalt = passObject.salt.replace('argon2id$', '');
+    hash = await getArgon2(passObject.password, argonSalt);
+    salt = passObject.salt;
+  } else {
+    salt = passObject.salt;
+    const encoded = hex2oldEncoding(salt);
+    hash = await getPBKDF2(passObject.password, encoded);
+  }
+  return { salt, hash };
```
CandelR: It might be a good idea to add jsdoc to these functions

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
CandelR: Is the backend ready?

src/app/share/types/index.ts
```diff
@@ -10,6 +10,7 @@ export type AdvancedSharedItem = SharedFolders &
     sharingId?: string;
     sharingType: 'public' | 'private';
     encryptedPassword: string | null;
+    hybridModeEnabled: boolean | false;
```
CandelR: `boolean` already includes `false`, it is redundant

src/app/share/views/SharedLinksView/context/SharedViewContext.ts
```diff
@@ -32,6 +33,7 @@ export interface ShareViewState {
   };
   ownerBucket: null | string;
   ownerEncryptionKey: null | string;
+  ownerHybridModeEnabled: false | boolean;
```
CandelR: same here with `false` type

test/unit/services/keys.service.test.ts
```diff
@@ -1,19 +1,19 @@
...
 
 import { isValid } from '../../../src/app/crypto/services/utilspgp';
 import { generateNewKeys } from '../../../src/app/crypto/services/pgp.service';
-import { getAesInitFromEnv } from '../../../src/app/crypto/services/keys.service';
 
-import { config } from 'dotenv';
-config();
+import { describe, expect, it } from 'vitest';
+
+//config();
```
CandelR: remove commented line

test/unit/services/keys.service.test.ts
```diff
@@ -1,19 +1,19 @@
...
 
 //TODO: Disabled because validateFormat functions is not exists, review it and remove it or change them
 describe('# keys service tests', () => {
-  const aesInit = getAesInitFromEnv();
+  //const aesInit = getAesInitFromEnv();
```
CandelR: remove this commented line 😶‍🌫️

package.json
TamaraFinogina: How can I check it?

src/app/crypto/services/utils.ts
TamaraFinogina: Done

src/app/share/types/index.ts
TamaraFinogina: Done

src/app/share/views/SharedLinksView/context/SharedViewContext.ts
TamaraFinogina: Done

test/unit/services/keys.service.test.ts
TamaraFinogina: Done

test/unit/services/keys.service.test.ts
TamaraFinogina: Done

src/app/crypto/services/utils.ts
```diff
@@ -52,9 +82,24 @@ async function getHmacSha512(encryptionKey: Buffer, dataArray: string[] | Buffer
   return hmac.digest();
 }
 
-function extendSecret(message: Uint8Array, length: number): Promise<string> {
-  return blake3(message, length);
+/**
+ * Extands the given secret to the required number of bits
```
CandelR: nit: typo Extands -> Extends

CandelR (approved):

src/app/crypto/services/utils.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

test/unit/services/utils.test.ts
TamaraFinogina: Done

sg-gs (approved):

src/app/drive/components/ShareInviteDialog/ShareInviteDialog.tsx
TamaraFinogina: Still pending

CandelR (approved):

PR #1294, opened 2024-10-01
[PB-2978]: fix/disallow folder uploads in shared

xabg2:
A notification has been added indicating that folders cannot be uploaded to a shared folder.

PR #1296, opened 2024-10-03
[PB-2987]: fix/rename item issues in Shared Section

xabg2:
Correct behavior for Rename Item action in Shared section (for both personal and business workspace).

### DIFF
--- a/src/app/share/views/SharedLinksView/hooks/useSharedContextMenu.ts
+++ b/src/app/share/views/SharedLinksView/hooks/useSharedContextMenu.ts
@@ -73,7 +73,7 @@ const useSharedContextMenu = ({
 
     const getItemContextMenu = (item: AdvancedSharedItem) => {
       const userUUID = item?.user?.uuid;
-      const isNotRootFolderAndIsNotViewerUser = !isCurrentUserViewer && !isRootFolder;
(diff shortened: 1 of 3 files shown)


xabg2:
ℹ️  Merge [this PR](https://github.com/internxt/drive-web/pull/1294) first.

src/app/share/views/SharedLinksView/hooks/useSharedContextMenu.ts
```diff
@@ -71,6 +73,8 @@ const useSharedContextMenu = ({
 
     const getItemContextMenu = (item: AdvancedSharedItem) => {
       const userUUID = item?.user?.uuid;
+      const isNotRootFolderAndIsNotViewerUser = !isCurrentUserViewer && !isRootFolder;
+      const isRootFolderAndIsOwner = isRootFolder && isItemOwnedByCurrentUser(userUUID);
```
CandelR: Perhaps better name for the constant would be `isSubFolderAndEditorUser`, like the second one. 
Adding `not` to the names adds some complexity to the reading that may confuse the reader. 
Better to use statements whenever possible.

CandelR (approved):

PR #1300, opened 2024-10-03
[PB-1678]:Fix/Update wrong login credentials text

Jona-Internxt:
Update wrong login text and transaltions

### DIFF
--- a/src/app/auth/components/LogIn/LogIn.tsx
+++ b/src/app/auth/components/LogIn/LogIn.tsx
@@ -172,7 +172,11 @@ export default function LogIn(): JSX.Element {
         navigationService.history.push(`/activate/${email}`);
       }
 
-      setLoginError([castedError.message]);
+      setLoginError([
+        castedError.message.includes('Wrong login credentials')
+          ? translate('auth.login.wrongLogin')
+          : castedError.message,
+      ]);
       setShowErrors(true);
       if ((err as AppError)?.status === 403) {
         await sendUnblockAccountEmail(email);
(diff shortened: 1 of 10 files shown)


src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -172,7 +172,11 @@ export default function LogIn(): JSX.Element {
         navigationService.history.push(`/activate/${email}`);
       }
 
-      setLoginError([castedError.message]);
+      setLoginError([
+        castedError.message.includes('Wrong login credentials')
```
CandelR: Instead of comparing the error message to display that message, I think that will be bettter try doing it with the http status code. That way even if the backend change the message it will still work if the numerical code is the same.

Jona-Internxt: Changed!

src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -172,7 +172,7 @@ export default function LogIn(): JSX.Element {
         navigationService.history.push(`/activate/${email}`);
       }
 
-      setLoginError([castedError.message]);
+      setLoginError([(err as AppError)?.status === 401 ? translate('auth.login.wrongLogin') : castedError.message]);
```
xabg2: Maybe we can extract this conditional out of the setState. E.g.:
```
const UNAUTHORIZED_STATUS_CODE = 401;

function getLoginErrorMessage(err: unknown): string {
  const appError = err as AppError;
  if (appError?.status === UNAUTHORIZED_STATUS_CODE) {
    return translate('auth.login.wrongLogin');
  }
  return appError?.message || 'An unexpected error occurred';
}

setLoginError([getLoginErrorMessage(err)])
```

Jona-Internxt: Sounds good. Done!

CandelR (approved):

xabg2 (approved):

PR #1304, opened 2024-10-04
[PB-2439]:chore/[Technical debt] - Unify login and signup

Jona-Internxt:


src/app/auth/services/auth.service.ts
```diff
@@ -410,9 +420,101 @@ export const unblockAccount = (token: string): Promise<void> => {
   return authClient.unblockAccount(token);
 };
 
+export const signUp = async (
```
xabg2: Instead of passing multiple parameters, you can pass a single object that contains all the necessary properties:

```
type SignUpParams = {
  doSignUp: RegisterFunction | UpdateInfoFunction;
  email: string;
  password: string;
  token: string;
  isNewUser: boolean;
  redeemCodeObject: boolean;
  dispatch: AppDispatch;
};

export const signUp = async (params: SignUpParams) => {
  const { doSignUp, email, password, token, isNewUser, redeemCodeObject, dispatch } = params;
  // Function implementation
}
```

src/app/auth/services/auth.service.ts
```diff
@@ -410,9 +420,101 @@ export const unblockAccount = (token: string): Promise<void> => {
...
+  dispatch(userActions.setUser(user));
+  await dispatch(userThunks.initializeUserThunk());
+  dispatch(productsThunks.initializeThunk());
+
+  if (!redeemCodeObject) dispatch(planThunks.initializeThunk());
+  if (isNewUser) dispatch(referralsThunks.initializeThunk());
+  return { token: xToken, user: xUser, mnemonic };
+};
+
+export const logIn = async (
```
CandelR: Can you add tests to these functions?
They are quite sensitive to changes and it would be a good idea to have tests

src/app/auth/services/auth.service.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      token: mockToken,
+      user: mockUser,
+      mnemonic: mockMnemonic,
+    };
+
+    expect(authService.signUp).toHaveBeenCalledWith(expect.any(Object));
+    expect(result).toEqual(expectedResult);
+  });
+
+  it('should throw an error for invalidMethod authMethod', async () => {
```
CandelR: Maybe you could group the tests of the same function by `describe` blocks, so they would be grouped by function within the auth service tests

CandelR (commented):
Good work 🚀

src/app/auth/services/auth.service.test.ts
Jona-Internxt: Ok, done!

xabg2 (approved):
🚀

CandelR (approved):

PR #1317, opened 2024-10-21
Feat/PB-3080 generate pq keys for current users

TamaraFinogina:
Add support for both the current and hybrid encryption modes so the current users can migrate to hybrid settings.


patches/@internxt+sdk+1.5.17.patch
sg-gs, line 1: Why is this file here?

test/unit/services/pgp.service.test.ts
```diff
@@ -119,3 +121,52 @@ describe('Encryption and Decryption', () => {
...
+    const encryptedMessage = await standardEncryptMessageWithPublicKey({
+      message,
+      publicKeyInBase64,
+    });
+
+    expect(encryptedMessage).toBeDefined();
+  });
+
+  it('should generate keys, encrypt and decrypt a message successfully', async () => {
+    // Step 1: Generate new keys
```
sg-gs: Remove the comments, the code should be self-descriptive. Unless the involved logic is too complex (which we strive to avoid) the comments are discouraged.

sg-gs (changes requested):

patches/@internxt+sdk+1.5.17.patch
TamaraFinogina: @sg-gs I needed to update the sdk library:
- Add upgradeHash function
- Add flag hybridModeEnabled to the types ShareFolderWithUserPayload, WorkspaceUser, WorkspaceSetupInfo, and InviteMemberBody

Without the patch, drive-web won't build as SDK changes are not merged yet (PR is here: https://github.com/internxt/sdk/pull/251)

src/app/crypto/services/pgp.service.ts
```diff
@@ -57,9 +59,6 @@ export const encryptMessageWithPublicKey = async ({
   publicKeyInBase64: string;
   publicKyberKeyBase64: string;
 }): Promise<string> => {
-  if (message.length != 32) {
```
TamaraFinogina, line 60: @sg-gs I had to remove the 256-bit restriction on input messages. 

I was wrong to assume that a mnemonic is 256 bits. It's a string of much larger and varied lengths. So, I use hashing to expand the Kyber secret to whatever size we need for XORing. This is what the original Kyber paper does, but since NIST prefers a fixed-length output they removed it.

Kyber authors used sha3 but also suggested Blake as a faster alternative. I use Blake3 here only because the library allows any output length for blake3 and only a few options for sha3 output (224, 256, 384, and 512).

src/app/crypto/services/utils.ts
```diff
@@ -15,27 +16,33 @@ async function passToHash(passObject: PassObjectInterface): Promise<{ salt: stri
...
-  };
   if (passObject.salt) {
-    // salt is given, hence it's a double hash computed as argon2id(PBKDF2(pwd))
-    // Argon2id samples salt internally and stores it as part of output
+    // if salt =>   argon2id(PBKDF2(pwd))
     salt = CryptoJS.enc.Hex.parse(passObject.salt);
     const oldhash = CryptoJS.PBKDF2(passObject.password, salt, { keySize: 256 / 32, iterations: 10000 });
-    hash = await argon2.hash(oldhash, params);
+    const argonSalt = crypto.randomBytes(16);
+    hash = await argon2id({
```
TamaraFinogina: @sg-gs The library with blake3 had argon2, so I switched to it. 

Parameters are from OWASP recomendations: m=19456 (19 MiB), t=2, p=1 (Do not use with Argon2i) https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id

test/unit/services/pgp.service.test.ts
TamaraFinogina: Done

src/app/drive/services/download.service/downloader.ts
sg-gs, line 1: Maybe it's only me, but these linting changes are driving me crazy a bit. Do we have the style enforced for everyone here @xabg2? I do not understand how many changes are going on here from @TamaraFinogina unless we have these files not linted for a long time or we have a mismatch of configurations here.

src/app/network/crypto.ts
```diff
@@ -32,16 +32,17 @@ export function generateHMAC(
   return hmac.digest();
 }
 
-function getDeterministicKey(key: string, data: string): Buffer {
+async function getDeterministicKey(key: string, data: string): Promise<string> {
```
sg-gs: Remove async and await, if you return a Promise, that's enough

src/app/network/crypto.ts
```diff
@@ -139,51 +140,40 @@ export async function getEncryptedFile(
     const status = await readable.read();
 
     if (!status.done) {
-      hasher.process(status.value);
+      hasher.update(status.value);
```
sg-gs, line 148: Did you see any improvements in performance @TamaraFinogina ? I assume the answer it's yes by the WASM usage, it would be nice to highlight that on the PR description.

sg-gs (changes requested):

src/app/drive/services/download.service/downloader.ts
xabg2: I think the config is not working well, we should check the config.

src/app/network/crypto.ts
TamaraFinogina: Done

src/app/network/crypto.ts
TamaraFinogina: How can I test it? It should be faster, but do you know how can I get the numbers?

src/app/network/crypto.ts
```diff
@@ -139,51 +139,40 @@ export async function getEncryptedFile(
...
-    new Blob(blobParts, { type: 'application/octet-stream' }),
-    createHash('ripemd160')
-      .update(Buffer.from(hasher.result as Uint8Array))
-      .digest('hex'),
-  ];
-}
-
-export function sha256(input: Buffer): Buffer {
-  return createHash('sha256').update(input).digest();
+  const ripemd160Hash = await ripemd160(hasher.digest());
```
TamaraFinogina: @sg-gs Why here we have `ripemd160` of `sha256`? I didn't change this part but I'm curious.

sg-gs: Inherited from storj

PR #1334, opened 2024-10-31
[PB-2981]: feat/[B2B – Drive] - Searching for items in B2B doesn't return any result, instead it shows the items in B2C. When clicking on a result, the app crashes.

Jona-Internxt:


src/app/core/components/Navbar/NavbarGlobalSearch.tsx
```diff
@@ -96,6 +96,13 @@ const Navbar = (props: NavbarProps) => {
     }
   }, [filters]);
 
+  useEffect(() => {
```
xabg2, line 99: nit: you can extract this to a function called resetGlobalSearch or something like that.

xabg2 (approved):

PR #1351, opened 2024-11-12
[_]: feat/notifying the user if something fails while accepting a workspace invitation

xabg2:
Display a message when the workspace invitation is invalid.


Jona-Internxt (approved):

src/app/core/components/Sidenav/PendingInvitationsDialog.tsx
```diff
@@ -44,8 +48,19 @@ const PendingInvitationsDialog = ({
...
+          text: translate('notificationMessages.invalidWorkspaceInvitationError'),
+          type: ToastType.Error,
+        });
+      }
       const error = errorService.castError(err);
       errorService.reportError(error);
+      notificationsService.show({
+        text: translate('notificationMessages.errorAcceptingWorkspaceInvitation'),
+        type: ToastType.Error,
+      });
```
CandelR: this will display two different error messages

CandelR (approved):

PR #1352, opened 2024-11-12
[PB-3292] fix/navigation back and forth

rafijv:
- Fix forward and backward navigation when switching between personal and business workspace

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,14 +37,13 @@ import { AppDispatch, RootState } from './app/store';
 import { sessionActions } from './app/store/slices/session';
 import { uiActions } from './app/store/slices/ui';
 import { initializeUserThunk } from './app/store/slices/user';
-import { workspacesActions, workspaceThunks } from './app/store/slices/workspaces/workspacesStore';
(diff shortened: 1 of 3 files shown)


xabg2 (approved):

src/App.tsx
```diff
@@ -92,6 +94,15 @@ const App = (props: AppProps): JSX.Element => {
     i18next.changeLanguage();
   }, []);
 
+  useEffect(() => {
+    if (!isWorkspaceIdParam) {
+      localStorageService.set(STORAGE_KEYS.B2B_WORKSPACE, 'null');
+      localStorageService.set(STORAGE_KEYS.WORKSPACE_CREDENTIALS, 'null');
+      dispatch(workspacesActions.setSelectedWorkspace(null));
+      dispatch(workspacesActions.setCredentials(null));
```
CandelR: How about extracting this to a function that is `resetB2BWorkspaceCredentials` or something similar, so that we know that this causes that remove b2b workspace selection and return to personal workspace. 
Maybe it's not the best name I've suggested, if you can think of something that fits better, great!

CandelR (approved):

Jona-Internxt (approved):

PR #1362, opened 2024-11-26
[PB-3398] Feat/switch to hash wasm

TamaraFinogina:
Remove deprecated asmcrypto.js, switch to hash-wasm and check everything works as before


CandelR (approved):

sg-gs (commented):
Regarding: 
- About the mention of NIST test vectors on SHA tests: could you reference where those values come from? By reference, I mean hyperlinks, so we keep that as documentation for the future. 
- Same with RIPEMD160 tests, where do the references come from?

It would be nice to have that in a Markdown file in the same folder as the utils.ts file (`src/app/crypto/services`) _briefly_ explaining how are we validating the implementation of those methods.

sg-gs:
I have re-inforced the testing of this PR by comparing files integrity against the master version: 
- [X] Uploading from this version and downloading with the master's version.
- [X] Uploading from the master's version and downloading with this branch's version.

Checked for files of different sizes (so we test also multipart uploads, for instance)

PR #1363, opened 2024-11-26
[PB-3401] switch to hmac from hash-wasm

TamaraFinogina:
Switch to a faster hmac from hash-wasm and make sure that all operations give the identical results


sg-gs (approved):

CandelR (changes requested):
Same comment as here https://github.com/internxt/drive-web/pull/1367#pullrequestreview-2468219052, point to a branch that has been taken from today's master

PR #1367, opened 2024-11-28
[PB-3079] Feat/switch to argon2

TamaraFinogina:
PR switches drive-web to Argon2

We now use hash = PBKDF2(pwd, salt). Unfortunately, the parameters are outdated, so we need to upgrade. The idea is to compute Argon2 for all new users and update the hash after login for old ones.

    If no salt is given (new user), then it samples argon_salt, computes hash = Argon2 ( pwd, argon_salt) and sends (hash, 'argon2id$argoin_salt) to the database
    if salt starts with argon2id$, then it's already migrated user and we compute hash = Argon2( pwd, argon_salt)
    Else, then it's the user that hasn't migrated yet. We compute hash = PBKDF2(pwd, salt)

NOTE: Argon2 follows password [hash competition output format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md): the output will be $argon2id$v=19$m=65536,t=2,p=1$gZiV/M1gPc22ElAH/Jh1Hw$CWOrkoo7oJBQ/iyh7uJ0LO2aLEfrHwTWllSAxT0zRno or the same line but in HEX.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "docx-preview": "^0.3.3",
     "fflate": "^0.8.2",
     "file-saver": "^2.0.5",
+    "hash-wasm": "^4.11.0",
     "heic2any": "^0.0.4",
     "history": "^4.10.1",
     "https-browserify": "^1.0.0",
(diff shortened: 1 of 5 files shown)


CandelR (changes requested):
@TamaraFinogina take the branches out of the master branch and merge them to some intermediate branch that comes out of today's master, if we take them out or put the PR of vitest that broke the login we will be blocked again

PR #1371, opened 2024-11-28
[PB-3401]  switch to hmac from hash-wasm

TamaraFinogina:
Switch to a faster hmac from hash-wasm and make sure that all operations give the identical results


sg-gs (approved):

sg-gs (approved):
Run the same test as in https://github.com/internxt/drive-web/pull/1369#issuecomment-2506307019

Doc is present, tests are robust. LGTM

PR #1372, opened 2024-11-28
[PB-3079] switch to argon2

TamaraFinogina:
PR switches drive-web to Argon2

We now use hash = PBKDF2(pwd, salt). Unfortunately, the parameters are outdated, so we need to upgrade. The idea is to compute Argon2 for all new users and update the hash after login for old ones.

 - If no salt is given (new user), then it samples argon_salt, computes `hash = Argon2 (pwd, argon_salt)` and sends `(hash, 'argon2id$argoin_salt)` to the database
- If salt starts with argon2id$, then it's already migrated user and we compute `hash = Argon2( pwd, argon_salt)`
- Else, then it's the user that hasn't migrated yet. We compute `hash = PBKDF2(pwd, salt)`

NOTE: Argon2 follows password [hash competition output format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md): the output will be $argon2id$v=19$m=65536,t=2,p=1$gZiV/M1gPc22ElAH/Jh1Hw$CWOrkoo7oJBQ/iyh7uJ0LO2aLEfrHwTWllSAxT0zRno or the same line but in HEX.


src/app/crypto/services/utils.ts
```diff
@@ -2,8 +2,40 @@ import CryptoJS from 'crypto-js';
...
+  sha256,
+  sha512,
+  ripemd160,
+  createSHA1,
+} from 'hash-wasm';
 import { Buffer } from 'buffer';
+import crypto from 'crypto';
+
+/**
+ * Argon2id parameters taken from RFC9106 (variant for memory-constrained environments)
```
TamaraFinogina, line 20: @sg-gs. Specs for Argon2  (RFC [9106)](https://www.rfc-editor.org/rfc/rfc9106) offer 2 options:

1. Argon2id with t=1 iteration, p=4 lanes, m=2^(21) (2 GiB of RAM), 128-bit salt, and 256-bit tag size. 
2. Argon2id with t=3 iterations, p=4 lanes, m=2^(16) (64 MiB of RAM), 128-bit salt, and 256-bit tag size.

I went with the second one, cause 2Gib of RAM seems harsh for a client-side app

test/unit/services/utils.passToHash.test.ts
```diff
@@ -0,0 +1,297 @@
...
+    const hashLength = 20;
+    const result = await getPBKDF2(password, salt, iterations, hashLength);
+    const testResult = '4b007901b765489abead49d926f721d065a429c1';
+    expect(result).toBe(testResult);
+  });
+
+  it('getPBKDF2 should pass test 4 for PBKDF2-HMAC-SHA-1 from RFC 6070', async () => {
+    const password = 'password';
+    const salt = 'salt';
+    const iterations = 16777216;
```
TamaraFinogina, line 44: this test is veeery slow, we can remove if it's a problem

src/app/auth/services/auth.service.ts
```diff
@@ -190,6 +198,18 @@ export const doLogin = async (
       localStorageService.set('xMnemonic', clearMnemonic);
       localStorageService.set('xNewToken', newToken);
 
+      const newHash = await passToHash({ password });
+      const encryptedHashedNewPassword = encryptText(newHash.hash);
+      const encryptedHashedNewPasswordSalt = encryptText(newHash.salt);
+      const encryptedMnemonic = user.mnemonic;
+
+      await authClient.changePasswordWithLink(
+        token,
```
TamaraFinogina: should be token or newToken here?

CandelR: None of them, it is a backend generated token to make the password change possible, this solution will not work for what you need

CandelR: You can use the `changePassword` function, but you will have to do it only the first time this change is made, so you don't have to call this endpoint every time you login.

CandelR (changes requested):

src/app/auth/services/auth.service.ts
TamaraFinogina: ok, I'll change it

TamaraFinogina: @CandelR Done

src/app/auth/services/auth.service.ts
```diff
@@ -190,6 +198,12 @@ export const doLogin = async (
...
 
+      const salt = await getSalt(user.email);
+
+      if (!salt.startsWith('argon2id$')) {
+        await changePassword(password, password, user.email);
+      }
+
       return {
         user: clearUser,
         token: token,
```
TamaraFinogina, line 209: @CandelR This token would be wrong cause it was updated inside, no?

CandelR (changes requested):
After hash change (password change with new hash) is not possible to access to Setting security section
Screenshot 2024-12-05 at 08 48 35

CandelR (approved):

CandelR:
passed to draft so that it does not merge unintentionally

CandelR (commented):
@larry-internxt Will this also affect [cli](https://github.com/internxt/cli) login?

larryrider:
> @larry-internxt Will this also affect [cli](https://github.com/internxt/cli) login?

@CandelR If the server will still accept the old hash, this change will not have any impact on the CLI login. However, it might be a good idea to reuse this code for the CLI as well. We could even consider integrating this into inxt-js or the inxt-sdk, whichever makes the most sense, so all the apps (including desktop) could use it.

CandelR:
> > @larry-internxt Will this also affect [cli](https://github.com/internxt/cli) login?
> 
> @CandelR If the server will still accept the old hash, this change will not have any impact on the CLI login. However, it might be a good idea to reuse this code for the CLI as well. We could even consider integrating this into inxt-js or the inxt-sdk, whichever makes the most sense, so all the apps (including desktop) could use it.

The backend does not accept both, it overwrites the new one, so it would be good if we could add this implementation to some internxt library to be able to share it on all platforms, since it also applies to mobile. Desktop does the login through the web, so there should be no problem, but if the suite of apps is going to grow we will need to have the code as accessible as possible to all platforms.
On top of that it will also be easier to maintain as there will only be one project where the code is located and the others will use it.

TamaraFinogina:
Switched to backlog: https://inxt.atlassian.net/browse/PB-3079

PR #1373, opened 2024-11-28
[PB-3080] Add post-quantum (Kyber) 

TamaraFinogina:
**Add a post-quantum layer.**

Quantum computers will only be a problem for asymmetric encryption, which we use when we send encrypted mnemonics (ecc from OpenPGP in our case). The idea is to use hybrid encryption - ecc and KyberKEM - for sending the mnemonic. Thus, if ecc is broken or Kyber needs different parameters for security, the mnemonic is still safe.

For that, each user should additionally have a Kyber keys pair `(publicKyberKey, privateKyberKey)`. Then, we split mnenomic into two parts (`mnemonic = part1 XOR part2`) and encrypt `par1` with ecc and `part2` with Kyber. 

Encrypt:

```
(ciphertext, secret) await kem.encapsulate(publicKyberKey)
longSecret = await extendSecret(secret, len(mnemonic))
message = mnemonic XOR longSecret
eccCiphertext =  await openpgp.encrypt( message, publicKey)
send `(eccCiphertext, ciphertext)
```

Decrypt: 

```
secret  = await kem.decapsulate(ciphertext, privateKyberKey)
message  =  await openpgp.decrypt(eccCiphertext, privateKey)
longSecret = await extendSecret(secret, len(message))
mnemonic = message XOR longSecret
```

**NOTE 1:**  KyberKEM always outputs a fixed-length secret, but mnemonic can have different sizes. Hence, we extend it with the `extendSecret` function (that is just Blake3 hash).  It can also shrink the secret if needed, but usually mnemonic is bigger than the secret.

**NOTE 2:** During the migration period, we must support both old and new sharing methods. I've added a flag `hybridModeEnabled` to differentiate. But we can remove it and use ciphertext size or a prefix instead.


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -332,9 +332,11 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
 
     if (itemToShare?.item.uuid) {
       const encryptionKey = isAdvancedShareItem(itemToShare.item) ? itemToShare?.item?.encryptionKey : undefined;
+      const hybridModeEnabled = isAdvancedShareItem(itemToShare.item) ? itemToShare?.item?.hybridModeEnabled : false;
```
CandelR, line 335: Is the `hybridModeEnabled` field already added to the backend?

src/app/auth/services/auth.service.ts
```diff
@@ -106,13 +106,16 @@ export const is2FANeeded = async (email: string): Promise<boolean> => {
...
+  const { privateKeyArmored, publicKeyArmored, revocationCertificate, publicKyberKeyBase64, privateKyberKeyBase64 } =
+    await generateNewKeys();
 
   return {
     privateKeyArmored,
     privateKeyArmoredEncrypted: aes.encrypt(privateKeyArmored, password, getAesInitFromEnv()),
     publicKeyArmored,
     revocationCertificate,
+    publicKyberKeyBase64,
+    encPrivateKyberKey: aes.encrypt(privateKyberKeyBase64, password, getAesInitFromEnv()),
```
CandelR, line 118: the `enc` of `encPrivateKyberKey` means encrypted? 
you could add the whole word to make the const name more descriptive

CandelR (changes requested):

src/app/drive/components/ShareDialog/ShareDialog.tsx
TamaraFinogina: No, Kyber PR is blocked by the backend. However, those fields are in SDK already

TamaraFinogina:
there is another PR for kyber

PR #1375, opened 2024-12-02
[PB-3409] Remove crypto js 

TamaraFinogina:
## Description

Switch AES encryption/decryption from crypto-js to crypto and ensure that new functions can decrypt old ciphertexts

## Related Issues

Fixes [PB-3409](https://inxt.atlassian.net/browse/PB-3409)

## How Has This Been Tested?

Unit tests to check that old encryption can be decrypted by the new function and that old decrypt can handle new ciphertexts. 

## Additional Notes

Crypto-js is much older than I thought. It relies on MD5 and uses an old OpenSSL method for creating the key (https://docs.openssl.org/1.0.2/man3/EVP_BytesToKey/). By default, crypto-js performs only a single iteration of MD5 over the password, which is extremely weak unless the password is long and truly random. We need to migrate to something stronger. 

P.S. I've prepared a branch with stronger encryption (uses aes-256-gcm and argon2 for key derivation) that also supports old decryption as well for compatibility.  Once this PR or argon2 PR is merged, I'll create a PR for the encryption upgrade.

[PB-3409]: https://inxt.atlassian.net/browse/PB-3409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/crypto/services/utils.ts
+++ b/src/app/crypto/services/utils.ts
@@ -4,65 +4,7 @@ import { aes, items as itemUtils } from '@internxt/lib';
 import { AdvancedSharedItem } from '../../share/types';
 import { createSHA512, createHMAC, sha256, createSHA256, sha512, ripemd160 } from 'hash-wasm';
 import { Buffer } from 'buffer';
-
-/**
- * Converts HEX string to Uint8Array the same way CryptoJS did it (for compatibility)
- * @param {string} hex - The input string in HEX
- * @returns {Uint8Array} The resulting Uint8Array identical to what CryptoJS previously did
- */
-function hex2oldEncoding(hex: string): Uint8Array {
-  const words: number[] = [];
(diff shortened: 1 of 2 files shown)


CandelR (commented):
@TamaraFinogina Have you seen the comments on security?

TamaraFinogina:
@CandelR Yes. Unfortunately, this code (with md5 and 1 iteration) is what CryptoJS is doing now, it's pretty bad. I've prepared another branch (https://github.com/internxt/drive-web/tree/feat/upgrade-aes), which changes encryption to something better and can still decrypt old ciphertexts. I'm waiting for argon2 to be merged to avoid introducing argon2 in 2 different PRs

PR #1376, opened 2024-12-02
[PB-241]: fix/folder upload can't be cancelled

xabg2:
## Description
The folder used for the “cancel upload” operation is the root folder, so when the user cancels the upload, the remaining tasks will be removed and the root folder will be deleted with all uploaded items, so it will also disappear from Drive.

## Related Issues
Fixes [PB-241](https://inxt.atlassian.net/browse/PB-241).

## Checklist

## How Has This Been Tested?
- Upload 1 folder without levels (subfolders) and cancel it.
- Upload 1 folder with levels (nested folders) and cancel it.
- Upload several folders without levels and cancel them.
- Upload several folders with levels and cancel them.

[PB-241]: https://inxt.atlassian.net/browse/PB-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Additional notes
When the user is uploading a folder and then, the user cancels the upload, this folder remains uploaded on Drive instead of being deleted. 

It appears that the folder being deleted is the last uploaded folder, rather than the root folder (the parent folder and all levels/files), so the folder remains on Drive with the content that was uploaded (removing the last uploaded folder).

### DIFF
--- a/src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
+++ b/src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
@@ -17,15 +17,7 @@ import { getUniqueFolderName } from '../folderUtils/getUniqueFolderName';
 import { StorageState } from '../storage.model';
 import { deleteItemsThunk } from './deleteItemsThunk';
 import { uploadItemsParallelThunk } from './uploadItemsThunk';
-
-// TODO: REMOVE IROOT from this file, it is in types.ts
-export interface IRoot {
-  name: string;
-  folderId: string | null;
-  childrenFiles: File[];
-  childrenFolders: IRoot[];
-  fullPathEdited: string;
-}
+import { IRoot } from '../types';
 
 interface UploadFolderThunkPayload {
   root: IRoot;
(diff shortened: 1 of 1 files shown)


rafijv (approved):

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -153,7 +145,10 @@ export const uploadFolderThunk = createAsyncThunk<void, UploadFolderThunkPayload
         //Added wait in order to allow enough time for the server to create the folder
         await wait(500);
 
-        rootFolderItem = createdFolder;
+        if (!rootFolderItem) {
+          rootFolderItem = createdFolder;
```
CandelR, line 149: I don't see the point of this, every time a folder is created in the upload, the rootFolder is replaced. In production if it works to delete the folder when the upload is cancelled, it is possible that it does not identify that this change solves. Can you explain what I'm missing?

CandelR (changes requested):

CandelR (approved):

PR #1401, opened 2024-12-23
[PB-471]:fix/Generic messages

Jona-Internxt:
## Description

Various issues has been solved:

- Changed generic error messages for specific error messages.
- Removed duplicated error messages.
- Fixed issues related with removing items (refresh problems).
- Error messages for duplicated invitations.
- Scroll for modal when is not completely visible.
- Avoid recharge item in preview view if name is has been changed.

## Checklist

## How Has This Been Tested?

It has been tested locally following the workflows described in [the task from jira](https://inxt.atlassian.net/browse/PB-471).

## Additional Notes

The `isToastNotificacionOpen` has been removed. The useEffect from NoitificationToast was causing a recharg in the `ShareDialog`. If something wrong related with notification happens, review this part.

In `FileViewerWrapper` has been removed some `setBlob(null)` to avoid recharge items when change name. Pay attention to that in the review in case you think this could cause any problems.


src/app/drive/components/DriveExplorer/DriveExplorerList/DriveExplorerList.tsx
```diff
@@ -495,10 +495,13 @@ const DriveExplorerList: React.FC<DriveExplorerListProps> = memo((props) => {
           selectedItems={props.selectedItems}
           keyboardShortcuts={['unselectAll', 'selectAll', 'multiselect']}
           keyBoardShortcutActions={{
-            onBackspaceKeyPressed: () =>
-              isSelectedSharedItems
-                ? props.onOpenStopSharingAndMoveToTrashDialog()
-                : moveItemsToTrash(props.selectedItems),
+            onBackspaceKeyPressed: () => {
+              if (props.selectedItems.length === 1) {
+                isSelectedSharedItems
```
CandelR: nit :Add the condition to a variable with descriptive name in order to give it more context

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
```diff
@@ -84,34 +84,35 @@ const FileViewerWrapper = ({
...
-        .catch((error) => {
-          if (error.name === 'AbortError') {
-            return;
-          }
-          console.error(error);
-          setBlob(null);
-          errorService.reportError(error);
-          setIsDownloadStarted(false);
-        });
+    if (currentFile?.plainName === currentFile.name) {
```
CandelR, line 87: Why this condition?

src/app/store/slices/sharedLinks/index.ts
```diff
@@ -97,8 +97,11 @@ const shareItemWithUser = createAsyncThunk<string | void, ShareFileWithUserPaylo
       if (castedError.message === 'unauthenticated') {
         return navigationService.push(AppView.Login);
       }
+      const userInvited = castedError.message.includes('User already has a role');
```
CandelR, line 100: is the code returned by backend a specific number or is it a 500? I say this in case instead of using `the` includes with the message we can use the numerical code of the status of the response

src/app/store/slices/storage/storage.thunks/createFolderThunk.ts
```diff
@@ -106,7 +106,9 @@ export const createFolderThunkExtraReducers = (builder: ActionReducerMapBuilder<
       if (requestOptions?.showErrors) {
         const errorMessage = action.error.message?.includes('already exists')
           ? t('error.folderAlreadyExists')
-          : t('error.creatingFolder');
+          : action.error.message?.includes('Invalid folder name')
```
CandelR: Same here

src/app/drive/components/FileViewer/FileViewerWrapper.tsx
Jona-Internxt: Because when user changes the name, this is different and is the criteria I found for not reloading the file if the user only changed the name.

src/app/store/slices/sharedLinks/index.ts
Jona-Internxt: It's returning a 400 status, but I don't think it's appropriate for this, because we can get this code for other reasons. Maybe it's a good idea to get a 409 from the backend so we can handle this better.

src/app/drive/components/DriveExplorer/DriveExplorerList/DriveExplorerList.tsx
Jona-Internxt: In fact this is a bug, because it should be >0 in case you select various items.

src/app/store/slices/storage/storage.thunks/createFolderThunk.ts
Jona-Internxt: Here is the same, we always get a 400, so we cannot separate messages.

CandelR (commented):
Check the sonar cloud new issues

Jona-Internxt:
> Check the sonar cloud new issues

I fixed one of them. The other is a TODO that has to be checked when Modal component is integrated.

rafijv (approved):

CandelR (approved):

PR #1407, opened 2025-01-07
[_]: fix/check if image is valid before creating thumbnails

xabg2:
## Description
Check if an image is valid (the image can be previewed) before creating an image thumbnail, because if the image is not valid, the upload will hang.

This gives a bad impression to the user because the file is uploaded but shows at 100% in the task logger all the time, as it keeps creating the thumbnail.

## Checklist


src/app/drive/services/thumbnail.service.ts
```diff
@@ -48,6 +63,7 @@ const getImageThumbnail = (file: File): Promise<ThumbnailGenerated['file']> => {
       ThumbnailConfig.Quality,
       0,
       (uri) => {
+        console.log('URI:', uri);
```
CandelR: Remember to remove the logs

CandelR (approved):

PR #1414, opened 2025-01-09
[BR-888]: feat/use the same date format through the app

xabg2:
## Description
Now, we use the same format for all dates in the app (D MMM, YYYY [at] HH:mm).

## Checklist

- [x] QA Passed


src/app/backups/components/BackupsAsFoldersList/BackupsAsFoldersList.tsx
```diff
@@ -88,7 +88,7 @@ export default function BackupsAsFoldersList({
               );
             },
             (item) => {
-              return <div>{dateService.format(item.createdAt, 'DD MMMM YYYY. HH:mm')}</div>;
+              return <div>{dateService.format(item.createdAt, `D MMM, YYYY [${translate('general.at')}] HH:mm`)}</div>;
```
CandelR: In order to not has hardcoded same format in different places, maybe we could add a function to return this format passing the `translate` function to make the translation work. This way, we could manage all the date format from one place and not have to search every place that is implemented

xabg2: Indeed. Working on that.

CandelR (approved):

PR #1420, opened 2025-01-14
[PB-3627]: feat/send-kyber-private-key-reencrypted-on-pwd-change-request

TamaraFinogina:
## Description

Re-encrypt and send Kyber key on password update, upgrade to SDK 1.9.3

## Related Issues

Relates to [PB-3627](https://inxt.atlassian.net/browse/PB-3627)

## Related Pull Requests

- [Branch PB-2666-implement-pqc](https://github.com/internxt/drive-web/tree/feature/PB-2666-implement-pqc)

## Checklist

- [x] Changes have been tested locally.
- [ ] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [x] SonarCloud issues have been reviewed and addressed.
- [ ] QA Passed

## Additional Notes

Backend changes are pending

[PB-3627]: https://inxt.atlassian.net/browse/PB-3627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1438, opened 2025-01-23
[PB-3627]: put checks to avoid errors if user.keys does not exist

TamaraFinogina:
## Description

This is just in case users who are already logged in and have a user without field `keys` in their local storage. 

## Checklist

- [ ] QA Passed

## How Has This Been Tested?

unit tests

## Additional Notes

Maybe it's not necessary for parsedUser

### DIFF
--- a/src/app/auth/components/SignUp/WorkspaceGuestSignUp.tsx
+++ b/src/app/auth/components/SignUp/WorkspaceGuestSignUp.tsx
@@ -132,23 +132,27 @@ function WorkspaceGuestSingUpView(): JSX.Element {
       localStorageService.set('xNewToken', xNewToken);
 
       const decryptedPrivateKey = decryptPrivateKey(xUser.privateKey, password);
-      const decryptedPrivateKyberKey = decryptPrivateKey(xUser.keys.kyber.privateKey, password);
-
       const privateKey = xUser.privateKey ? Buffer.from(decryptedPrivateKey).toString('base64') : undefined;
-      const privateKyberKey = xUser.keys.kyber.privateKey
-        ? Buffer.from(decryptedPrivateKyberKey).toString('base64')
-        : '';
+
+      let privateKyberKey = '';
+      if (xUser.keys.kyber.privateKey) {
+        const decryptedPrivateKyberKey = decryptPrivateKey(xUser.keys.kyber.privateKey, password);
+        privateKyberKey = Buffer.from(decryptedPrivateKyberKey).toString('base64');
+      }
+
+      const publicKey = xUser.keys.ecc.publicKey ?? xUser.publicKey;
+      const publicKyberKey = xUser.keys.kyber.publicKey ?? '';
 
       const user = {
         ...xUser,
         privateKey,
         keys: {
           ecc: {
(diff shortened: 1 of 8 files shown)


src/app/auth/components/SignUp/WorkspaceGuestSignUp.tsx
```diff
@@ -132,23 +132,27 @@ function WorkspaceGuestSingUpView(): JSX.Element {
...
-        : '';
+
+      let privateKyberKey = '';
+      if (xUser.keys?.kyber?.privateKey) {
+        const decryptedPrivateKyberKey = decryptPrivateKey(xUser.keys.kyber.privateKey, password);
+        privateKyberKey = Buffer.from(decryptedPrivateKyberKey).toString('base64');
+      }
+
+      const publicKey = xUser.keys?.ecc?.publicKey ?? xUser.publicKey;
+      const publicKyberKey = xUser.keys?.kyber?.publicKey ?? '';
```
CandelR: As this block of code is repeated in different files, extract it to a function, and reuse it in all the necessary places. 
In this way only changing a piece of code we could keep this behaviour and if we need to change it we could avoid possible failures, because this way we have to know or look for all the places where this behaviour is implemented and it is easy to forget some point

CandelR (changes requested):

src/app/auth/components/SignUp/WorkspaceGuestSignUp.tsx
TamaraFinogina: @CandelR You mean the entire get privateKyberKey, right?

TamaraFinogina: @CandelR  Done

CandelR (approved):

CandelR:
**Do not merge** until pass QA

sg-gs (approved):
Running the QA. These changes affect: 

Workspace guest signup:
- [x] It should not break the current implementation
- [x] When joining as a guest in the older version by being invited in the new version and then leaving, the content of that guest is readable by the owner
- [x] When joining as a guest in the new version by being invited in the older version and then leaving, the content of that guest is readable by the owner. 

Private sharing guest signup:
- [x] It should not break the current implementation
- [x] When joining as a guest in the older version by being invited in the new version, the shared content is readable by the invited user
- [x] When joining as a guest in the new version by being invited in the older version, the shared content is readable by the invited user

Signup: 
- [x] Signing up with the new version and logging afterwards with the old one, keeps the same keys.

Login
- [x] Logging on the new version with an existing account from the old one, keeps the same keys.

TamaraFinogina:
@sg-gs Password change is in another PR: https://github.com/internxt/drive-web/pull/1420

Here I just removed an unused input from 1 internal function (the only function that called it passed '' as the last parameter)

sg-gs:
Now I realize I mixed things. It is the signup, not the change password which is affected by the auth.service changes @TamaraFinogina

sg-gs:
QA success @TamaraFinogina. Merging @CandelR

PR #1446, opened 2025-01-29
[_] Chore/add test for workspace guest signup

TamaraFinogina:
## Description

Add tests for WorkspaceGuestSignUp.ts

## Checklist

- [ ] QA Passed

## How Has This Been Tested?

Test that signing up works with and without kyber (for old and new users)

## Additional Notes

This is the only way I could make .svg mock work


src/app/auth/components/SignUp/WorkspaceGuestSignUp.test.tsx
```diff
@@ -0,0 +1,459 @@
...
+import WorkspaceGuestSingUpView from './WorkspaceGuestSignUp';
+import { userActions } from 'app/store/slices/user';
+import * as keysService from 'app/crypto/services/keys.service';
+import { encryptTextWithKey } from 'app/crypto/services/utils';
+import { UserSettings } from '@internxt/sdk/dist/shared/types/userSettings';
+import { useSignUp } from 'app/auth/components/SignUp/useSignUp';
+import { Buffer } from 'buffer';
+import { generateMnemonic } from 'bip39';
+
+if (typeof globalThis.process === 'undefined') {
```
larryrider: I think it's a better solution to fix the 'process.env' error via the "setupTests.ts" file, instead of adding it here (and in every test)
You should get the same result by adding these lines to the setupTests.ts:
`import process from 'process';`
`globalThis.process = process;`

What do you think?

CandelR (approved):
Awesome tests @TamaraFinogina 🚀

PR #1450, opened 2025-02-03
[PB-4489] Remove unnessesary Base64 encoding of private key (pgp one) before storing it locally

TamaraFinogina:
## Description

When we decrypt pgp private key, we convert it TO Base64. 
Then to use it we convert it FROM Base64. 
This PR fixes that. 

**For compatibility `smartKeyDecode` function can handle both encoded and non-encoded keys.** 

## Checklist

- [ ] QA Passed

## How Has This Been Tested?

Unit tests

## Additional Notes

The conversion was done so local storage keeps the Base64 version. It doesn't seem helpful and only causes unnecessary back-forth conversions.


src/app/auth/services/auth.service.ts
```diff
@@ -165,10 +165,7 @@ export const doLogin = async (
 
       if (encryptedPrivateKey) {
         await assertPrivateKeyIsValid(encryptedPrivateKey, password);
-        await assertValidateKeys(
-          Buffer.from(privateKey, 'base64').toString(),
-          Buffer.from(publicKey, 'base64').toString(),
-        );
+        await assertValidateKeys(privateKey, Buffer.from(publicKey, 'base64').toString());
```
CandelR, line 173: why only change this behaviour for private key and not also for public one?

test/unit/services/keys.service.test.ts
```diff
@@ -38,38 +40,96 @@ describe('Generate keys', () => {
     expect(keys).toHaveProperty('kyber.privateKeyEncrypted');
     expect(keys).toHaveProperty('kyber.publicKey');
   });
-});
 
-describe('# keys service tests', () => {
-  it('Should not update private key if encryption & encoding is fine', async () => {
-    const keys = await generateNewKeys();
-    const plainPrivateKey = keys.privateKeyArmored;
+  it('Should not decrypt null, empry, underfined or short private key', async () => {
```
CandelR: typo:  `empry, underfined`-> `empty, undefined`

src/app/crypto/services/pgp.service.ts
```diff
@@ -163,6 +163,25 @@ export const encryptMessageWithPublicKey = async ({
...
+  try {
+    return await openpgp.readPrivateKey({ armoredKey: key });
+  } catch (error) {
+    console.warn('Direct key parsing failed, trying base64 decoding...', error);
+
+    try {
+      const decodedKey = Buffer.from(key, 'base64').toString();
+      return await openpgp.readPrivateKey({ armoredKey: decodedKey });
+    } catch (base64Error) {
+      console.error('Both direct and base64 decoding failed.', base64Error);
```
CandelR: the logs are not necessary

src/app/auth/services/auth.service.ts
TamaraFinogina: @CandelR public one is converted to Base64 after generation and sent like that to the backend: https://github.com/internxt/drive-web/blob/737e3568645a0dfd0cdb3b60bb27ad6624b120a0/src/app/crypto/services/pgp.service.ts#L30 So, unless we decode it before storing in local, it's in Base64.

CandelR (approved):

PR #1456, opened 2025-02-05
[PB-3717] bugfix/retry upload file

Jona-Internxt:
## Description

Retry upload request for files if something is wrong (maximum 3 attempts, waiting 1000 ms for each retry)

## Related Issues

Issue reported [here](https://inxt.atlassian.net/browse/PB-3717).

## Checklist

- [x] QA Passed

## How Has This Been Tested?

Tested locally, simulating uploading file is failing firsts attempts.

### DIFF
--- a/src/app/network/upload.test.ts
+++ b/src/app/network/upload.test.ts
@@ -1,0 +1,57 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { uploadFile } from './upload';
+import { ConnectionLostError } from './requests';
+
+const uploadMock = vi.fn();
+
+const uploadParams = {
+  filesize: 50000,
+  filecontent: new File(['content'], 'file.txt'),
+  creds: { user: 'user', pass: 'password' },
+  mnemonic: 'mnemonic123',
+  progressCallback: vi.fn(),
+};
+
+vi.mock('./NetworkFacade', () => ({
+  NetworkFacade: vi.fn().mockImplementation(() => ({
+    uploadMultipart: vi.fn().mockResolvedValue('multipart-upload-success'),
+    upload: uploadMock,
+  })),
+}));
+
+describe('uploadFile', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should successfully upload a file', async () => {
+    uploadMock.mockResolvedValue('upload-success');
+
+    const result = await uploadFile('bucket123', uploadParams);
+
+    expect(result).toBe('upload-success');
+  });
+
+  it('should retry upload on failure and succeed on second attempt', async () => {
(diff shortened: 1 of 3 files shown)


src/app/network/upload.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  it('should successfully upload a file', async () => {
+    uploadMock.mockResolvedValue('upload-success');
+
+    const result = await uploadFile('bucket123', uploadParams);
+
+    expect(result).toBe('upload-success');
+  });
+
+  it('should retry upload on failure and succeed on second attempt', async () => {
```
sg-gs, line 36: You have to assert that the function inside `uploadFile` that you expect to be called again is, in fact, called again the same times that you are expecting it to be called (in this case, two).

src/app/network/upload.ts
```diff
@@ -130,34 +128,47 @@ export async function uploadFile(bucketId: string, params: IUploadParams): Promi
...
+      } catch (err) {
+        if (connectionLost) {
+          throw new ConnectionLostError();
+        }
+
+        console.warn(`Attempt ${attempt} of ${MAX_TRIES} failed:`, err);
+        Sentry.captureException(err, { extra: (err as ErrorWithContext).context });
+
+        if (attempt === MAX_TRIES) {
+          throw err; // ❌ Last try failed
```
sg-gs: If you do this one: 

```
const lastTryFailed = attempt === MAX_TRIES

if (lastTryFailed) {
   throw err;
}
```

Then the code is self-explanatory and the comment is not needed, which indeed is the best way to prevent outdated comments, as comments do not fail if they are wrong, the code does.

sg-gs (changes requested):

CandelR (approved):

src/app/network/upload.ts
```diff
@@ -130,34 +128,49 @@ export async function uploadFile(bucketId: string, params: IUploadParams): Promi
...
+        }
+
+        await new Promise((res) => setTimeout(res, RETRY_DELAY));
       }
+    }
 
-      Sentry.captureException(err, { extra: err.context });
-
-      Sentry.captureException(err, { extra: err.context });
+    throw new Error('Unexpected error in retryUpload');
```
TamaraFinogina: Only this line is not covered by tests. Any way to test it?

Jona-Internxt: The line` throw new Error('Unexpected error in retryUpload')` ensures the function always returns a value. It's unreachable and can't be simulated in tests, but it’s necessary to comply with the Promise<string> return type.

sg-gs: If that is the case, use the for inside the try-catch, not outside. Changing the code to match the types suggests an issue on the design of the code flow.

sg-gs (changes requested):

TamaraFinogina (approved):

sg-gs (approved):

PR #1463, opened 2025-02-12
store email in lowercase

Jona-Internxt:
## Description

Store email in lowercase after changing.

## Checklist

- [x] QA Passed

## Additional Notes

I took advantage of this and changed a padding marging in DriveExplorer topbar.

### DIFF
--- a/src/app/newSettings/Sections/Account/Account/AccountSection.tsx
+++ b/src/app/newSettings/Sections/Account/Account/AccountSection.tsx
@@ -49,7 +49,8 @@ const AccountSection = ({ changeSection, onClosePreferences }: AccountSectionPro
   };
 
   const changeUserEmail = async (newEmail) => {
-    await userService.changeEmail(newEmail);
+    const emailLowerCase = newEmail.toLowerCase();
+    await userService.changeEmail(emailLowerCase);
     notificationsService.show({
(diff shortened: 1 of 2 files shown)


src/app/newSettings/Sections/Account/Account/AccountSection.tsx
```diff
@@ -49,7 +49,8 @@ const AccountSection = ({ changeSection, onClosePreferences }: AccountSectionPro
   };
 
   const changeUserEmail = async (newEmail) => {
-    await userService.changeEmail(newEmail);
+    const emailLowerCase = newEmail.toLowerCase();
+    await userService.changeEmail(emailLowerCase);
```
CandelR, line 53: This is something that the backend should control, not just the frontend. 
Do you know if this is being done?
If we do it on the frontend, have you checked that it is sent in lower case in all the necessary places? login, sign up, etc?

CandelR (approved):

src/app/newSettings/Sections/Account/Account/AccountSection.tsx
Jona-Internxt: I agree that this should be controled from the backend, but for now is doing in frontend when the user signs up, so we have to control as well when changes email.

PR #1478, opened 2025-02-26
[PB-3490]:Feat/handle upload

Jona-Internxt:
## Description

The system allows uploading of files within folders to continue even if an individual file fails. . The new interface for failed files allows to retry the upload process for each file.

## Checklist

- [ ] QA Passed


src/app/network/upload.ts
```diff
@@ -151,6 +151,10 @@ export async function uploadFile(bucketId: string, params: IUploadParams): Promi
           });
         }
 
+        // TODO: Remove this after QA
+        const forceFail = Math.random();
+        if (file.name.includes('test_fail_file') && forceFail > 0.25) throw new Error('test error');
```
CandelR: maybe better to force fail always that include that name in order to test with 100% of accuracy that is working, in other way the QA would be random

src/app/store/slices/storage/hooks/useReduxActions.ts
```diff
@@ -65,5 +65,17 @@ export const useReduxActions = () => {
     );
   };
 
-  return { downloadItemsAsZip, downloadItems, uploadFolder, uploadItem, uploadSharedItem };
+  const uploadRetryItem = (data: { uploadFile: File; parentFolderId: string; taskId: string; fileType: string }) => {
+    dispatch(
+      uploadItemsThunk({
+        files: [data.uploadFile],
+        parentFolderId: data.parentFolderId,
+        taskId: data.taskId,
```
CandelR, line 73: in this way, when click retry, the file appears individually in the task logger, is it?

src/app/store/slices/storage/storage.model.ts
```diff
@@ -3,6 +3,7 @@ import { AdvancedSharedItem, SharedNamePath } from 'app/share/types';
 import { OrderDirection, OrderSettings } from '../../../core/types';
 import { DriveItemData, FileViewMode, FolderPath, FolderPathDialog } from '../../../drive/types';
 import { IRoot } from './types';
+import { UploadManagerFileParams } from 'app/network/UploadManager';
```
CandelR: I think this import is not being used

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -488,6 +493,8 @@ export const uploadItemsThunkExtraReducers = (builder: ActionReducerMapBuilder<S
     .addCase(uploadItemsParallelThunk.fulfilled, () => undefined)
     .addCase(uploadItemsParallelThunk.rejected, (state, action) => {
       const requestOptions = Object.assign(DEFAULT_OPTIONS, action.meta.arg.options ?? {});
+      const taskId = action.meta.arg.taskId;
```
CandelR, line 496: why if it has `taskId` is from retryManager?

src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
```diff
@@ -79,12 +81,15 @@ const PendingBlock = (): JSX.Element => {
...
   return isHovered && isUploadTask ? (
-    <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
+    <>
+      {haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}
+      <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
+    </>
   ) : (
     <div className="flex h-8 w-8 items-center justify-center">
-      <Success width={20} height={20} />
+      {haveWarnings ? <WarningIcon width={20} height={20} /> : <Success width={20} height={20} />}
```
CandelR, line 92: The warning only is displayed when it finish the upload? not when it is in progress and the files are failing?

CandelR (changes requested):

src/app/network/upload.ts
Jona-Internxt: If we do that, we cannot test the retry of the file

src/app/store/slices/storage/hooks/useReduxActions.ts
Jona-Internxt: Yes, exactly.

src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
Jona-Internxt: After talking with Design, the warning will appear at the end, instead the blue check.

src/app/network/upload.ts
TamaraFinogina: Can it always fail the first time but work on the second (or whichever is the last retry attempt)?

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
Jona-Internxt: When we retry the task, we have to set status as failed if the task has been rejected again. So we check if taskId exits and if task is retrying, then set status to failed.

src/app/network/upload.ts
Jona-Internxt: I think it's better this way, because the second retry could also fail and it's something that needs to be tested in QA. The new interface should handle that. The best way to test this is to create a folder with about ten 'test_fail_file' files and check it.

CandelR (approved):

src/app/network/upload.ts
TamaraFinogina: Yes, but when it's random failure, we might not always test it. So it will be up to chance

CandelR (approved):

src/app/network/upload.ts
Jona-Internxt: Remember remove this before merge.

PR #1517, opened 2025-04-08
[PB-3870]: fix/use of plainName property for backups names

xabg2:
## Description

This PR updates the way user backup folder names are displayed.
Instead of decrypting the names on the client side, we now rely on the already decrypted names provided by the backend.

Previously, client-side decryption could fail in certain cases due to inconsistencies in the encryption keys used (e.g., some folders may have been encrypted with a different or outdated key). This led to runtime errors and incomplete rendering of user backups.

By using the decrypted names directly from the server, we ensure a more reliable and consistent user experience, while also reducing client-side complexity and potential decryption errors.

## Checklist

- [ ] QA Passed

## How Has This Been Tested?

Go to **Backups** > check that the devices name is displayed correctly (both List and Breadcrumbs components).

## Additional Notes
-

### DIFF
--- a/src/app/backups/components/DeviceList/DeviceList.tsx
+++ b/src/app/backups/components/DeviceList/DeviceList.tsx
@@ -13,9 +13,10 @@ import UilWindows from '@iconscout/react-unicons/icons/uil-windows';
 import UilDesktop from '@iconscout/react-unicons/icons/uil-desktop';
 import dateService from '../../../core/services/date.service';
 import sizeService from '../../../drive/services/size.service';
-import { DriveFolderData } from '@internxt/sdk/dist/drive/storage/types';
+
 import { skinSkeleton } from 'app/shared/Skeleton';
 import { List } from '@internxt/ui';
+import { DriveFolderData } from 'app/drive/types';
 
 interface Props {
   items: (Device | DriveFolderData)[];
@@ -91,7 +92,7 @@ const DeviceList = (props: Props): JSX.Element => {
                   </div>
                   <div className="grow cursor-default truncate pr-3">
                     <button className="z-10 shrink cursor-pointer truncate" onClick={() => onDeviceClicked(device)}>
-                      {device.name}
+                      {(device as DriveFolderData).plainName}
                     </button>
                   </div>
                 </div>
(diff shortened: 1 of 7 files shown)


src/app/backups/components/DeviceList/DeviceList.tsx
```diff
@@ -91,7 +92,7 @@ const DeviceList = (props: Props): JSX.Element => {
                   </div>
                   <div className="grow cursor-default truncate pr-3">
                     <button className="z-10 shrink cursor-pointer truncate" onClick={() => onDeviceClicked(device)}>
-                      {device.name}
+                      {(device as DriveFolderData).plainName}
```
CandelR, line 95: the `Device` type has same or similar fields as `DriveFolderData`?

src/app/backups/hooks/useBackupDeviceActions.ts
```diff
@@ -1,7 +1,6 @@
-import { DriveFolderData } from '@internxt/sdk/dist/drive/storage/types';
 import { Device } from 'app/backups/types';
 import { deleteBackupDeviceAsFolder } from 'app/drive/services/folder.service';
-import { DriveItemData, DriveFolderData as DriveWebFolderData } from 'app/drive/types';
+import { DriveFolderData, DriveItemData } from 'app/drive/types';
```
CandelR, line 3: have you tried to update the sdk types in order to update them and use it in all projects?

src/app/shared/components/Breadcrumbs/Containers/BreadcrumbsBackupsView.tsx
```diff
@@ -39,7 +39,7 @@ const BreadcrumbsBackupsView = ({ backupsAsFoldersPath, goToFolder, goToFolderRo
     if (currentDevice && 'mac' in currentDevice) {
       items.push({
         uuid: currentDevice.id.toString(),
-        label: currentDevice.name,
+        label: (currentDevice as unknown as DriveFolderData).plainName ?? currentDevice.name,
```
CandelR, line 42: This is risky, avoid to use this type of cast if is not in testing

src/app/backups/hooks/useBackupDeviceActions.ts
xabg2: No. In fact, we should do it in another PR as we need to shift many imports through the project.

PR #1521, opened 2025-04-11
[PB-3966] Feat: Continue downloading the folder even if a file fails

evillalba94:
## Description

Currently, when downloading, if for any reason a file fails to download, the entire download is lost. We avoid breaking the zip file and allow the download process to continue.

## Checklist

- [x] QA Passed

## How Has This Been Tested?

There is code that must be REMOVED after the respective QA is performed.

To cause errors in file uploads, include the following `internxt_test_file` in the file name. This can be at the beginning, at the end, or as desired.

Files containing this name will fail when downloaded together; however, if downloaded individually, they will work. This allows for testing the individual download through the `Failed to download` modal, which the user can open when they see a warning triangle in the tasks.


src/app/drive/services/downloadManager.service.ts
```diff
@@ -232,19 +242,29 @@ export class DownloadManagerService {
...
+        options: {
+          closeWhenFinished: true,
+          ...credentials,
+        },
+        abortController,
+      });
+
+      if (res?.allItemsFailed) {
+        throw new Error(ErrorMessages.ServerUnavailable);
+      }
```
CandelR: It doesn't make much sense to throw a server unavailable error if all files fail, this should be thrown by the server and managed in the front.
Is this due to something I'm missing?

src/app/drive/services/downloadManager.service.ts
```diff
@@ -300,73 +327,96 @@ export class DownloadManagerService {
...
-          mnemonic: (driveItem as AdvancedSharedItem).credentials?.mnemonic ?? credentials.mnemonic,
-          options: {
-            abortController,
-            notifyProgress: (totalBytes, downloadedBytes) => {
-              const progress = downloadedBytes / totalBytes;
+      // TODO: QA REMOVE
+      if (driveItem.name.includes('internxt_test_file') || driveItem.plainName?.includes('internxt_test_file')) {
+        failedItems.push(driveItem);
+        return;
+      }
```
CandelR: Perfect, I left a check to remember to remove before merge

- [x] Removed forced error for QA

src/app/drive/services/downloadManager.service.ts
```diff
@@ -300,73 +327,96 @@ export class DownloadManagerService {
...
-        },
-        updateNumItems: incrementItemCount,
-        options: {
-          destination: folderZip,
-          closeWhenFinished: false,
-          credentials: {
-            user: (driveItem as AdvancedSharedItem).credentials?.networkUser ?? credentials.credentials.user,
-            pass: (driveItem as AdvancedSharedItem).credentials?.networkPass ?? credentials.credentials.pass,
+      try {
+        const res = await downloadFolderAsZip({
```
CandelR: it would be good to give a more descriptive name to this const, `res` looks like a server response

src/app/drive/services/downloadManager.service.ts
```diff
@@ -381,12 +431,39 @@ export class DownloadManagerService {
...
+          throw error;
+        }
       }
     }
 
+    if (failedItems.length > 0) {
+      if (failedItems.length === items.length) {
+        const allEqual = failedItems.every(
+          (item, index) => item.id === items[index].id && item.isFolder === items[index].isFolder,
+        );
```
CandelR: failed items will be always in same order as original items list?

src/app/drive/services/downloadManager.service.ts
```diff
@@ -381,12 +431,39 @@ export class DownloadManagerService {
...
+          throw new Error(ErrorMessages.ServerUnavailable);
+        }
+      }
+      downloadTask.failedItems = failedItems;
+    }
+
     await folderZip.close();
   };
+
+  private isRequiredHandleError = (error: any) => {
```
CandelR: try to give `error` the correct type

src/app/drive/services/folder.service.ts
```diff
@@ -298,11 +298,18 @@ export async function downloadFolderAsZip({
   updateNumItems: () => void;
   options: DownloadFolderAsZipOptions;
   abortController?: AbortController;
-}): Promise<void> {
+}): Promise<{
+  totalItems: DriveFileData[] & SharedFiles[];
+  failedItems: DriveFileData[] & SharedFiles[];
+  allItemsFailed: boolean;
```
CandelR: Maybe adding this to types can help you to reuse it for the file, as I see that `DriveFileData[] & SharedFiles[]` are repeated.

src/app/drive/services/folder.service.ts
```diff
@@ -349,39 +356,62 @@ export async function downloadFolderAsZip({
...
-        const isCachedFileOlder = checkIfCachedSourceIsOlder({ cachedFile, file });
-
-        if (cachedFile?.source && !isCachedFileOlder) {
-          return cachedFile.source.stream();
+        addUniqueItem(file);
+        // TODO: QA REMOVE
+        if (file.name.includes('internxt_test_file') || file.plainName?.includes('internxt_test_file')) {
+          failedItems.push(file as DriveFileData);
+          return;
+        }
```
CandelR: I add the check ✅ 

- [x] Removed forced error for QA

src/app/drive/services/folder.service.ts
```diff
@@ -349,39 +356,62 @@ export async function downloadFolderAsZip({
...
+          await updateDatabaseFileSourceData({
+            folderId: file.folderId,
+            sourceBlob,
+            fileId: file.id,
+            updatedAt: file.updatedAt,
+          });
+
+          return sourceBlob.stream();
+        } catch (error: any) {
+          const serverUnavailableError = error.message === 'Server unavailable';
```
CandelR: by convention, the boolean variables is starts with prefixes as  `is`, `has`, `can`, `should`, etc.
Comparing the message in this way is very susceptible to not being correct with respect to changes, if you want to check the message I think it would be better to use an includes, as it provides a bit more flexibility, although this kind of comparison is not the best. If possible it is better to use the error codes.

src/app/drive/services/folder.service.ts
```diff
@@ -416,34 +446,50 @@ export async function downloadFolderAsZip({
...
-
-        if (cachedFile?.source && !isCachedFileOlder) {
-          updateProgress(1);
-          return cachedFile.source.stream();
+        addUniqueItem(file);
+        // TODO: QA REMOVE
+        if (file.name.includes('internxt_test_file') || file.plainName?.includes('internxt_test_file')) {
+          failedItems.push(file);
+          return;
+        }
```
CandelR: I add the check ✅ 

- [x] Removed forced error for QA

src/app/drive/services/folder.service.ts
```diff
@@ -416,34 +446,50 @@ export async function downloadFolderAsZip({
...
+            updatedAt: file.updatedAt,
+          });
+
+          return sourceBlob.stream();
+        } catch (error: any) {
+          const serverUnavailableError = error.message === 'Server unavailable';
+          const isLostConnectionError = error instanceof ConnectionLostError || error.message === 'Network Error';
+          if (serverUnavailableError || isLostConnectionError) {
+            throw error;
+          }
```
CandelR: This check is in different places, extract it to avoid repeating the code and make it easier to maintain and apply the suggested changes

src/app/network/DownloadManager.ts
```diff
@@ -185,4 +211,13 @@ export class DownloadManager {
...
   };
+
+  private static compareArraysByIdAndFolder = (arr1: DownloadItemType[], arr2: DownloadItemType[]) => {
+    if (arr1.length !== arr2.length) return false;
+
+    return arr1.every((obj1, index) => {
+      const obj2 = arr2[index];
+      return obj2?.id === obj1.id && obj2?.isFolder === obj1.isFolder;
+    });
+  };
```
CandelR: This function has same problem as before, if it has the same items in different order will not work correctly.
Also add tests for this function

src/app/drive/services/downloadManager.service.test.ts
```diff
@@ -454,6 +459,7 @@ describe('downloadManagerService', () => {
         showErrors: true,
       },
       taskId: mockTaskId,
+      failedItems: [],
```
TamaraFinogina, line 475: Maybe create a constructor for the expectedTask? It has many repeated fields

src/app/drive/services/downloadManager.service.ts
CandelR: This remains the same. I don't think it's appropriate to throw a server error from the frontend, if the backend throws it for some reason it should be controlled and show a more user friendly message. 
Have you left it like that for some reason that I don't know?

src/app/network/download.ts
```diff
@@ -166,55 +166,103 @@ async function getRequiredFileMetadataWithAuth(
...
+    connectionLost = true;
+    window.removeEventListener('offline', connectionLostListener);
+  };
+
+  window.addEventListener('offline', connectionLostListener);
+  const timeoutId = setTimeout(() => {
+    if (!navigator.onLine) {
+      connectionLost = true;
+    }
+  }, timeoutMs);
```
CandelR: Better not to add more complexity to the download function, having the `downloadManager` you can add the detection there and this function will only take care of downloading the file

src/app/network/download.ts
```diff
@@ -166,55 +166,103 @@ async function getRequiredFileMetadataWithAuth(
...
+    connectionLost = true;
+    window.removeEventListener('offline', connectionLostListener);
+  };
+  window.addEventListener('offline', connectionLostListener);
+
+  const timeoutId = setTimeout(() => {
+    if (!navigator.onLine) {
+      connectionLost = true;
+    }
+  }, timeoutMs);
```
CandelR: Same here

src/app/drive/services/filesZip.service.ts
```diff
@@ -39,22 +41,30 @@ async function addFilesToZip<T extends File>(
...
+          const downloadPromise = downloadFile(file)?.then(async (fileStream) => {
+            if (!fileStream) {
+              return undefined;
+            }
             const fileBlob = await binaryStreamToBlob(fileStream, file.type || '');
             return { name: file.plainName ?? file.name, type: file.type, blob: fileBlob };
           });
 
-          downloadPromises.push(downloadPromise);
+          if (downloadPromise) {
```
TamaraFinogina: Sonar doesn't like it: "Expected non-Promise value in a boolean conditional."

src/app/drive/services/downloadManager.service.ts
evillalba94: When creating a zip file for download, we evaluate whether any file failed, and if so, we save it and then display it to the user. We no longer return the error directly when a file fails. Instead, we continue downloading the remaining files. Therefore, since a file skip is performed to continue the download, we now check if all files failed.

That's why we trigger this error, and then capture it in the `DownloadManager` and display a message.

There are several scenarios:
1- Disconnection failure.
2- One or Multiple files failure.
3- All files failure.

CandelR:
Hey @evillalba94, I'm leaving the PRs in draft until they are ready for review 👌

evillalba94:
@CandelR Ready for review. Although the coverage is failing, these tests were addressed in the next PR #1522. 

PR #1522 targets this branch `feat/pb-3966-continue-downloading`

src/app/drive/services/downloadManager.service.test.ts
```diff
@@ -580,6 +602,143 @@ describe('downloadManagerService', () => {
...
+    const checkAndHandleConnectionLossSpy = vi.spyOn(DownloadManagerService.instance, 'checkAndHandleConnectionLoss');
+
+    await DownloadManagerService.instance.downloadFolder(mockTask, mockUpdateProgress, mockIncrementItemCount);
+
+    expect(mockTask.failedItems).toEqual([mockFile, mockFile2]);
+    expect(handleConnectionLostSpy).toHaveBeenCalledWith(5000);
+    expect(checkAndHandleConnectionLossSpy).toHaveBeenCalled();
+  });
+
+  it('should downloadFolder throw if all files inside the folder fails', async () => {
```
CandelR: nit: forgot to specify what should throw

src/app/drive/services/downloadManager.service.test.ts
```diff
@@ -825,9 +997,640 @@ describe('downloadManagerService', () => {
...
+
+    describe('checkAndHandleConnectionLoss', () => {
+      it('should throw ConnectionLostError if connection is lost', async () => {
+        navigatorOnLineSpy.mockReturnValue(false);
+        await expect(DownloadManagerService.instance.checkAndHandleConnectionLoss(true)).rejects.toThrow(
+          ConnectionLostError,
+        );
+      });
+
+      it('should throw ConnectionLostError if conn is true and there is a zip value', async () => {
```
CandelR: I would say that part of the test description is missing if I understand correctly, what does it mean if conn is true? 
if the connection is lost?

CandelR (commented):
Hey @evillalba94, ready for QA. I'll leave it without the approve in order to we not forget remove the hardcoded conditions for QA before merge 👌 
Also, try to address the sonar cloud new issues to reduce them

src/app/tasks/components/TaskToRetryItem/TaskToRetryItem.test.tsx
```diff
@@ -58,8 +58,8 @@ describe('TaskToRetyItem', () => {
     expect(mockDownloadItem).toHaveBeenCalledWith(mockFile);
   });
 
-  it('should render a spinner when the file status is "uploading"', () => {
-    const uploadingFile = { ...mockFile, status: 'uploading' };
```
TamaraFinogina, line 62: What happens if the file status is uploading? Should it also render a spinner?

src/app/tasks/components/TaskToRetry/TaskToRetry.tsx
```diff
@@ -3,39 +3,63 @@ import { useReduxActions } from 'app/store/slices/storage/hooks/useReduxActions'
...
-      RetryManager.changeStatus(fileParams.params.taskId ?? '', 'failed');
+  const downloadItem = async ({ taskId, params, type }: RetryableTask) => {
+    if (type == 'upload') {
+      const data = {
+        uploadFile: params?.filecontent?.content,
+        parentFolderId: params?.parentFolderId,
+        taskId: taskId ?? params?.taskId ?? '',
+        fileType: params?.filecontent?.type ?? '',
+      };
+      RetryManager.changeStatus(taskId ?? params?.taskId ?? '', 'retrying');
```
TamaraFinogina, line 32: Uploading and retrying the download have the same status, right?

src/app/tasks/components/TaskToRetry/TaskToRetry.test.tsx
```diff
@@ -24,8 +43,10 @@ describe('TaskToRetry', () => {
   const mockTranslate = vi.fn();
   const mockChangeStatus = vi.spyOn(RetryManager, 'changeStatus');
 
-  const files: FileToRetry[] = [
+  const files: RetryableTask[] = [
     {
+      taskId: 'task-1',
```
TamaraFinogina, line 48: taskId should be in RetriableTask directly or in params?

src/app/tasks/components/TaskToRetry/TaskToRetry.test.tsx
```diff
@@ -3,19 +3,38 @@ import { describe, it, vi, expect, beforeEach } from 'vitest';
...
+  useTranslationContext: vi.fn(),
 }));
 
 vi.mock('app/store/slices/storage/hooks/useReduxActions', () => ({
   useReduxActions: vi.fn(),
 }));
 
-vi.mock('app/i18n/provider/TranslationProvider', () => ({
-  useTranslationContext: vi.fn(),
+vi.mock('@internxt/ui', () => ({
```
TamaraFinogina, line 22: Nice!!

src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
```diff
@@ -81,17 +81,23 @@ const PendingBlock = (): JSX.Element => {
...
+  if (isHovered && taskType.includes('upload')) {
+    return (
+      <>
+        {haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}
+        <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
+      </>
+    );
+  } else if (isHovered && taskType.includes('download')) {
+    return <>{haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}</>;
+  } else {
```
TamaraFinogina, line 94: Why only upload needs  `<TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />`? Because maybe both cases can be combined

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.test.ts
```diff
@@ -182,10 +182,14 @@ describe('uploadItemsThunkExtraReducers', () => {
...
 
     const rejectedHandler = cases.get(uploadItemsParallelThunk.rejected);
-    const RetryIsRetryingFileSpy = vi.spyOn(RetryManager, 'isRetryingFile');
+    const RetryIsRetryingFileSpy = vi.spyOn(RetryManager, 'isRetryingTask');
     const RetryChangeStatusSpy = vi.spyOn(RetryManager, 'changeStatus');
 
-    RetryManager.addFile(sampleFile);
+    RetryManager.addTask({
+      type: 'upload',
+      taskId: sampleFile.taskId || 'task1',
```
TamaraFinogina, line 190: why you need || 'task1'? It's the test and the file you mock

src/app/drive/services/filesZip.service.ts
```diff
@@ -39,22 +41,30 @@ async function addFilesToZip<T extends File>(
...
+              type: string;
+              blob: Blob;
+            }
+          | undefined
+        >[] = [];
 
         for (const file of filesChunk) {
-          const downloadPromise = downloadFile(file).then(async (fileStream) => {
+          const downloadPromise = downloadFile(file)?.then(async (fileStream) => {
+            if (!fileStream) {
```
TamaraFinogina, line 55: And if in this case, use `new ReadableStream()` - an empty stream, or reject the promise? I'm a bit concerned that all the code that used to receive a promise of an object won't be ready to deal with undefined and won't check for it.

src/app/drive/services/folder.service.ts
```diff
@@ -349,39 +357,60 @@ export async function downloadFolderAsZip({
...
       async (file) => {
-        updateNumItems();
-        const lruFilesCacheManager = await LRUFilesCacheManager.getInstance();
-        const cachedFile = await lruFilesCacheManager.get(file.uuid?.toString());
-        const isCachedFileOlder = checkIfCachedSourceIsOlder({ cachedFile, file });
-
-        if (cachedFile?.source && !isCachedFileOlder) {
-          return cachedFile.source.stream();
+        addUniqueItem(file);
+        // TODO: QA REMOVE
```
TamaraFinogina, line 372: @CandelR  another TODO

src/app/network/DownloadManager.test.ts
```diff
@@ -686,4 +816,334 @@ describe('downloadManager', () => {
...
+      } as DownloadTask;
+
+      const removeTaskSpy = vi.spyOn(retryManager, 'removeTaskByIdAndParams');
+
+      DownloadManager['removeRetryItems'](items, downloadTask);
+
+      expect(removeTaskSpy).not.toHaveBeenCalled();
+    });
+
+    it('should handle an empty failed items list', () => {
```
TamaraFinogina, line 929: What about an empty item list but non-empty failedItems?

src/app/network/DownloadManager.test.ts
```diff
@@ -686,4 +816,334 @@ describe('downloadManager', () => {
...
+        taskId: 'task-2',
+        merge: { status: TaskStatus.Success },
+      });
+      expect(downloadItemsSpy).toHaveBeenCalled();
+    });
+  });
+
+  describe('removeRetryItems', () => {
+    it('should remove retry items that are not in the failed items list', () => {
+      const items = [
```
TamaraFinogina, line 888: Is there a test for failedItems containing elements that are not in items? I cannot find it

src/app/tasks/components/TaskToRetry/TaskToRetry.tsx
evillalba94: @TamaraFinogina Yes, instead of using `uploading` or `downloading` we are using `retrying`. At this point in the process the user has previously attempted to upload and download, so now it is time to retry the process.

src/app/tasks/components/TaskToRetry/TaskToRetry.test.tsx
evillalba94: Previously, the `taskId` was obtained through `params`; now, with the modification, it can be obtained directly from `RetryableTask`.

src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
evillalba94: Because Downloads don't perform a search or redirection to a folder as such. Therefore, they weren't combined.

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.test.ts
evillalba94: The var `sampleFile` is of type `UploadManagerFileParams`, and within that type the `taskId` is `optional`. That's why I added the value `|| 'taskId'` since it cannot be `undefined`.

So I had the option to remove `sampleFile.taskId` or leave it, and I decided to leave it.

src/app/drive/services/filesZip.service.ts
evillalba94: image

Yes, of course, as you can see, this is the filter applied to that promise. That way, we only continue with the added and validated objects.

CandelR (approved):

PR #1528, opened 2025-05-07
PoC openapi fetch

dajimenezriv-internxt:
## Description

This is proof of concept to check if it's worth the use of openapi-typescript and openapi-fetch.

`openapi-typescript` is a library that pointing to an openapi schema can generate a file containing all types of that api.
`openapi-fetch` is a library that using a generated schema.d.ts from `openapi-typescript` can ensure end-to-end type-safety between a server and a client.

This PoC tests this using the endpoint of getWorkspaces. Currently, we implement an endpoint in drive-server-wip, then we implement that endpoint in sdk and the types, and then the clients update the version of the sdk and can use the new endpoint with the new types. However, this involves the manual creation of types on the sdk and it's easy to make mistakes (indeed, in the endpoint of getWorkspaces types are not right). We can see that the sdk says that workspaceUser returns the fields `freeSpace`, `isManager`, `isOwner` and `usedSpace`, but it doesn't.

The new solution will involved to start declaring the types right in drive-server-wip, which is easier since we are using typescript, and you cannot say that you are returning something that you are not, and after that, all clients will execute the generate schema script:

```bash
openapi-typescript http://api.internxt.com/api-json -o ./<path-to-schema>.d.ts
```

After that the client will be able to use the endpoint with the new types.

## Checklist

- [ ] QA Passed


CandelR:
I close it to remove PRs that will not be merged in the short term. I keep the branch to see these changes at a later time

PR #1538, opened 2025-05-22
[PB-4382]: feat/shorten-public-share-links

larryrider:
## Description

It extends the existing file-sharing flow to produce a shortened, zero-knowledge link that is at least 50% shorter than the current URLs. The encryption key has been reduced and the sharing id has been encoded to reduce its length

## Checklist

- [ ] QA Passed

## Testing Process

- Obtain the generated shortened link from your desired test item.
- Verify Link Length Reduction by comparing the length of the new shortened link to the original full-length link
- Access and Download the File. Ensure the item downloads successfully without errors.
- Validate File Integrity by computing the checksum (e.g., SHA-256) of the downloaded file vs the original file. Confirm that the checksums match, ensuring successful decryption and integrity.


src/app/share/views/ShareView/ShareFileView.tsx
```diff
@@ -55,8 +57,13 @@ interface ShareViewState {
 
 export default function ShareFileView(props: ShareViewProps): JSX.Element {
   const { translate } = useTranslationContext();
-  const sharingId = props.match.params.token;
+
   const code = props.match.params.code;
+  let sharingId = props.match.params.token;
+  if (!validateUuidv4(sharingId)) {
+    sharingId = base64UrlSafetoUUID(sharingId);
+  }
```
CandelR: what will happen with ShareFolderView? Doesn't the same logic apply? If so, perhaps this should be extracted to keep it in one place and apply to all that is needed 🤔

src/app/share/services/share.service.ts
```diff
@@ -371,12 +370,22 @@ export const getPublicShareLink = async (
...
-    const publicShareLink = `${selectedDomain}/d/sh/${itemType}/${sharingId}/${plainCode}`;
+    if (selectedDomain) {
+      selectedDomain = `${selectedDomain}/d`;
+    } else {
+      selectedDomain = window.location.origin;
+    }
+
+    const removedUuidDecoration = sharingId.replace(/-/g, '');
+    const base64endoded = Buffer.from(removedUuidDecoration, 'hex').toString('base64');
+    const encodedSharingId = toBase64UrlSafe(base64endoded);
```
CandelR: What do you think if we extract this to a function that prepares the url to shorten the url? To be able to add test and to assure that in all the cases it is going to work

CandelR (approved):

PR #1539, opened 2025-05-26
[_]:fix/ Add logic to handle closing of banner for free users

jaaaaavier:
## Description
- Added `onCloseFreeBanner` method to `BannerManager` to store close date for free user banners.
- Updated `BannerWrapper` to support `onCloseFreeBanner` as separate handler.

This enables better control and tracking of banners shown to free users.
## Related Issues
Addressing customer complaints regarding excessive banner frequency

## Checklist

- [ ] QA Passed

### DIFF
--- a/src/app/banners/BannerManager.ts
+++ b/src/app/banners/BannerManager.ts
@@ -67,11 +67,9 @@ export class BannerManager {
     };
   }
 
-  public onCloseBanner(): void {
-    localStorageService.set(BANNER_NAME_IN_LOCAL_STORAGE, this.todayDate);
-  }
+  public onCloseBanner(type: 'free' | 'subscription'): void {
+    const key = type === 'free' ? BANNER_NAME_FOR_FREE_USERS : BANNER_NAME_IN_LOCAL_STORAGE;
 
-  public onCloseFreeBanner(): void {
-    localStorageService.set(BANNER_NAME_FOR_FREE_USERS, this.todayDate);
+    localStorageService.set(key, this.todayDate);
   }
 }
--- a/src/app/banners/BannerManager.ts
+++ b/src/app/banners/BannerManager.ts
@@ -8,13 +8,16 @@ const BANNER_NAME_FOR_FREE_USERS = 'show_free_users_banner';
 export class BannerManager {
   private readonly plan: PlanState;
   private readonly offerEndDay: Date;
-  private readonly bannerItemInLocalStorage: string | null;
   private readonly todayDate: string;
+
+  private readonly localStorageKeys = {
+    freeUsers: BANNER_NAME_FOR_FREE_USERS,
+    general: BANNER_NAME_IN_LOCAL_STORAGE,
+  };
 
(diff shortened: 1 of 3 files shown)


src/app/banners/BannerManager.ts
```diff
@@ -70,4 +70,8 @@ export class BannerManager {
   public onCloseBanner(): void {
     localStorageService.set(BANNER_NAME_IN_LOCAL_STORAGE, this.todayDate);
   }
+
```
xabg2: Perhaps we can reuse the same function and pass the name of the key (banner) through parameters. This way we avoid having as many onCloseBanner functions as banners we have.

jaaaaavier: Done!

xabg2: I was referring to the key used in localStorage, which in this case would be:
- `BANNER_NAME_FOR_FREE_USERS`
- `BANNER_NAME_IN_LOCAL_STORAGE`

The conditional logic you added could grow significantly and become harder to maintain.
By passing the key of the localStorage item directly, the logic stays cleaner and easier to manage.

jaaaaavier: Okay

src/app/banners/BannerManager.ts
```diff
@@ -8,35 +8,39 @@ const BANNER_NAME_FOR_FREE_USERS = 'show_free_users_banner';
...
 
   private isOfferExpired(): boolean {
     return new Date() > this.offerEndDay;
   }
 
-  private isLocalStorageExpired(): boolean {
-    return (this.bannerItemInLocalStorage ?? '') < this.todayDate;
+  private isLocalStorageExpired(key: string): boolean {
+    const storedDate = localStorageService.get(key) ?? '';
+    return storedDate < this.todayDate;
```
CandelR, line 30: This function is not right, you can pass it any key and it doesn't have to return a date. Check it and add tests for the `BannerManager`

src/app/banners/BannerManager.ts
```diff
@@ -67,7 +71,7 @@ export class BannerManager {
     };
   }
 
-  public onCloseBanner(): void {
-    localStorageService.set(BANNER_NAME_IN_LOCAL_STORAGE, this.todayDate);
+  public onCloseBannerByKey(key: string): void {
+    localStorageService.set(key, this.todayDate);
```
CandelR, line 75: Same here, you can pass any key, check and add tests

src/app/payment/utils/checkStarWarsCode.test.ts
```diff
@@ -4,6 +4,7 @@ import localStorageService from 'app/core/services/local-storage.service';
 import paymentService from '../services/payment.service';
 import errorService from 'app/core/services/error.service';
 
+// Mocks reforzados
```
CandelR, line 7: Forgot to remove this unnecessary comment

CandelR (changes requested):

PR #1540, opened 2025-05-26
[PB-4467]: feat/centralize environment variables

xabg2:
## Description
This PR centralizes all environment variable access into a single configuration file.

This is the first step in the migration from CRA to Vite. Since Vite uses `import.meta.env` instead of `process.env`, and requires variable names to start with `VITE_` instead of `REACT_APP`, this change ensures that the migration will only require updates in one place: `env.service.ts`.

By centralizing environment access, we improve maintainability and reduce the risk of inconsistent usage throughout the codebase.

## Checklist

- [x] QA Passed

## Testing Process

To validate this change, the application should be tested thoroughly to ensure that environment variables are correctly resolved and that all core functionalities continue to work as expected. This includes interacting with key features such as uploading, downloading, renaming files, and general navigation throughout the app. The goal is to confirm that no regressions were introduced by centralizing environment variable access.


src/app/drive/services/network.service/index.ts
```diff
@@ -185,7 +186,7 @@ export class Network {
         {
           label: 'OneStreamOnly',
           params: {
-            useProxy: process.env.REACT_APP_DONT_USE_PROXY !== 'true',
+            useProxy: envConfig.services.dontUseProxy !== 'true',
```
TamaraFinogina, line 189: Maybe change to .useProxy? and use it directly without !== true?

src/app/newSettings/Sections/Account/Plans/api/plansApi.test.ts
```diff
@@ -21,13 +21,19 @@ vi.mock('../../../../../core/services/env.service', () => ({
   default: {
     isProduction: vi.fn(),
   },
+  envConfig: {
```
TamaraFinogina, line 26: Shouldn't it be restored back after the test?

TamaraFinogina:
@xabg2 Tests are failing. I restarted them, but they failed again with the same error

src/app/drive/services/network.service/index.ts
xabg2: We could do so, but in this PR I intend to simply use the variables we centralize. Changing this involves changing the name of the variable and so on, and I think it should not be within this PR.

sg-gs (commented):
Although this change is under 500 LOCs, from my POV the impact could be a disaster is something fails. Hence, to be time-efficient, I would suggest two things in order to ensure the quality (apart from the @ValentynaBlahodyrQA's QA, ofc): 
- Adding a unit test just to ensure that the environment variables imported from the new centralised point have, in fact, the expected values (by using the .env.template values or something like that)
- To make a list of all the endpoints affected by this changes, and either test those functionalities or at least trigger the requests so we can verify through the 'Network' window that the corresponding domains and values are what they are expected to be. 

Please take into account this can break a lot of features if something goes wrong, including but not limited to: customer payments, account creation, workspaces guest sign up, backups, sharing, etc.

CandelR (commented):
LGTM, waiting for the QA of the remaining enpdoints to approve

CandelR (approved):

PR #1546, opened 2025-05-30
[PB-4369]: chore/migrate limit and usage calls clean

jzunigax2:
## Description

Migrate the create get usage and limit endpoints to use the new endpoints exposed in drive-server-wip

## Related Pull Requests

#internxt/sdk/pull/289
#internxt/drive-server-wip/pull/549

## Checklist

- [x] Changes have been tested locally.
- [x] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [x] SonarCloud issues have been reviewed and addressed.
- [x] QA Passed

## How Has This Been Tested?

Checking account limit and usage both on the lower section of the left sidebar and on settings under account details.

PR #1548, opened 2025-06-03
[PB-4520]: feat/migrate-storage-endpoints

larryrider:
## Description
Migrates storage endpoints from drive-old to drive-new.

- Delete Folder: DELETE drive-old/storage/folder/{folderId} → DELETE drive/folders/{folderId}
- Delete File: DELETE drive-old/storage/folder/${folderId}/file/${fileId} → DELETE drive/files/{fileId}
- Storage Limit: GET drive-old/limit → GET drive/users/limit
- Storage Usage: GET drive-old/usage → GET drive/users/usage

## Related Pull Requests

https://github.com/internxt/sdk/pull/297

## Checklist

- [X] SonarCloud issues have been reviewed and addressed.
- [x] QA Passed


src/app/drive/services/limit.service.ts
```diff
@@ -3,8 +3,8 @@ import { SdkFactory } from '../../core/factory/sdk';
 import { HUNDRED_TB } from 'app/core/components/Sidenav/Sidenav';
 
 async function fetchLimit(): Promise<number> {
-  const storageClient = SdkFactory.getInstance().createStorageClient();
-  return storageClient.spaceLimit().then((response) => {
+  const storageClient = SdkFactory.getNewApiInstance().createNewStorageClient();
+  return storageClient.spaceLimitV2().then((response) => {
```
CandelR, line 7: This change is at https://github.com/internxt/drive-web/pull/1546, if this merges, maybe we should close it

src/app/store/slices/plan/index.ts
```diff
@@ -203,7 +203,7 @@ export const planSlice = createSlice({
         state.businessPlanUsageDetails = {
           driveUsage,
           backupsUsage,
-        } as unknown as UsageResponse;
+        } as unknown as UsageResponseV2;
```
CandelR: Could you remove the `unknown` cast?

src/app/drive/services/usage.service.ts
```diff
@@ -8,21 +8,21 @@ export interface UsageDetailsProps {
   backups: number;
 }
 
-export async function fetchUsage(): Promise<UsageResponse> {
-  const storageClient = SdkFactory.getInstance().createStorageClient();
-  const driveUsage = await storageClient.spaceUsage();
+export async function fetchUsage(): Promise<UsageResponseV2> {
+  const storageClient = SdkFactory.getNewApiInstance().createNewStorageClient();
+  const driveUsage = await storageClient.spaceUsageV2();
```
CandelR, line 13: Same with this related with same PR https://github.com/internxt/drive-web/pull/1546

CandelR (approved):

src/app/store/slices/plan/index.ts
larryrider: Sure, but now I think its better to fix the property at all instead of force-casting it

larryrider: fixed!

package.json
```diff
@@ -173,7 +173,7 @@
     "webpack-bundle-analyzer": "^4.9.1"
   },
   "engines": {
-    "node": ">=14.17.0 < 21"
+    "node": ">=14.17.0"
```
CandelR: In prod we use node 18, I left a checkbox to tick when have checked that it works as expected:
- [ ] Checked that deploy works correctly

src/app/network/crypto.test.ts
```diff
@@ -112,52 +111,6 @@ describe('Test crypto.ts functions', () => {
     expect(hash).toBe('422dab11a4f44c2ceab1f8ef39827109989607d6');
   });
 
-  it('getEncryptedFile should return the same result as before', async () => {
```
CandelR: Leave the reason in `Notes` section why this test has been deleted please 🗒️

package.json
```diff
@@ -66,7 +67,7 @@
     "react-international-phone": "^4.3.0",
     "react-live-chat-loader": "^2.8.2",
     "react-loading-skeleton": "^2.2.0",
-    "react-pdf": "^7.7.3",
+    "react-pdf": "^9.2.1",
```
TamaraFinogina: Revert it. It's breaking the PDF  preview.  Tried to change it once too, and no luck, don't know what's causing it
Screenshot 2025-06-13 at 14 14 45

package.json
larryrider: I think its better to move this to another PR, so we do not mix things

src/app/network/crypto.test.ts
larryrider: I think its better to move this to another PR, so we do not mix things

package.json
larryrider: I think its better to move this to another PR, so we do not mix things.
I have reverted it

src/app/drive/components/FileViewer/viewers/FilePdfViewer/FilePdfViewer.tsx
```diff
@@ -57,7 +57,7 @@ const DEFAULT_ZOOM = 1;
 
 const FilePdfViewer = (props: FormatFileViewerProps): JSX.Element => {
   const { translate } = useTranslationContext();
-  const [fileUrl] = useState(URL.createObjectURL(props.blob));
+  const [fileUrl] = useState(URL.createObjectURL(new Blob([props.blob], { type: 'application/pdf' })));
```
TamaraFinogina: was it related to react-pdf change?

TamaraFinogina:
The test that was working before fails now: `src/app/share/views/SharedGuestSignUp/ShareGuestSignUpView.test.tsx > onSubmit > when called with new valid data, then user with decrypted keys is saved in local storage`

larryrider:
> The test that was working before fails now: `src/app/share/views/SharedGuestSignUp/ShareGuestSignUpView.test.tsx > onSubmit > when called with new valid data, then user with decrypted keys is saved in local storage`

those tests are randomly failing, but they are not related to my changes. It is a known bug that its still pending to fix. They fail, but after some retries, they start working again

src/app/drive/components/FileViewer/viewers/FilePdfViewer/FilePdfViewer.tsx
larryrider: yes it was, I have removed it from here as it is going to be added on its own PR

CandelR (approved):
Approved, let's see if we can check the problem with the tests that fail continuously, it's starting to be annoying and in the end we may miss something

PR #1550, opened 2025-06-04
[PB-4136]: feat/migrate from CRA to Vite

xabg2:
## Description

This PR includes the migration from Create React App (CRA) to Vite, along with the necessary configuration changes to support the new build tool.

**Key updates:**
- Replaced CRA tooling with Vite
- Updated scripts in package.json
- Adjusted configuration for environment variables, aliases, and static assets
- Ensured compatibility with existing features and libraries
- Improved build speed and development performance

## Checklist

- [x] QA Passed

## Testing Process

Test all the functionalities/endpoints of the application.


vite.config.ts
```diff
@@ -0,0 +1,57 @@
...
+      stream: 'stream-browserify',
+      zlib: 'browserify-zlib',
+      util: 'util',
+      url: 'url',
+    },
+  },
+  optimizeDeps: {
+    esbuildOptions: {
+      define: {
+        global: 'globalThis',
```
TamaraFinogina, line 56: Do we still need process.env? because I think we can do this for vite to see it: define: { 'process.env': ... }

package.json
```diff
@@ -143,9 +144,10 @@
     "@types/react-dom": "^18.2.13",
     "@types/react-redux": "^7.1.17",
     "@types/react-router-dom": "^5.3.3",
+    "@types/react-window": "^1.8.8",
     "@types/streamsaver": "^2.0.1",
     "@types/wicg-file-system-access": "^2020.9.4",
-    "@vitejs/plugin-react": "^4.3.3",
+    "@vitejs/plugin-react": "^4.5.0",
```
TamaraFinogina: Latest one is plugin-react@4.5.1, but I guess it will be detected automatically

vitest.shared.ts
```diff
@@ -1,25 +1,23 @@
...
 import path from 'path';
-import { defineConfig } from 'vitest/config';
-import svgr from '@svgr/rollup';
+import svgr from 'vite-plugin-svgr';
 
-export default defineConfig({
+export default {
   plugins: [
     react(),
     replace({
```
TamaraFinogina: Do we need this? I haven't found any use of 'process.browser'. Plus, when in local I deleted all replace, build and tests worked.
Screenshot 2025-06-06 at 11 00 01

package.json
```diff
@@ -86,9 +86,10 @@
   "scripts": {
     "preinstall": "node scripts/use-yarn.js",
     "prepare": "husky install",
-    "dev": "craco start",
-    "start": "craco start",
-    "build": "craco build",
+    "dev": "vite",
+    "build": "vite build",
+    "start": "vite",
```
TamaraFinogina: Why do we need "dev" and "start" if both are the same?

package.json
```diff
@@ -165,6 +167,9 @@
     "stylelint-scss": "^6.8.1",
     "tailwindcss": "^3.3.4",
     "typescript": "^4.4.2",
+    "vite": "^6.3.5",
```
CandelR: I think it would be better to leave the fixed version, without the ^

package.json
```diff
@@ -135,6 +135,7 @@
     "@types/async": "^3.2.7",
     "@types/file-saver": "^2.0.3",
     "@types/filesystem": "^0.0.36",
+    "@types/jszip": "^3.4.1",
```
TamaraFinogina, line 137: Why do we need it? Don't see it used anywhere

package.json
xabg2: Yeah, it has been updated 3 days ago. I'm going to update it and see if it works well 👍🏽

src/app/banners/BitDefenderBanner.tsx
```diff
@@ -31,19 +31,11 @@ const BitdefenderBanner = ({ showBanner, onClose }: BitdeFenderBannerProps): JSX
...
-            style={{ border: 'none', padding: 0, background: 'none', cursor: 'pointer' }}
-          >
-            <img
-              src={BitDefenderBanner}
-              alt="BitDefender Banner"
-              className="w-full h-auto"
-            />
+        <div className="w-[450px] lg:w-[750px]">
+          <button onClick={handleOnClick} style={{ border: 'none', padding: 0, background: 'none', cursor: 'pointer' }}>
+            <img src={BitDefenderBanner} alt="BitDefender Banner" className="w-full h-auto" />
```
CandelR, line 36: Avoid only formatted files in this type of PRs, as they are noisy and difficult the review

package.json
xabg2: We had these 2 before. Some of us use dev and other uses start but we can remove one of them.

package.json
xabg2: Yes, we use JSZip to handle folder downloads. This dependency defines the types for the package so that TypeScript can properly recognize and validate its usage.

CandelR (approved):

src/app/banners/BitDefenderBanner.tsx
TamaraFinogina: @CandelR I think it's because of tests. Style is only applied to files if all pre-commit things pass. So, if this file was uploaded with -n for git commit due to failed tests, then here we go. Do we have a task for fixing the tests btw?

TamaraFinogina (approved):

src/app/drive/services/icon.service.ts
```diff
@@ -2,21 +2,21 @@ import { FunctionComponent, SVGProps } from 'react';
...
-import { ReactComponent as VideoFile } from '../../../assets/icons/file-types/video.svg';
-import { ReactComponent as WordFile } from '../../../assets/icons/file-types/word.svg';
-import { ReactComponent as XlsFile } from '../../../assets/icons/file-types/excel.svg';
-import { ReactComponent as XmlFile } from '../../../assets/icons/file-types/code.svg';
-import { ReactComponent as CsvFile } from '../../../assets/icons/file-types/csv.svg';
-import { ReactComponent as ZipFile } from '../../../assets/icons/file-types/zip.svg';
+import LightFolder from '../../../assets/icons/light/folder.svg?react';
+import DefaultFile from '../../../assets/icons/file-types/default.svg?react';
+import AudioFile from '../../../assets/icons/file-types/audio.svg?react';
+import CodeFile from '../../../assets/icons/file-types/code.svg?react';
```
TamaraFinogina, line 8: @xabg2 Btw, Sonar complains that XmlFile and CodeFile import the same

xabg2: Solved!

CandelR (approved):

PR #1551, opened 2025-06-05
[_]: feat/Integrate Google Sheets API for logging offline conversions

jaaaaavier:
## Description
This PR adds Google Sheets integration for logging offline conversions

A sendConversionToAPI function was implemented to handle basic conversion data (GCLID, name, value, currency, and timestamp). It adjusts the timestamp (+2h) . A Vitest integration test was also added, running in a Node environment, this checks everything works as expected .

## Checklist

- [ ] QA Passed
- [x] Added env variables to production env ⚠️ IMPORTANT BEFORE MERGE ⚠️ 

## Testing Process
We wrote a test using Vitest that simulates a conversion and calls our sendConversionToAPI function. We ran the test using npx vitest run, and upon checking the Google Sheet, we confirmed the data had been successfully appended. This verified the integration end-to-end.

## Additional Notes
Expected Release: As soon as posible


CandelR (changes requested):
Hey, setting a private key as an environment variable is dangerous because it could be leaked, and this could lead to impersonation with gmail. 
This would be better done on the server side, so better to look for an alternative way to add this functionality than directly in drive-web

src/app/payment/views/IntegratedCheckoutView/CheckoutViewWrapper.tsx
```diff
@@ -438,6 +439,7 @@ const CheckoutViewWrapper = () => {
             seatsForBusinessSubscription,
           });
 
+        const gclid = localStorage.getItem('gclid');
```
CandelR: Where is the `gclid` object stored in the local storage? 
Add the key to the local storage service keys to have all keys in one place

jaaaaavier: Done

CandelR: but where is the `gclid` stored in the local storage? because here you try to get it but if it is never stored you will not get anything. 
I ask this because I don't see it in this PR, I don't know if it was already used before

CandelR (approved):

PR #1562, opened 2025-06-13
[PB-4465]: Feat/update-react-pdf

larryrider:
## Description

- Updates react-pdf to its latest version. 
- Allows the latest NodeJS version. (v23 is working fine)
- Removes asmcripto.js as the sha256 dependency is working as expected so another library to test this library is no needed.
- Fixes 'Each child in a list should have a unique key prop' error
- Fixes 'validateDOMNesting: <button> cannot appear as a descendant of <button>' error

## Related Issues

- Solves security issue: https://github.com/internxt/drive-web/security/dependabot/90

## Checklist

- [ ] QA Passed

## Additional Notes

There are some tests removed. The context is taht the library asmcripto.js was breaking vercel in NodeJS 22, and it has been deprecated/outdated.
Right now asmcripto is currently only used to test that other dependency (hash-wasm) is generating the same sha256. This tests were added when migrating from one dependency to the other. That's why they are no longer needed.

### DIFF
--- a/src/app/network/crypto.test.ts
+++ b/src/app/network/crypto.test.ts
@@ -12,7 +12,6 @@ import {
 } from './crypto';
 import { Buffer } from 'buffer';
 import crypto from 'crypto';
-import { Sha256 } from 'asmcrypto.js';
 import { getSha256 } from '../crypto/services/utils';
 import { Aes256gcmEncrypter, sha512HmacBufferFromHex } from '@internxt/inxt-js/build/lib/utils/crypto';
 import { mnemonicToSeed } from 'bip39';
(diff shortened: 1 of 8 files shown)


src/app/network/crypto.test.ts
sg-gs, line 1: Why these tests have been removed?

larryrider: @sg-gs The library asmcripto.js was breaking vercel in NodeJS 22, and it has been deprecated/outdated. 
Right now asmcripto is currently only used to test that other dependency (hash-wasm) is generating the same sha256. This tests were added by @TamaraFinogina when migrating from one dependency to the other. I have asked her about this and she said that they are no longer needed as the migration worked as expected.

TamaraFinogina: Yes, the test was just to make sure the migration to hash-wasm will not break anything.

TamaraFinogina:
@larryrider I checked the PDF viewer and it works. LGTM

TamaraFinogina (approved):

src/app/network/crypto.test.ts
```diff
@@ -112,52 +111,6 @@ describe('Test crypto.ts functions', () => {
...
-    }
-
-    const oldFile = createMockFile('file.txt', 13, 'text/plain');
-    const oldCipher = crypto.createCipheriv('aes-256-ctr', encryptionKey, iv);
-    const [oldEncryptedFile, oldHash] = await oldGetEncryptedFile(oldFile, oldCipher);
-    expect(encryptedFile).toStrictEqual(oldEncryptedFile);
-    expect(hash).toBe(oldHash);
-  });
-
   function getReadableStream() {
```
TamaraFinogina, line 114: getReadableStream() can be removed too I think

CandelR (approved):

PR #1565, opened 2025-06-16
[_]: feat/change offset hours for the gsheet send

jaaaaavier:
## Description
Renamed the date formatting function to better reflect what it actually does: format a date with a fixed UTC+1 offset (used for winter time in Spain, for example).

The previous name suggested it supported any custom timezone, but in reality we were always passing +1. Now the function is simpler, clearer, and harder to misuse.

Also cleaned up some internal logic:

- Removed the offset parameter (no longer needed)
- Replaced the hardcoded 3600000 with a named constant
- Updated usage and tests

## Checklist

- [x] QA Passed

## Testing Process
Maked an other purchase and xheked arrives the correct format, also checked the result of the function formatDateToCustomTimezoneString and works as expected


src/app/analytics/googleSheet.service.ts
```diff
@@ -16,10 +16,7 @@ function formatDateToCustomTimezoneString(date: Date, offsetHours: number): stri
   const minutes = String(adjusted.getMinutes()).padStart(2, '0');
   const seconds = String(adjusted.getSeconds()).padStart(2, '0');
 
-  const offset =
-    offsetHours >= 0
-      ? `+${String(offsetHours).padStart(2, '0')}00`
-      : `-${String(Math.abs(offsetHours)).padStart(2, '0')}00`;
+  const offset = '+0100';
```
CandelR: Add tests for this function and try not to hardcode the offset like this

CandelR (changes requested):

src/app/analytics/googleSheet.service.ts
```diff
@@ -4,21 +4,20 @@ import { CouponCodeData } from 'app/payment/types';
...
-  const year = adjusted.getFullYear();
-  const month = String(adjusted.getMonth() + 1).padStart(2, '0');
-  const day = String(adjusted.getDate()).padStart(2, '0');
-  const hours = String(adjusted.getHours()).padStart(2, '0');
-  const minutes = String(adjusted.getMinutes()).padStart(2, '0');
-  const seconds = String(adjusted.getSeconds()).padStart(2, '0');
+  const year = adjusted.getUTCFullYear();
+  const month = String(adjusted.getUTCMonth() + 1).padStart(2, '0');
+  const day = String(adjusted.getUTCDate()).padStart(2, '0');
+  const hours = String(adjusted.getUTCHours() + 1).padStart(2, '0');
```
CandelR: What is the reason for the +1? Is the time no longer adjusted in the adjusted variable?

src/app/analytics/googleSheet.service.ts
```diff
@@ -4,21 +4,20 @@ import { CouponCodeData } from 'app/payment/types';
 import { getProductAmount } from 'app/payment/utils/getProductAmount';
 
 const GSHEET_API = envConfig.app.websiteUrl;
+const WINTER_OFFSET_HOUR = 1;
 
-function formatDateToCustomTimezoneString(date: Date, offsetHours: number): string {
-  const adjusted = new Date(date.getTime() + offsetHours * 60 * 60 * 1000);
+export function formatDateToCustomTimezoneString(date: Date): string {
+  const offsetHours = WINTER_OFFSET_HOUR;                
+  const adjusted = new Date(date.getTime() + offsetHours * 3_600_000);
```
CandelR: Avoid magic numbers (3_600_000), add i to a const

src/app/analytics/googleSheet.service.ts
```diff
@@ -4,21 +4,20 @@ import { CouponCodeData } from 'app/payment/types';
 import { getProductAmount } from 'app/payment/utils/getProductAmount';
 
 const GSHEET_API = envConfig.app.websiteUrl;
+const WINTER_OFFSET_HOUR = 1;
 
-function formatDateToCustomTimezoneString(date: Date, offsetHours: number): string {
-  const adjusted = new Date(date.getTime() + offsetHours * 60 * 60 * 1000);
+export function formatDateToCustomTimezoneString(date: Date): string {
```
CandelR: I think you should change the name of the function, since it is not to format a date to a custom zone, if you want to get always the same time zone, add jsdoc explaining the purpose of the function so that there is no confusion and adapt the name.
Also update the description of the PR to explain the purpose more specifically, as it does not indicate what you want to achieve.

src/app/analytics/googleSheet.service.ts
jaaaaavier: Done

src/app/analytics/googleSheet.service.ts
jaaaaavier: Yes is redundant

src/app/analytics/googleSheet.service.ts
jaaaaavier: Done

CandelR (commented):
Ups, check the googleSheet tests, are failing

CandelR (approved):

PR #1566, opened 2025-06-16
[PB-2615]: feature/show correct shared item type

xabg2:
## Description

This PR fixes an issue where shared items incorrectly retained a 'public' type after being marked as 'restricted'. Previously, when an item was shared with specific users, the item remained marked as 'public' until at least one user accepted the invitation.

Now, when an item is shared with specific users and its sharing type is set to 'restricted', the item will immediately reflect this 'restricted' status, regardless of whether the invited users have accepted the invitation.

## Checklist

- [ ] QA Passed

## Testing Process

Share an item with someone and change the shared item to Restricted > Close the dialog > Reopen it and check that the shared item type is Restricted instead of Public.

Check it also in production to see the old/new behavior.

## Additional Notes

⚠️   This PR depends on this [SDK version](https://github.com/internxt/sdk/pull/300) to work. Do not merge it until the new SDK version is available.

**Update:**
✅ The SDK has been updated


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -246,11 +246,19 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
 
     const isItemNotSharedYet = !isAdvancedShareItem(itemToShare?.item) && !itemToShare.item.sharings?.length;
 
+    const sharingInfo = await shareService.getSharingInfo(itemId, itemType).catch((error) => {
+      if (error?.code === 'SHARING_NOT_FOUND') {
```
CandelR: This ‘code’ field that returns the call is correct?

vite.config.ts
```diff
@@ -54,6 +54,7 @@ export default defineConfig({
     open: true,
   },
   resolve: {
+    preserveSymlinks: process.env.NODE_ENV === 'development',
```
CandelR: 👌

src/app/drive/components/ShareDialog/ShareDialog.tsx
xabg2: Mmm no, it seems to be used only when calculating the size of the shared folder... removing it.

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -246,11 +246,16 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
+
+    sharingType = sharingInfo?.type ?? 'public';
+    isAlreadyPasswordProtected = sharingInfo?.publicSharing.isPasswordProtected ?? false;
+
     if (!isItemNotSharedYet) {
       try {
         const sharingData = await shareService.getSharingType(itemId, itemType);
-        sharingType = sharingData.type;
-        isAlreadyPasswordProtected = sharingData.encryptedPassword !== null;
         setSharingMeta(sharingData);
```
CandelR: What is the difference between `getSharingType` and `getSharingInfo` if we now get the type from the `getSharingInfo` call? Do we still need `getSharingType`?

xabg2: Yes, `getSharingType` retrieves the item's [metadata](https://github.com/internxt/sdk/blob/master/src/drive/share/types.ts#L344). However, it only works once the item is already shared.

On the other hand, `getSharingInfo` provides information configured by the user, even if the item is not shared yet (the invitation is not accepted or the shared link is used by nobody), such as: the current sharing type (e.g., public, restricted), whether a password is set, and the invitations sent for that link.

TL;DR:
`getSharingInfo` is primarily used to reflect the user’s configured sharing settings and actions in the frontend UI, while getSharingType focuses on the technical metadata of the item once it’s shared.

CandelR (approved):

PR #1573, opened 2025-06-19
[_] Remove all tests that fail sometimes

TamaraFinogina:
## Description

Some tests occasionally fail, but it's not clear why. The idea is to isolate failing tests and fix them one by one or find the common cause.

## Checklist

- [ ] QA Passed

## Testing Process

unit tests


sg-gs (commented):
I understand that this is due to some of them not working. Is preferable to comment them rather than wiping out everything. On the other hand there should be something else going on, as some of them like the Impact ones were crafted by @xabg2 not too long ago. 

Is this related to Vite, by any chance?

xabg2:
@sg-gs , the tests are failing randomly because of vitest. It seems that vitest runs the tests in parallel so some mocks are creating conflicts between them. The correct way to address this would be to use spyOn whenever possible and not mock everything.

I think, instead of deleting tests, it would be better to use `.skip` for those that fail from time to time. @TamaraFinogina

sg-gs:
If we switched from vitest to jest, we should ensure that nothing is broken, otherwise we leave the project with any kind of coverage every time we decide to change technologies/libraries. Please @TamaraFinogina  apply the suggested fixes by @xabg2. We cannot leave things broken or skipped or commented when they were working unless a major reason (which is not the case) arises

TamaraFinogina:
@sg-gs @xabg2 Some failures could be related to Vite, specifically to envConfig. I noticed that sometimes tests fail because the envConfig is altered by another test, and all variables the test needs suddenly become undefined. Even if beforeEach explicitly sets them to the test value. I think Vitest is just not good with mocking constants

TamaraFinogina:
@xabg2 As for removing module mocks, not sure it's the right way. I tried to remove as many as I could (and use spyOn only), but the number of test failures just increased. Maybe I need to remove all vi.mock to see the difference, don't know/

sg-gs:
So, we need to clarify where the fail is as if it is related to Vite, this should have been handled the moment we did the switch, not after, as it makes the tests useless and hence, the project stability disappears. @CandelR 

If not, it is still important to handle the issue as soon as possible, otherwise new PRs could be blocked as the validation is just not there, with all the corresponding consequences (bugs coming again, critical failures on payments and Impact integrations, etc ...)

**Tests are the core of the project**, let's clarify the situation so we do not go back in quality on drive web @TamaraFinogina @xabg2 @CandelR

PR #1574, opened 2025-06-20
[PB-4486]: feat/migrate-users-endpoints

larryrider:
## Description

- Migrates users endpoints from drive-old to drive-new.
    - Upload avatar: PUT drive-old/user/avatar → PUT drive/users/avatar
    - Refresh user: GET drive-old/user/refresh → GET drive/users/c/{user-id}
- Removes unused legacy code (migration is now finished)

## Checklist

- [ ] QA Passed

## Testing Process

Verify using the network tab from a browser that the user endpoints is using the drive-new backend


src/app/payment/services/checkout.service.ts
```diff
@@ -54,7 +54,7 @@ const getCustomerId = async ({
   customerId: string;
   token: string;
 }> => {
-  const checkoutClient = await SdkFactory.getInstance().createCheckoutClient();
+  const checkoutClient = await SdkFactory.getNewApiInstance().createCheckoutClient();
```
CandelR, line 57: just remember that this change will require re-testing the payment flow

CandelR (approved):

src/app/payment/services/checkout.service.ts
larryrider: Not really, since the "instance" is not being used by either the CheckoutClient or the PaymentsClient. It dont really cares if it uses getInstance or getNewApiInstance, the apiUrl is always the same

PR #1593, opened 2025-06-30
[PB-4573]: feat/check if user already uploaded a file

xabg2:
## Description

This PR updates the logic used to determine whether the onboarding tutorial should be shown.
Instead of relying on a localStorage flag, we now check if the user has any uploaded files in their account.
This ensures that users who have already completed the tutorial won’t see it again, even if they open the app in incognito mode or from a different browser.

## Checklist

- [x] QA Passed

## Testing Process

- Create new account
- Complete the tutorial
- Open the app in another browser
- Check that the tutorial is not displayed in your new account

**Additional check:**
- Log in in an old account, check that the tutorial is not displayed in some browsers (at least 2 different browsers)


src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -318,18 +317,17 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
...
-    };
-  }, []);*/
+    newStorageService
+      .hasUploadedFiles()
+      .then(({ hasUploadedFiles }) => {
+        setHasAnyUploadedFile(hasUploadedFiles);
+      })
+      .catch((error) => {
+        errorService.reportError(error);
+      });
```
CandelR: Maybe it is not necessary to execute this call every time.
If I remember correctly, the onboarding is only shown on the day that the user has registered. 
So if it is no longer that day, it is not necessary to execute the call. 
I would also check if there are any other conditions to take into account in order not to make unnecessary calls to the backend.
While we are at it, maybe you can take the logic out of showing the onboarding to a hook.
If this last suggestion is too time consuming, there is no need to do it now

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -208,20 +207,21 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
...
   const successNotifications = useTaskManagerGetNotifications({
     status: [TaskStatus.Success],
   });
   const divRef = useRef<HTMLDivElement | null>(null);
   const hasSignedToday = useAppSelector(userSelectors.hasSignedToday);
 
   const showTutorial =
+    hasAnyUploadedFile !== undefined &&
+    !hasAnyUploadedFile &&
     envService.isProduction() &&
```
CandelR, line 223: why have removed `isSignUpTutorialCompleted`?

xabg2: I think it is no longer necessary. Imagine the user creates the account in Chrome and completes the tutorial.
If he opens Brave and logs into our product, he will have to do the onboarding again, because the local storage value will not be there.

src/app/drive/components/DriveExplorer/DriveExplorer.tsx
xabg2: Yep, I thought about extracting the logic to a custom hook, but I think we need this bug fixed ASAP 😄. Anyway, let me see what I can do with this.

on the other hand, I will add a check if the account has been created today.

CandelR: Ok, so let's focus on the bug and the refactor some other time, which may take longer than necessary :)

CandelR (approved):

PR #1604, opened 2025-07-04
[PB-4381]: feat/use the avatar blob from the local DB or fetch it again

xabg2:
## Description

This PR improves avatar handling in the Shared dialog by:
- Using the cached avatar blob from IndexedDB when available.
- Automatically downloading and updating the avatar if a newer version exists.

This ensures faster loading and up-to-date avatars in shared views.

## Checklist

- [ ] QA Passed

### DIFF
--- a/src/app/core/components/Sidenav/WorkspaceSelector.tsx
+++ b/src/app/core/components/Sidenav/WorkspaceSelector.tsx
@@ -107,8 +107,6 @@ const WorkspaceSelector: React.FC<WorkspaceSelectorProps> = ({
     };
   });
 
-  console.log(selectedWorkspace);
-
   return (
     <div className="relative mb-2 inline-block w-full" ref={dropdownRef}>
       {/* TOGGLE BUTTON */}
--- a/src/app/core/components/Sidenav/WorkspaceSelector.tsx
+++ b/src/app/core/components/Sidenav/WorkspaceSelector.tsx
@@ -3,6 +3,7 @@ import { useEffect, useRef } from 'react';
 
 import { useTranslationContext } from '../../../i18n/provider/TranslationProvider';
 import WorkspaceAvatarWrapper from '../../../newSettings/Sections/Workspace/Overview/components/WorkspaceAvatarWrapper';
+import AvatarWrapper from 'app/newSettings/Sections/Account/Account/components/AvatarWrapper';
 
 export interface Workspace {
   uuid: string;
@@ -44,12 +45,16 @@ const WorkspaceCard = ({
   return (
     <button className="w-full px-2 py-3 text-left hover:bg-gray-5 dark:hover:bg-gray-10" onClick={handleOnClick}>
       <div className="flex w-full flex-row items-center justify-between space-x-2">
-        <WorkspaceAvatarWrapper
-          diameter={28}
-          workspaceId={workspace.uuid}
-          fullName={workspace?.name ?? ''}
-          avatarSrcURL={workspace.avatar}
-        />
+        {workspace && workspace?.type === 'Personal' ? (
+          <AvatarWrapper avatarSrcURL={workspace.avatar ?? null} fullName={workspace.name ?? ''} diameter={28} />
+        ) : (
+          <WorkspaceAvatarWrapper
+            diameter={28}
+            workspaceId={workspace?.uuid ?? ''}
+            fullName={workspace?.name ?? ''}
+            avatarSrcURL={workspace?.avatar ?? null}
+          />
+        )}
         <div className="flex grow flex-col truncate">
           <p className="truncate text-sm font-medium leading-4 text-gray-100">{workspace.name}</p>
           <p className="truncate text-xs font-medium leading-3 text-gray-60">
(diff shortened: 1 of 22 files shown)


src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -172,6 +181,32 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
...
+    }
+
+    const existsNewAvatar = databaseAvatarData.uuid !== extractAvatarURLID(url);
+
+    if (existsNewAvatar) {
+      return downloadAndSaveAvatar(url);
+    }
+
+    setAvatarBlob(databaseAvatarData.avatarBlob);
+  }, []);
```
CandelR: This is repeated several times throughout the project. I think it would be better to extract it to only keep it in one place, this way we also separate logic from the component

src/hooks/useAvatar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  saveAvatarToDatabase: (url: string, blob: Blob) => Promise<void>;
+  deleteDatabaseAvatar: () => Promise<void>;
+  downloadAvatar: (url: string) => Promise<Blob>;
+};
+
+const showUpdateAvatarErrorToast = () =>
+  notificationsService.show({
+    text: t('error.updateAvatarError'),
+    type: ToastType.Error,
+  });
```
CandelR: I think it would be better to pass it through props, so we respect the principle of single responsibility

src/hooks/useAvatar.tsx
```diff
@@ -0,0 +1,77 @@
...
+        await deleteDatabaseAvatar();
+        setAvatarBlob(null);
+      } catch {
+        showUpdateAvatarErrorToast();
+        setAvatarBlob(null);
+      }
+    };
+
+    handleAvatarData();
+  }, [avatarSrcURL]);
```
CandelR: Maybe we should add some kind of protection against race conditions in case avatarSrcURL changes rapidly, so as not to accumulate simultaneous downloads and unexpected states. Like two avatars start downloading, and one overwrites the other

src/app/drive/components/ShareDialog/ShareDialog.tsx
```diff
@@ -581,6 +571,7 @@ const ShareDialog = (props: ShareDialogProps): JSX.Element => {
                     <User
                       user={user}
                       key={user.email}
+                      avatarSrc={avatarBlob ? URL.createObjectURL(avatarBlob) : user.avatar}
```
CandelR: the user.avatar is not the same as the one that is passed to the hook?

src/hooks/useAvatar.tsx
xabg2: It's a generic message, I think it should be there directly instead of pass it through props 3 times in different places, wdyt?

CandelR: Not the message, the call to the notification display. In this way we remove one dependency from the hook file and you can use the form that you want in each site, it is not necessary that it is always the notification servicee, maybe in another one it is a simple error message in the component to give you an example

src/app/drive/components/ShareDialog/ShareDialog.tsx
xabg2: Yes, it is there as a fallback. We check if there is a new avatar to download and save it to the database. In case the custom hook fails, we have `user.avatarBlob` as a fallback, just in case.... Maybe it is still valid.

src/app/core/components/Sidenav/WorkspaceSelector.tsx
```diff
@@ -112,12 +119,20 @@ const WorkspaceSelector: React.FC<WorkspaceSelectorProps> = ({
...
+              diameter={28}
+            />
+          ) : (
+            <WorkspaceAvatarWrapper
+              diameter={28}
+              workspaceId={selectedWorkspace?.uuid ?? ''}
+              fullName={selectedWorkspace?.name ?? ''}
+              avatarSrcURL={selectedWorkspace?.avatar ?? null}
+            />
+          )}
```
CandelR, line 133: Why is it necessary to differentiate now between personal account and workspace and not before to get the avatar?

xabg2: The reason is that one component uses the [workspace functions](https://github.com/internxt/drive-web/pull/1604/files#diff-70c2d6ef877c986f36c6933cc91a1ceb30e32fe226a491cb91063f1c2e0efb42) (e.g., `getDatabaseWorkspaceAvatar(workspaceId)`), while the other uses the [individual profile functions](https://github.com/internxt/drive-web/pull/1604/files#diff-14fb1bf1da801dac1eabb6668cccaf2513ca5e19665ac33eace76032f5376e46) (e.g., `getDatabaseProfileAvatar`). I’m not sure why this distinction was made initially, especially since both components ultimately render the same Avatar component. To simplify, we could consider using directly the Avatar component and handle the fetching of the avatarBlob in the parent component. This entails doing a refactor on how avatars are being handled/rendered right now. Do you think we can address it in this task itself, or should we do it in a different task?

CandelR: Fine, there is no need to apply further changes to this task if it already solves the avatar problem

CandelR (approved):

PR #1605, opened 2025-07-15
[PB-4671] Optimize encryption of big files

TamaraFinogina:
## Description

Benchmarks [PB-4623](https://inxt.atlassian.net/browse/PB-4623) demonstrated that the upload can be optimized by reducing the number of buffer copies. In my tests (see jira), the time to upload 8.7 GB decreased from 10 minutes in production to 6 minutes.

## Related Issues

[PB-4623](https://inxt.atlassian.net/browse/PB-4623)

## Checklist

- [x] QA Passed

## Testing Process

Uploaded from local (`yarn dev`), downloaded from prod, checked that the video works well. 
Repeated the same process with Vercel

[PB-4623]: https://inxt.atlassian.net/browse/PB-4623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PB-4623]: https://inxt.atlassian.net/browse/PB-4623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/network/crypto.ts
```diff
@@ -92,42 +91,57 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
 }
 
-/**
- * Given a stream and a cipher, encrypt its content on pull
- * @param readable Readable stream
- * @param cipher Cipher used to encrypt the content
- * @returns A readable whose output is the encrypted content of the source stream
- */
-export function encryptReadablePull(readable: ReadableStream<Uint8Array>, cipher: Cipher): ReadableStream<Uint8Array> {
+export function encryptStreamInParts(
```
sg-gs, line 96: As this change is pretty delicate and any failure is prone to potential data loss, we need at least to cover 100% of this function. It has 4 situations, unless I am wrong: 

1. encryptedChunk < uploadChunkSize: keep reading
2. encryptedChunk >= uploadChunkSize: chunks can be enqueued
3. stream is finished: sent whatever is missing in the buffer (considering a pointer -as the buffer is circular- which in your case is `bufferLength`)

sg-gs: Also, the newBuffer thing is a bit confusing, why not consider using something like this?

```
export function encryptStreamInParts(
  plainFile: { size: number; stream(): ReadableStream<Uint8Array> },
  cipher: Cipher,
  uploadChunkSize: number,
): ReadableStream<Uint8Array> {
  const reader = plainFile.stream().getReader();
  let buffer = new Uint8Array(0);
  let finished = false;

  return new ReadableStream<Uint8Array>({
    async pull(controller) {
      // Read only if there is space to read
      while (buffer.length < uploadChunkSize && !finished) {
        const { done, value } = await reader.read();
        
        if (done) {
          const final = cipher.final();
          buffer = concat(buffer, final);
          finished = true;
        } else {
          const encrypted = cipher.update(value);
          buffer = concat(buffer, encrypted);
        }
      }

      // If there is data to send, send it. 
      if (buffer.length > 0) {
        const size = Math.min(buffer.length, uploadChunkSize);
        controller.enqueue(buffer.slice(0, size));
        buffer = buffer.slice(size);
      }

      // You have finished (as the if before will leave only the missing content in the buffer)
      if (finished && buffer.length === 0) {
        controller.close();
      }
    }
  });
}

function concat(a: Uint8Array, b: Uint8Array): Uint8Array {
  const result = new Uint8Array(a.length + b.length);
  result.set(a);
  result.set(b, a.length);
  return result;
}
```

sg-gs (changes requested):

src/app/network/crypto.ts
TamaraFinogina: @sg-gs, It's more memory-efficient to pre-allocate a bigger buffer than to concatenate two smaller ones (the first improvement I saw in benchmarks was when I removed mergeBuffers functions and did merge inline). 

Ideally, the `if` part should never happen, because `encryptedChunk` is much smaller than `uploadChunkSize` (around 65 KB vs 30 MB), which is why a pre-allocated buffer should always be used. I added this `if` just in case we want to play with encryption or chunks in the future.

TamaraFinogina: @sg-gs I can add tests after I'm back

sg-gs: Okay, considering that memory efficiency you highlight, is possible to still keep the logic proposal I pasted here? The double if and the return for finishing adds cognitive complexity without the need to do so

src/app/network/crypto.test.ts
```diff
@@ -27,47 +24,16 @@ describe('Test crypto.ts functions', () => {
...
-      const seed = (await mnemonicToSeed(mnemonic)).toString('hex');
-      return oldGetDeterministicKey(seed, bucketId).toString('hex').slice(0, 64);
-    }
-    function encryptMeta(fileMeta: string, key: Buffer, iv: Buffer): string {
-      const cipher: crypto.CipherCCM = Aes256gcmEncrypter(key, iv);
-      const cipherTextBuf = Buffer.concat([cipher.update(fileMeta, 'utf8'), cipher.final()]);
-      const digest = cipher.getAuthTag();
-      return Buffer.concat([digest, iv, cipherTextBuf]).toString('base64');
-    }
-    async function oldEncryptFilename(mnemonic: string, bucketId: string, filename: string): Promise<string> {
```
TamaraFinogina: I removed this test. It was needed when we migrated to hash-wasm to ensure correctness. Now it's redundant

src/app/network/crypto.test.ts
```diff
@@ -135,13 +101,136 @@ describe('Test crypto.ts functions', () => {
     expect(receivedBlobs.length).toBe(3);
   });
 
-  it('getSha256 should generate the same result as sha256 from crypto', async () => {
-    function oldSha256(input: Buffer): Buffer {
```
TamaraFinogina, line 148: Same with this test. It was only needed during migration to hash-wasm

src/app/network/crypto.ts
TamaraFinogina: @sg-gs I've added 3 tests:

1. File encryption in chunks gives the same result as encrypting the entire file
2. Case `chunkSize < uploadChunkSize` works, the last chunk is not lost,  and the buffer is NOT re-sized
3. Case `chunkSize > uploadChunkSize` works, the last chunk is not lost, and the buffer is re-sized

chunkSize = encryptedChunk for aes-ctr, so it's just a matter of creating the test file with smaller/bigger chunks than what we need to upload

TamaraFinogina: @sg-gs You mean the part with `finished = true;`? I think it's possible. It's just MDN docs prefer to close and return: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream

```
  if (done) {
              console.log("done", done);
              controller.close();
              return;
            }
```

Maybe @CandelR knows more about this. I've thought that there might be some unexpected error if there is a delay between done=true and the actual stream closing.

P.S.  I simplified the code a bit more

TamaraFinogina: @sg-gs On a second thought, bringing` && !finished` to the while loop will add some micro delays. Checking the condition will require evaluating a more complex condition: an extra logical AND and NOT (+ reading `finished`). So, it's 2 extra operations + variable accessing per loop. I would expect it to take a couple of more CPU cycles per loop. The delay should not be significant, though.

TLDR: I can do a performance test to check if there is any difference.

src/app/network/crypto.ts
```diff
@@ -97,72 +97,84 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
+        const { value, done } = await reader.read();
+
+        const encryptedChunk = done ? cipher.final() : cipher.update(value);
 
-      if (!status.done) {
-        controller.enqueue(cipher.update(status.value));
-      } else {
-        controller.close();
+        const requiredLength = bufferLength + encryptedChunk.length;
+        if (requiredLength > buffer.length) {
```
sg-gs, line 122: Why do you have preAllocated if you are still needed to expand?

src/app/network/crypto.ts
sg-gs: Not a performance topic @TamaraFinogina, my worries are about the complexity. The cryptography-related things should as simple as possible taking into account what needs to work efficiently. The function now uses a mix of buffers, checking it is done or not multiple times and using another buffer even if you already have a preAllocated buffer. 

We need it to be simpler than this one, multiple issues may arise for this explosion of branches to follow. The idea I gave at first is far simpler to understand. We may replace concat with a fixed buffer or by overwriting the buffer if you want.

sg-gs (changes requested):

src/app/network/crypto.ts
TamaraFinogina: @sg-gs, this `if` should never happen in a normal case. However, since I cannot get the `encryptedChunk` length before I start reading, this is a precaution to avoid buffer overflow errors.

src/app/network/crypto.ts
TamaraFinogina: @sg-gs, we cannot remove `if (requiredLength > buffer.length)`. With pre-allocation, we gain speed, but 'out of boundary' errors are fatal. Unless we can get `encryptedChunk.length` before we enter the while loop, we have to check for buffer bounds. With the current parameters (aes-crt, chunk size, and extra allocated memory), this check will always be false, and buffer expansion will remain as a precaution.  

If we remove buffer pre-allocation, we lose the speed. : 
Screenshot 2025-08-07 at 16 01 13

TamaraFinogina: @sg-gs I've changed the code to add the explicit `finished` flag. Now it should be similar to the code you posted, but with buffer pre-allocations.

sg-gs: I see, great benchmark, checking the code

src/app/network/crypto.ts
```diff
@@ -97,72 +97,85 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
-      } else {
+      if (bufferLength > 0) {
+        controller.enqueue(buffer.slice(0, bufferLength));
+        bufferLength = 0;
+      }
+
+      if (finished && bufferLength === 0) {
         controller.close();
       }
     },
```
sg-gs, line 147: Consider adding this, to correctly manage the entire flow: 
```
 cancel() {
    reader.cancel();
  }
```

sg-gs (changes requested):

src/app/network/NetworkFacade.ts
```diff
@@ -110,7 +112,7 @@ export class NetworkFacade {
          * TODO: Memory leak here, probably due to closures usage with this variable.
          * Pending to be solved, do not remove this line unless the leak is solved.
          */
-        fileToUpload = new Blob([]);
+        fileToUpload = new Uint8Array();
```
CandelR, line 114: nit: if we assign `null` instead of creating a new array, shouldn't it work the same but without creating a new array?

src/app/network/crypto.ts
```diff
@@ -97,72 +97,85 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
+export function encryptStreamInParts(
+  plainFile: { size: number; stream(): ReadableStream<Uint8Array> },
+  cipher: Cipher,
+  uploadChunkSize: number,
+  extraPreAllocatedSpace: number,
+): ReadableStream<Uint8Array> {
+  const readable = plainFile.stream();
+
   const reader = readable.getReader();
+  const preAllocated = uploadChunkSize + extraPreAllocatedSpace;
```
CandelR: nit: perhaps the name `bufferSize` makes the purpose of the const clearer

CandelR (approved):
I see it fine 🚀 👌  
I had a doubt about in which case the buffer size could be reallocated because one is already created with the size of the chunk + the extra one, but I have seen the comment that it is a precaution.
Waiting to see if @sg-gs has anything else to comment

src/app/network/NetworkFacade.ts
TamaraFinogina: It should, but it will be `fileToUpload = null as any` because of the types check

src/app/network/crypto.ts
TamaraFinogina: @sg-gs I've added cancel and a comment regarding bugger reallocation for clarity

P.S. Checked in vercel and calcel works

src/app/network/crypto.ts
TamaraFinogina: @CandelR Done! I also moved constants to a separate file, so it's easier to control their values and names

sg-gs:
I am removing @larry-internxt from the reviewers as he will be off the next week and we would deploy it then 
cc @TamaraFinogina @CandelR

src/app/network/crypto.ts
```diff
@@ -97,72 +93,90 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
+  cipher: Cipher,
+  uploadChunkSize: number,
+  allowedChunkOverhead: number,
+): ReadableStream<Uint8Array> {
+  const readable = plainFile.stream();
+
   const reader = readable.getReader();
+  const bufferSize = uploadChunkSize + allowedChunkOverhead;
+  let buffer = new Uint8Array(bufferSize);
+  let bufferLength = 0;
```
sg-gs: This is not the `bufferLength`, is the `offset` (the pointer you have to manage the buffer). The naming of this one is a bit confusing

TamaraFinogina:
@sg-gs @CandelR I did the following changes:
-  renamed `bufferLength` to `offset`
- added `console.warn` in case buffer overflow protection is triggered
- sending exactly `uploadChunkSize` now (except the very last chunk)
- modified the test to check that the warning is displayed and the very last chunk is not lost

P.S. I used `if (offset >= uploadChunkSize)` instead of `if (offset > 0)`, because I didn't want to complicate the `if` with the case `offset < uploadChunkSize`. Can change it if you think it would be more readable.

TamaraFinogina:
Run test in Vercel + Firefox, all chunks are of the same size (31,46 MB as specified by the `UPLOAD_CHUNK_SIZE` variable)

src/app/network/crypto.ts
```diff
@@ -97,72 +93,99 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
+          buffer = newBuffer;
+        }
 
-      if (!status.done) {
-        controller.enqueue(cipher.update(status.value));
-      } else {
+        buffer.set(encryptedChunk, offset);
+        offset += encryptedChunk.length;
+
+        if (done) finished = true;
```
sg-gs: `finished = done`

src/app/network/crypto.ts
```diff
@@ -97,72 +93,99 @@ export function encryptReadable(readable: ReadableStream<Uint8Array>, cipher: Ci
...
+  const bufferSize = uploadChunkSize + allowedChunkOverhead;
+  let buffer = new Uint8Array(bufferSize);
+  let offset = 0;
+  let finished = false;
+
+  return new ReadableStream<Uint8Array>({
     async pull(controller) {
-      console.log('2ND_STEP: PULLING');
-      const status = await reader.read();
+      while (offset < uploadChunkSize && !finished) {
```
sg-gs, line 114: What does happen if we have this case?
- `uploadChunkSize` is 10MB
- `value` from the reader is 1,1MB
- The 10th iteration, the situation is: 
   - `buffer` (size 10MB+64KB or whatever the overhead) has 9,9MB (1,1*9) of content
   - `encryptedChunk` is 1,1MB. 
   -  9,9+1,1 > 10MB+64KB. 

Then the overflow (which in theory, is a corner case) kicks in consistently,  fixing the situation and everything working smoothly. Hence, either the warning is not a warning or this happens more than expected. 

I would remove the warning then

src/app/network/crypto.test.ts
sg-gs, line 1: I am missing a test where the chunkSize is a multiple (an entire one) of the uploadChunkSize, to ensure that case is also handled

sg-gs (changes requested):

src/app/network/crypto.ts
TamaraFinogina: Yes, the overflow happens after 10 iterations, which leads to resizing to 11 MB. However, another resize can happen if (after some iterations) 11MB won't be enough. 

It should never happen with good parameters, because resizing can only happen if:

1. uploadChunkSize < encryptedChunk.length
2. AES-CTR is changed to another algorithm without adjusting other parameters
3. allowedChunkOverhead is changed to 0 or something smaller than 64*1024
4. Browsers suddenly change the default chunk size

About `allowedChunkOverhead = 64*1024`. This is the minimal possible value we can use as `encryptedChunk.length` is:

- `65536` in Firefox (exactly 480 pulls give uploadChunkSize, no allowedChunkOverhead space needed) 
- `2097157` in Chrome (15 pulls give uploadChunkSize + 65461, allowedChunkOverhead needed )
- `2097157` in Brave (same as in Chrome)
- `524288` in Safari (exactly 60 pulls give uploadChunkSize, no allowedChunkOverhead space needed) 

Btw, I can change it to double size `const bufferSize = 2*uploadChunkSize` so we don't deal with `allowedChunkOverhead` at all. It's a bit more memory, but should be ok still.

src/app/network/crypto.test.ts
TamaraFinogina: Added

src/app/network/crypto.ts
TamaraFinogina: Removed warning

src/app/network/crypto.ts
TamaraFinogina: Done

src/app/network/crypto.ts
```diff
@@ -131,7 +128,7 @@ export function encryptStreamInParts(
         buffer.set(encryptedChunk, offset);
         offset += encryptedChunk.length;
 
-        if (done) finished = true;
+        if (done) finished = done;
```
sg-gs, line 132: No if needed. finished = done always, it would work the same

src/app/network/crypto.test.ts
sg-gs: You added a test where the file size is multiple of the uploadChunkSize, although I was referring to "where the chunkSize is a multiple (an entire one) of the uploadChunkSize".

That covers the case of the reader giving you chunks which have an exact size matching with the uploadChunkSize after N iterations, covering the loop and the if behaviour when the buffer is exactly full. In fact, I should correct myself: test the case where the chunkSize is a multiple (an entire one) of the [uploadChunkSize+overhead] value

sg-gs (changes requested):

src/app/network/crypto.test.ts
TamaraFinogina: Oh, sorry, I thought you meant file size.

I [added](https://github.com/internxt/drive-web/pull/1605/commits/8f1690e55d70bf2e19da4e12037b3c26ee96ca6d) 3 more tests: `fileSize = chunkSize = uploadChunkSize`, `chunkSize = uploadChunkSize + overhead` and `uploadChunkSize = 2*chunkSize` (we already have tests for chunkSize > uploadChunkSize);

TamaraFinogina:
@sg-gs I've added tests

sg-gs (approved):
It's okay for me. Let's apply a QA for this one: 
- [x] Uploading a file <100MB from this version can be downloaded on production.
- [x] Uploading a file >100MB from this version can be downloaded on production.

[Ref](https://github.com/internxt/drive-web/blob/ef8a4ce6dd0f704a3c6af91d9729a162c7fd80ff/src/app/network/upload.ts#L101) to the multipart size criteria (100MB).

**All of these tests strictly require a sha256 hash integrity check.**

PR #1612, opened 2025-07-25
[_]: fix/user avatar

xabg2:
## Description
This PR implements a new thunk to refresh the user's avatar when it is expired. The function is triggered when the user opens the app or reloads it.

The idea is:

- Get the stored user data.
- Check the expiration date of the avatar:
   - If the avatar is expired, refresh the avatar URL and download the new blob.
   - If the avatar is still valid, use the cached blob.

By doing this, we ensure the app always has a valid avatar URL and a working blob to display the user's profile image properly.

## Checklist

- [ ] QA Passed

### DIFF
--- a/src/app/store/slices/user/index.ts
+++ b/src/app/store/slices/user/index.ts
@@ -21,7 +21,7 @@ import { workspacesActions } from '../../../store/slices/workspaces/workspacesSt
 
 import errorService from '../../../core/services/error.service';
 import { isTokenExpired } from '../../utils';
-import { syncAvatarIfNeeded } from '../../../utils/avatar/avatarUtils';
+import { refreshAvatar } from '../../../utils/avatar/avatarUtils';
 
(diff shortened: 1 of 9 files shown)


src/app/store/slices/user/index.ts
```diff
@@ -82,6 +83,37 @@ export const refreshUserThunk = createAsyncThunk<void, { forceRefresh?: boolean
...
+            ...currentUser,
+            avatar: refreshedAvatar ?? userAvatar,
+          }),
+        );
+        localStorageService.set(
+          LocalStorageItem.User,
+          JSON.stringify({
+            ...currentUser,
+            avatar: refreshedAvatar ?? userAvatar,
+          }),
```
CandelR: `setUser` already update the user local storage data

CandelR (approved):

PR #1614, opened 2025-07-30
[MKT_550]: fix/adding checkoutCard old features for a correct display

jaaaaavier:
## Description
In this PR, we introduce a change that fixes an existing issue in the checkout. We still have some pages where we sell the old plans, such as https://internxt.com/pccomponentes-products. On these pages, when clicking on the 2TB and 10TB plans, the checkout fails to display because it doesn’t have the corresponding features. For example, for the 5TB plan, the features shown are from the new 5TB plan. The goal of this PR is to fix this by adding the old features and applying some conditions so that the correct features are displayed for every plan.

New Changes:

1. Modified JSON structure for the checkoutPlan, now we have baseFeatures and specialFeatures for each one plan
2. Now we have a new utils file and his test that helps us to extract the correct path for title and features

## Checklist

- [ ] QA Passed

## Testing Process
Viewed these links, and checked everything works as expected

HomePage Plan
https://drive-web-git-checkout-fix-internxt-web-team.vercel.app/checkout?planId=price_1Qew3RFAOdcgaBMQ7DLcwILz&couponCode=SUMMER80&planType=individual&currency=eur&mode=payment

PcComponentes plan
https://drive-web-git-checkout-fix-internxt-web-team.vercel.app/checkout?planId=price_1OQ3H5FAOdcgaBMQwMJ734rd&couponCode=PCCOMPONENTES&planType=individual&currency=eur&mode=subscription


src/app/payment/components/checkout/CheckoutProductCard.tsx
```diff
@@ -75,15 +78,18 @@ export const CheckoutProductCard = ({
...
-    }) ?? translateList('preferences.account.plans.planFeaturesList.1GB.features');
+    translateList(
+      `preferences.account.plans.${planType}.${bytes ?? 'freeFeatures'}.${isOldPlan ? 'oldFeatures' : 'features'}`,
+      { returnObjects: true },
+    ) ?? translateList('preferences.account.plans.planFeaturesList.1GB.features');
 
   const comingSoonFeatureKeys = useMemo(() => {
     const result = translateList(`preferences.account.plans.${planType}.${bytes}.comingSoonFeatures`);
-    return Array.isArray(result) ? result : [];
+    return !isOldPlan && Array.isArray(result) ? result : [];
```
CandelR: I would add some security check so that if the result is not as expected it does not crash the web.
Have you checked what was the error that caused the crash? Start there and handle it, to avoid breaking the web

CandelR (changes requested):

src/app/i18n/locales/en.json
```diff
@@ -1905,6 +1917,35 @@
             ],
             "comingSoonFeatures": ["Device Cleaner", "Dark Web Monitor", "Meet", "Mail"]
           },
+          "2TB": {
```
xabg2: Maybe you could use a general template for all products instead of matching features by product individually. For example:
```json
"features": [
  "{{storage}} encrypted storage",
  "whatever"
]
```
Then dynamically replace the placeholder:
```ts
const features = translate(templateFeatures, { storage: "2TB" });
```
This way, even if the product isn't one of the predefined ones (i.e. not explicitly matched as we do with the new/actual plans), we still have a fallback to display predeterminate features—preventing broken or incomplete checkout experiences.
What do you think?

src/app/payment/components/checkout/CheckoutProductCard.tsx
xabg2, line 1: As I said in the previous comment, instead of checking if the product is old, we can simply do the following:
- If the product matches, display its features.
- If not, display a template for this case in which we use the product storage as the title and as the key for storage in the features.

src/app/payment/components/checkout/CheckoutProductCard.tsx
jaaaaavier: The error was that, since there were no features to display (because for the 2TB and 10TB plans they didn’t exist), the checkout appeared blank and it wasn’t possible to complete the purchase. By adding the corresponding features for those plans, this no longer happens.
I’m going to switch to the template suggested by @xabg2 and add it in all languages. I think it’s a good idea and it will solve the problem if something similar happens again in the future.

src/app/i18n/locales/en.json
jaaaaavier: Yes, I think it's a great idea, and it also works well for us to avoid breakages, as you said. I thought about it at first but dismissed it because the features aren't the same for each plan. But well, I think we'll easily find a solution.

src/app/payment/components/checkout/CheckoutProductCard.tsx
jaaaaavier: Understood

src/app/payment/components/checkout/CheckoutProductCard.tsx
```diff
@@ -52,6 +52,9 @@ export const CheckoutProductCard = ({
   const currencySymbol = Currency[selectedPlan.price.currency];
   const normalPriceAmount = selectedPlan.price.decimalAmount;
 
+  const oldPlanIds = ['price_1PNxZkFAOdcgaBMQi0UCtXBj', 'price_1OQ3H5FAOdcgaBMQwMJ734rd'];
```
jaaaaavier: On the other hand, what do you think about the way we're handling whether the 5TB plan is new or old? I know hardcoding it like this isn’t the best option. Any idea or suggestion? @CandelR @xabg2

xabg2: Is there any property of the product when fetching using the `/price-by-id` EP that can help us differentiate it?

jaaaaavier: The priceId itself is enough for us, but I'm not sure if the endpoint returns old ones.

CandelR (changes requested):
The website crashes when opening the settings modal. Try to test all flows that are affected by these changes.

jaaaaavier:
> The website crashes when opening the settings modal. Try to test all flows that are affected by these changes.


I just checked it and now I don’t see any crash, it’s pretty weird.

CandelR:
> > The website crashes when opening the settings modal. Try to test all flows that are affected by these changes.
> 
> I just checked it and now I don’t see any crash, it’s pretty weird.

Still crashing, navigate to the plans section

CandelR (changes requested):
After reviewing this PR, I think we should close it and rethink the approach.
Although the code works and solves the requirements, it introduces significant complexity which is going to make maintenance much more difficult:
- A lot of functions to resolve product features translations.
- Hardcoded special cases
- Fragile placeholders that silently break if translations change

Sometimes it's more worthwhile to keep things simple rather than making  everything dynamic. Over-engineering has made this difficult to understand  and maintain - any changes to translation keys or placeholder formats would break

jaaaaavier:
Closed with the purpose of creating something more organized.

PR #1617, opened 2025-07-30
[_]: feat/GTM AddShoppersTracking

jaaaaavier:
## Description
In this PR, we've implemented everything needed to support a new Google Tag Manager container on our website.  This container is intended to track all purchases made using the welcome coupon code, specifically for the company AddShoppers.While this could technically be done via Stripe, we've chosen this approach to test a campaign in which they also want to track clicks and other interactions.It's essentially a first step to see if a long-term collaboration with them is viable.

_Changes:_
We've added a new script to the index file to send data to the new container.
Although there's no issue with the G-XXXX key being visible, we can move it to an environment variable if you think it's better.

We now also want to store the email address of the user who makes the purchase, as it's one of the required values. So we save it and later store it in localStorage.

We've updated the file that sends the data, along with its test, to match what we need for the container.

## Checklist

- [ ] QA Passed

## Testing Process
During the testing process, here’s what we’ve done in the container: we defined two tags — a global one used for activation and data sending, and a conversion tag.

The first tag (the global one) can be tested and confirmed as active by adding the Vercel URL into a built-in tool within the container. It confirms that the tag is firing correctly, thanks to the script ending in G-XXXX.

The second tag is configured to fire when a user completes a purchase (i.e., when the URL contains /checkout/success).
This one couldn’t be tested via Vercel since the domain switches to drive.internxt... upon completing a real purchase. However, we simulated the condition by manually adding something like /checkout/success... to the Vercel URL, and confirmed that it does indeed activate as expected.

## Additional Notes
Working on confluence documentation


index.html
```diff
@@ -115,7 +115,18 @@
...
+      (function (w, d, s, l, i) {
+        w[l] = w[l] || [];
+        w[l].push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' });
+        var f = d.getElementsByTagName(s)[0],
+          j = d.createElement(s),
+          dl = l != 'dataLayer' ? '&l=' + l : '';
+        j.async = true;
+        j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
+        f.parentNode.insertBefore(j, f);
+       })(window, document, 'script', 'dataLayer', '%REACT_APP_GOOGLE_TAG_MANAGER%');
```
CandelR: The g-tag is twice. I guess you only want to have one

src/app/analytics/impact.service.ts
```diff
@@ -53,6 +55,12 @@ export function savePaymentDataInLocalStorage(
     localStorageService.set('priceId', selectedPlan.price.id);
     localStorageService.set('currency', selectedPlan.price.currency);
   }
+  if (email !== undefined) {
+    localStorageService.set('email', email);
+  }
```
CandelR: To make a purchase, user has to be logged in if I'm not mistaken, and the email address can be obtained from the user of the local storage. If there is a different reason, let me know

index.html
jaaaaavier: No, I want to have two, since it's information and scripts for two different agencies. It's not common, but when working with two separate agencies, it's one of the best options we have. It is possible, and sometimes the dataLayer is also changed, but in our specific case I don't understand why it hasn't worked. I've read that this can make things a bit more complicated. In the tests I did for both companies, the information was correctly collected and sent to both containers using the same dataLayer.

src/app/analytics/impact.service.ts
jaaaaavier: Yes i didn't note about that

src/app/analytics/addShoppers.services.ts
```diff
@@ -1,36 +1,36 @@
+declare global {
+  interface Window {
+    dataLayer: Record<string, any>[];
```
xabg2: Can you add this custom property for Window to [this file](https://github.com/internxt/drive-web/blob/master/src/react-app-env.d.ts#L26)?
That way we'll have everything centralized.

src/app/analytics/addShoppers.services.ts
```diff
@@ -1,36 +1,36 @@
...
+  couponCodeData,
+  email,
 }: {
-  orderId: string;
+  orderId: string | undefined;
   value: number;
-  currency: string;
-  offerCode?: string;
+  currency: string | undefined;
+  couponCodeData: string | undefined;
```
xabg2: `couponCodeData` is misleading, as the name suggests that we are obtaining data, but in reality we are obtaining a string. How about `couponCodeName` or something similar, to make it more understandable?

index.html
```diff
@@ -115,7 +115,18 @@
         j.async = true;
         j.src = 'https://www.googletagmanager.com/gtm.js?id=' + i + dl;
         f.parentNode.insertBefore(j, f);
-      })(window, document, 'script', 'dataLayer', '%REACT_APP_GOOGLE_TAG_MANAGER%');
+      })(window, document, 'script', 'dataLayer', 'GTM-5GWZKXKB');
```
xabg2: Add the new g-tag ID to an env variable as the other one.

src/app/analytics/addShopper.services.test.ts
```diff
@@ -1,69 +1,83 @@
...
-  });
-
-  it('should not execute if offerCode is not "welcome"', () => {
+  it('should push event to dataLayer if all fields are valid and couponCodeName is "welcome"', () => {
     sendAddShoppersConversion({
-      orderId: 'order999',
-      value: 10,
-      currency: 'usd',
-      offerCode: 'blackfriday',
+      orderId: 'order-123',
```
xabg2: It might be useful to extract it to an object, so you don't have to repeat the values every time you want to test it. 

Imagine you add a field or the type of a field changes. In that case, you would have to make the change five times. If you extract it to an object, you would only have to do it once.

src/app/analytics/addShopper.services.test.ts
```diff
@@ -1,69 +1,83 @@
...
+      value: 50,
+      currency: 'eur',
+      couponCodeName: 'WELCOME',
+      email: 'user@example.com',
     });
 
-    expect(document.head.querySelector('#AddShoppers')).toBeNull();
-    expect((window as any).AddShoppersConversion).toBeUndefined();
+    expect(window.dataLayer.length).toBe(1);
+    expect(window.dataLayer[0]).toEqual({
```
xabg2: Use `.toStrictEqual(...)` whenever possible.

src/app/analytics/addShopper.services.test.ts
```diff
@@ -1,69 +1,83 @@
...
-
-    expect((window as any).AddShoppersConversion).toEqual({
-      order_id: input.orderId,
-      value: input.value,
-      currency: 'EUR',
-      offer_code: input.offerCode,
+    testCases.forEach((input) => {
+      window.dataLayer = [];
+      sendAddShoppersConversion(input);
+      expect(window.dataLayer.length).toBe(0);
```
xabg2: Avoid using loops for assertions. When a test fails, you lose context about which specific test case caused the failure, making debugging much harder. Consider using individual test cases or assertions with descriptive messages.

xabg2 (approved):

PR #1633, opened 2025-08-19
[MKT_555]: feat/add OfferPromoCode to the data send to impact

jaaaaavier:
## Description
Updated trackPaymentConversion so that conversions are also sent when a couponCode is present, even if the source is 'direct' or missing.
The couponCode is now included in the event payload as order_promo_code.

**Reason & Context**

Previously we relied only on the referrer to decide whether to send events to Impact, which meant direct visits were not tracked.

Each affiliate already has a coupon code assigned in Impact, so we can attribute conversions using the coupon instead of the referrer.

This allows us to track redemptions in both Stripe and Impact, even when the visit comes directly.

The downside is that if a coupon is leaked or shared between users, the conversion would still be attributed to the affiliate. However, this is already mitigated with fraud detection and monitoring of leaked coupons.

## Related Pull Requests
https://github.com/internxt/impact-radius-integration/pull/1

## Checklist

- [ ] QA Passed


CandelR (approved):
Looks good for me, waiting @xabg2 approval that has context about IMPACT

CandelR:
Hey @jaaaaavier if this finally not has to be done remember to close the PR :)

PR #1653, opened 2025-09-09
[PB-4731]: add checkAvatarUrlWorking function and update refreshAvatar logic

terrerox:
## Description

Fix user avatar loading issues in AccountPopover component

## Checklist

- [ ] QA Passed


src/app/utils/avatar/avatarUtils.ts
```diff
@@ -55,18 +55,21 @@ function isAvatarExpired(url: string): boolean {
...
+  if (!avatarUrl) return null;
 
   const storedUserAvatar = await getDatabaseProfileAvatar();
 
   const shouldUpdate = !storedUserAvatar?.srcURL || isAvatarExpired(storedUserAvatar.srcURL);
 
-  if (!shouldUpdate) return storedUserAvatar?.srcURL;
+  if (!shouldUpdate) {
+    const isCurrentAvatarWorking = await userService.checkAvatarUrlWorking(storedUserAvatar?.srcURL);
+    if (!isCurrentAvatarWorking) return null;
```
xabg2: If the avatar does not work, it could be that the link has expired, for example - the function `checkAvatarUrlWorking` checks both, an empty/null URL string or if the fetch of the URL is ok or not. Wouldn't it be better to obtain a new user avatar URL and use it instead of returning null? Or am I missing something?

terrerox: Good catch! That approach was used because if the user removes their profile photo from one browser while also being logged in on another, the photo URL no longer exists. Instead of trying to use it, we just return `null` so the default icon with the user’s initials is displayed. That’s why the bug was happening. Do you recommend a different approach?

xabg2: Perhaps we can check if the EP we have to update the user's avatar returns a null value if the user has deleted it in another browser, as you said, since we have this “source of truth.” Mainly because of what I mentioned, if the avatar link has expired, we may have the same problem since you return null.

terrerox: ✅ **Done!**
The `avatar/refresh` endpoint is now the single **source of truth** for refreshing user avatars.

* If a user deletes their avatar in another session (or browser), refreshing the page will display the updated avatar state.
* The `avatar/refresh` endpoint also **caches the user’s avatar** to ensure fast response times.

You can check out the implementation here: [internxt/drive-server-wip#706](https://github.com/internxt/drive-server-wip/pull/706). @xabg2 @CandelR

src/hooks/useAvatar.tsx
```diff
@@ -35,7 +34,7 @@ export const useAvatar = ({
     async (url: string, signal?: AbortSignal) => {
       const databaseAvatarData = await getDatabaseAvatar().catch();
 
-      if (!databaseAvatarData || isAvatarExpired(databaseAvatarData.srcURL)) {
+      if (!databaseAvatarData) {
```
CandelR, line 37: why did you remove all the expiration check? in that way avatar is refreshed every time as user load web instead when it is needed if I'm not wrong

terrerox: A user might delete their profile image in another browser (and origin, e.g., localhost:3000). Since IndexedDB is scoped per origin and browser, they could still have a non-expired link locally, but it will fail because the file was removed from S3. @CandelR

CandelR (approved):
Approved @terrerox, merge it when pass the QA process :)

PR #1671, opened 2025-09-17
[PB-5020]: Fix sonar issues v2

terrerox:
## Description

Removed redundant cycles, reduced cognitive complexity, deleted an unused component and simplified nested code.

## Checklist

- [ ] QA Passed


src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,106 @@ export class DownloadManager {
     }
   };
 
+  private static readonly handleConnectionLostError = (err: unknown, taskId: string) => {
+    tasksService.updateTask({
+      taskId,
+      merge: { status: TaskStatus.Error, subtitle: t('error.connectionLostError') as string },
```
CandelR: I think that the cast to string is not necessary, could remove it :)

src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,106 @@ export class DownloadManager {
...
+    });
+    throw err;
+  };
+
+  private static readonly handleServerOrCancelledError = (
```
CandelR: this function name coud lead to confusion, the function returns boolean value, dependeing of certain conditions, maybe `shouldContinueAfterError` or something similar fits better with the purpose of the function

src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,106 @@ export class DownloadManager {
...
-          },
-        });
-        updateTaskWithErrorStatus = false;
-      } else {
-        this.removeRetryItems(items, downloadTask);
-      }
+      this.handleConnectionLostError(err, taskId);
     }
 
+    const updateTaskWithErrorStatus = this.handleServerOrCancelledError(
```
CandelR: same as the function name, this not fits boolean naming convention, seems more like a function name because it has an action on the name, something as `shouldContinueProcessing` fits better I think

src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,106 @@ export class DownloadManager {
...
+    const castedError = errorService.castError(err);
+    const errorText = items.length > 1 || !items[0].isFolder ? 'error.downloadingFile' : 'error.downloadingFolder';
+
+    notificationsService.show({
+      text: t(errorText, { message: castedError.message || '' }),
+      type: ToastType.Error,
+    });
+  };
+
+  private static readonly handleActiveTask = (
```
CandelR, line 290: why have added static in all new functions?

src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,106 @@ export class DownloadManager {
...
+        merge: { status: TaskStatus.Success },
+      });
+      return false;
+    }
+
+    this.removeRetryItems(downloadTask.items, downloadTask);
+    return true;
+  };
+
+  private static readonly reportErrorWithContext = (err: unknown, items: DownloadItemType[]) => {
```
CandelR, line 246: Does this function work for any error you want to report? Or is it for a specific case?

CandelR (changes requested):

src/app/network/DownloadManager.ts
terrerox: Since the `downloadQueue` public function is a static function, the rest of the private functions need to be static in order to be accessible by that function; otherwise, it will throw this error: `Property 'removeRetryItems' does not exist on type 'typeof DownloadManager'`.

src/app/network/DownloadManager.ts
terrerox: It works for any error I want to report, but the main reason was to reduce the code in the `reportError` function.

src/app/network/DownloadManager.ts
terrerox: Done!

src/app/network/DownloadManager.ts
terrerox: Done!

src/app/network/DownloadManager.ts
terrerox: Done!

src/app/network/DownloadManager.ts
```diff
@@ -164,79 +164,103 @@ export class DownloadManager {
...
-          },
-        });
-        updateTaskWithErrorStatus = false;
-      } else {
-        this.removeRetryItems(items, downloadTask);
-      }
+      this.handleConnectionLostError(err, taskId);
     }
 
+    const shouldContinueProcessing = this.shouldUpdateTaskWithError(downloadTask, isServerError, filePickerCancelled);
```
CandelR: To me, the name of the function and what it does still doesn't make sense; it's not clear.
It's called `shouldUpdateTaskWithError`, and it returns `true` if it doesn't update it and `false` if it does.
I recommend adding jsdoc to the functions of this class, because I find it a bit confusing. It's not the best solution, but I don't think we need to refactor this in this task either

CandelR (changes requested):
Remember to check the sonar cloud new issues

src/app/network/DownloadManager.ts
terrerox: Done! Do I need to wait for @internxt/cryptography ’s approval before sending this task to QA?

CandelR (approved):

src/app/network/DownloadManager.ts
CandelR: Not necessary in this one, pass it to QA

src/app/drive/types/download-types.ts
sg-gs, line 1: Types are not services, why is this file here?

src/app/network/DownloadManager.ts
```diff
@@ -164,83 +174,152 @@ export class DownloadManager {
...
+
+  /**
+   * Determines whether to continue processing after an error occurs
+   * Checks for retry tasks and handles task status updates accordingly
+   * @param downloadTask - The download task that encountered an error
+   * @param isServerError - Whether the error is a server error
+   * @param filePickerCancelled - Whether the file picker was cancelled
+   * @returns True if processing should continue, false otherwise
+   */
+  private static readonly shouldContinueAfterError = (
```
sg-gs, line 230: This function is a jack of all trades, master of none. It does far more than it's name indicates: 
- Decides if something should continue after error or not > OK
- Updates a task status (!)
- Removes the retry items (!)

Let's follow [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) as this kind of tricks are the main pain point when looking for bugs / side-effects or multiple executions of the same thing

src/app/network/DownloadManager.ts
```diff
@@ -164,83 +174,152 @@ export class DownloadManager {
...
+    return true;
+  };
+
+  /**
+   * Reports an error with contextual information about the download items
+   * Adds different context based on whether it's a single file, folder, or multiple items
+   * @param err - The error to report
+   * @param items - The download items that were being processed when the error occurred
+   */
+  private static readonly reportErrorWithContext = (err: unknown, items: DownloadItemType[]) => {
```
sg-gs, line 246: This is a good example of SRP: it just reports, as the function name suggests

sg-gs (changes requested):

src/app/drive/types/download-types.ts
sg-gs: This comment is unsolved @terrerox

terrerox: It seems you might be viewing an outdated diff, i've already moved it to the types folder. You can see it under Files changed. @sg-gs

sg-gs (approved):

PR #1674, opened 2025-09-17
[PB-5022]: Fix sonar issues v4

terrerox:
## Description

Reduce cognitive complexity, remove redundant cycles, and simplify nested code for:

File viewer

Upload folder

Upload files

Drag and drop

## Checklist

- [ ] QA Passed


src/app/drive/components/FileViewer/types.ts
sg-gs, line 1: Why not calling the file just 'types.ts'?

src/app/network/UploadFolderManager.ts
```diff
@@ -207,74 +208,62 @@ export class UploadFoldersManager {
...
-    if (!createdFolder) {
-      this.stopUploadTask(taskId, abortController);
-      this.killQueueAndNotifyError(taskId);
-      return;
-    }
-
-    if (!this.tasksInfo[taskId].rootFolderItem) {
-      this.tasksInfo[taskId].rootFolderItem = createdFolder;
-    }
+    return createdFolder || null;
```
sg-gs: Which is the utility of returning null?

src/app/drive/components/FileViewer/types.ts
terrerox: Good catch! I’ve changed the name. Thanks for pointing that out

src/app/network/UploadFolderManager.ts
terrerox: To ensure a falsy value for the condition on line 283, I simplified the code to return only undefined, since it is also falsy, instead of using `createdFolder || null`. This makes the code cleaner. What do you think?

sg-gs (approved):
LGTM

sg-gs:
Missing your review since last week @CandelR

PR #1682, opened 2025-09-30
[PB-4905]: Fix white screen blink on page load

terrerox:
## Description

The theme was being applied after React mounted, causing a visual flash. This fix adds an inline script in index.html that applies the dark mode class synchronously during page load, before React initialization. Also persists the isDark state properly using JSON.parse() instead of string comparison.
## Checklist

- [ ] QA Passed


src/app/theme/ThemeProvider.tsx
```diff
@@ -72,6 +72,10 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({ children }) => {
 
   const toggleTheme = (theme: Theme) => setCurrentTheme(theme);
 
+  const persistIsDark = (value: boolean) => {
```
CandelR: nit: the name `persistIsDark` is a bit confusing; you need to know what isDark is. `persistDarkTheme` makes it clearer what it does: persist the dark theme.

CandelR (approved):

PR #1685, opened 2025-10-01
[PB-4903]: Fix password modal closing on eye icon click

terrerox:
## Description

Add preventClosing prop to prevent modal from closing when clicking the password visibility toggle icon.

## Checklist

- [ ] QA Passed


src/app/newSettings/Sections/Account/Security/components/ChangePasswordModal.tsx
```diff
@@ -52,43 +56,45 @@ const ChangePasswordModal = ({
...
-            disabled={!passwordPayload.valid || passwordConfirmation !== passwordPayload.password}
-            className="ml-2"
-            dataTest="next-button"
-          >
-            {translate('modals.changePasswordModal.title')}
-          </Button>
-        </div>
-      </form>
+      <div onMouseDown={handleMouseDown} role="none">
+        <form onSubmit={handleSubmit}>
```
CandelR: Hey @terrerox, the project https://github.com/internxt/ui is ours. Perhaps it would be better to try adding the fix to the `Modal` component directly instead of adding a patch here, as it could happen in other places where the `Modal` component is used that the click bubbles up to the backdrop and closes because the click event comes from the children. 
What do you think?

terrerox: I see it’s already defined, thanks!

CandelR: What I have seen completely disables the functionality of closing the modal window when clicking outside of it, but I think what we should do is prevent the click event from bubbling in the components that we pass as children to the Modal, since what happens is undesirable

terrerox: Got it! I just opened a PR on [@internxt/ui](https://github.com/internxt/ui). Looking forward to your feedback. @CandelR

CandelR (approved):

PR #1692, opened 2025-10-15
[PB-4784]: feat enforce canonical drive domain redirect in production

terrerox:
Add logic to redirect all production traffic to the canonical drive.internxt.com domain. This ensures consistent URLs and prevents issues with multiple domains pointing to the same app.

Also refactor isProduction() check to be more robust by handling case-insensitive environment values and checking against a list of non-production environment identifiers.

## Checklist

- [ ] QA Passed

## Testing Process

Since this new behavior only occurs in **production**, I’ll modify the preview environment to reproduce it and verify if the issue behaves the same way before merging.

#### Steps to Test

1. Go to a **non-public URL**, for example:
   [https://feature-redirect-public-url.drive-web.pages.dev/login](https://feature-redirect-public-url.drive-web.pages.dev/login)
   **Expected Result:** The app should redirect to the main domain `https://drive.internxt.com`.

2. Go to a **public share URL**, for example:
   [https://feature-redirect-public-url.drive-web.pages.dev/sh/file/_gtOlt23Q5KBbyJqpMoT9w/eeae7a7d4b17278a85f600d3ff6b737cb18fbe96749841a5b695efcc1ef6aa28](https://feature-redirect-public-url.drive-web.pages.dev/sh/file/_gtOlt23Q5KBbyJqpMoT9w/eeae7a7d4b17278a85f600d3ff6b737cb18fbe96749841a5b695efcc1ef6aa28)
   **Expected Result:** The app **should not redirect** to `https://drive.internxt.com`.


infrastructure/preview.Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM node:18-alpine3.19
 
+ENV REACT_APP_IS_PREVIEW=true
```
xabg2: Why?

src/index.tsx
```diff
@@ -25,6 +25,48 @@ import { LiveChatLoaderProvider } from 'react-live-chat-loader';
...
+
+    return normalizedUrl.toString();
+  } catch {
+    return CANONICAL_DRIVE_ORIGIN;
+  }
+};
+
+const enforceCanonicalDriveDomain = () => {
+  const isBrowser = typeof globalThis !== 'undefined';
+  const isPreview = process.env.REACT_APP_IS_PREVIEW === 'true';
```
xabg2: `process.env` is no longer available in Vite. Vite exposes the environment variables under `import.meta.env`. More info [here](https://vite.dev/guide/env-and-mode.html#env-variables-and-modes).

terrerox: I’ve modified it using a different approach, what do you think?

src/index.tsx
```diff
@@ -25,6 +25,55 @@ import { LiveChatLoaderProvider } from 'react-live-chat-loader';
...
+  } catch {
+    return CANONICAL_DRIVE_ORIGIN;
+  }
+};
+
+const enforceCanonicalDriveDomain = () => {
+  const isBrowser = typeof globalThis !== 'undefined';
+
+  if (!isBrowser || !envService.isProduction()) {
+    return;
```
CandelR: Add tests for these functions. 
I believe this will break public shared links, as it will redirect to drive.internxt.com when entering another domain such as https://secure. internxt.com/sh/file/_gtOlt23Q5KBbyJqpMoT9w/eeae7a7d4b17278a85f600d3ff6b737cb18fbe96749841a5b695efcc1ef6aa28 for example.
Perhaps a whitelist of allowed domains should be added so that it does not redirect.

src/index.tsx
```diff
@@ -25,6 +25,55 @@ import { LiveChatLoaderProvider } from 'react-live-chat-loader';
...
+  }
+  const hostname = globalThis.location.hostname;
+  const isPreviewDeployment = hostname.includes('.pages.dev');
```
CandelR: Perhaps it is also better for preview environments to enable an environment variable, as this may be different at any given time and may not work correctly. This way, it couples with Cloudflare's preview links.

CandelR (commented):
Fill in the testing section to see how this feature will be tested

src/index.tsx
terrerox: But isn’t that the expected logic? To redirect all public share URLs that use secure.internxt.com to drive.internxt.com? And actually, it’s not just for shared URLs, according to the ticket, the logic must ensure that whenever the base URL isn’t https://drive.internxt.com, it redirects to that domain. @CandelR

CandelR: Only if the base URL or a derivative that is not a public shared file URL is used, that matche this style `https://secure/. internxt.com/sh/file/_gtOlt23Q5KBbyJqpMoT9w/eeae7a7d4b17278a85f600d3ff6b737cb18fbe96749841a5b695efcc1ef6aa28`

xabg2 (approved):

src/app/utils/canonicalDomain.utils.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      expect(mockEnvService.isProduction).toHaveBeenCalled();
+    });
+
+    // it('should skip redirect when dontRedirect is set', () => {
+    //   mockEnvService.isProduction.mockReturnValue(true);
+    //   mockEnvService.getVariable.mockReturnValue('true');
+
+    //   expect(() => enforceCanonicalDriveDomain()).not.toThrow();
+    //   expect(mockEnvService.getVariable).toHaveBeenCalledWith('dontRedirect');
+    // });
```
CandelR: Don't comment the tests, in this way we could ensure that if this does not pass we don't merge it. Don't worry if the test is failing while are waiting the QA, that is its purpose

terrerox: This is only to activate the behavior while Valentyna is testing. I couldn’t push the changes because they were failing. Once Valentyna finishes, I’ll delete that commit. @CandelR

PR #1698, opened 2025-10-21
[PB-4441] Refactor restructure Recents module to view-based architecture

terrerox:
## Description

Migrate Recents functionality from drive module to new view-based structure to align with project architecture guidelines. Move view component, services, and thunks to dedicated Recents folder under views.

Changes:
- Move RecentsView component to src/app/views/Recents/
- Relocate fetchRecentsThunk to Recents/services/
- Extract fetchRecents API service from fileService
- Update imports across affected modules
- Remove unnecessary type assertion

## Checklist

- [ ] QA Passed


src/views/Recents/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default } from './RecentsView';
+export { fetchRecents } from './services/fetchRecents';
+export { fetchRecentsThunk, fetchRecentsThunkExtraReducers } from './services/fetchRecentsThunk';
```
CandelR: Separate the thunk from services, the thunk is related with redux, add it to store folder in Recents to separate it from services. 
Also,  the current barrel export structure could be improved. Right now, it's mixing view and service exports in a single file, create a barrel for each responsibility, better to not mix them :)

src/views/Recents/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default } from './RecentsView';
+export * from './services';
+export * from './store';
```
CandelR: The barrel at `views/Recents/index.ts` should only export the view component itself, 
not re-export services and store. This creates unnecessary coupling and can lead to 
circular dependency issues

CandelR (changes requested):

src/views/Recents/services/fetchRecents.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    getNewApiInstance: vi.fn(),
+  },
+}));
+
+describe('fetchRecents', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should call getRecentFilesV2 with the correct limit', async () => {
```
xabg2: nit: better use a non-technical description for tests. Imagine that the function is renamed and now it is called getRecentFiles instead, you will update the tests because it will fail (or not) but the test description will cause confusion.

src/views/Recents/store/fetchRecentsThunk.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      { id: 1, name: 'encrypted-file1.txt', plainName: 'file1.txt' },
+      { id: 2, name: 'encrypted-file2.txt', plainName: 'file2.txt' },
+    ];
+    const mockFormattedRecents = [
+      { id: 1, name: 'file1.txt', plainName: 'file1.txt' },
+      { id: 2, name: 'file2.txt', plainName: 'file2.txt' },
+    ];
+    const mockRecentsLimit = 50;
+
+    (configService.getAppConfig as Mock).mockReturnValue({
```
xabg2: What about using spyOn instead of mocking that?

src/views/Recents/store/fetchRecentsThunk.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      fileExplorer: { recentsLimit: mockRecentsLimit },
+    });
+    (fetchRecents as Mock).mockResolvedValue(mockRecents);
+    (getItemPlainName as Mock).mockImplementation((item) => item.plainName);
+    (excludeHiddenItems as Mock).mockReturnValue(mockFormattedRecents);
+    (storageActions.setRecents as unknown as Mock).mockReturnValue({ type: 'storage/setRecents' });
```
xabg2: What about using spyOn?

xabg2 (approved):
LGTM! What do you think about the tests @terrerox ?

CandelR (approved):

PR #1700, opened 2025-10-23
[_] feature/Updated readme

CandelR:
## Description

- Updated the readme to provide clear example of new view based project structure
- Updated related tests
- Fixed TypeScript errors
- Added fallback to single download when try to use multipartdownload for shared files

## Related Issues

-
## Related Pull Requests

-

## Checklist

- [ ] QA Passed

## Testing Process

- No testing process


README.md
```diff
@@ -100,106 +100,278 @@ The [/src](./src) folder contains the source code.
...
+│   │
+│   ├── Shared/                    # Shared files view
+│   │   ├── index.tsx              # Shared page component
+│   │   ├── components/            # Shared-specific components
+│   │   │   ├── SharedFilesList.tsx
+│   │   │   └── ShareDialog.tsx
+│   │   ├── hooks/                 # Custom hooks for sharing
+│   │   │   └── useSharedFiles.ts
+│   │   ├── services/              # API calls for sharing
+│   │   │   └── shareService.ts
```
xabg2: I would use `share.service.ts` instead, as we do for types (shared.types.ts). WDYT?

README.md
```diff
@@ -100,106 +100,278 @@ The [/src](./src) folder contains the source code.
...
+│       ├── store/                 # Redux slice for trash
+│       │   └── trashSlice.ts
+│       └── types/                 # TypeScript types/interfaces
+│           └── trash.types.ts
+│
+├── shared/                        # Shared code across views
+│   ├── components/                # Reusable UI components
+│   │   ├── Button/
+│   │   │   ├── Button.tsx
+│   │   │   └── Button.types.ts
```
xabg2: Should the types file start in lower case?

README.md
```diff
@@ -100,106 +100,278 @@ The [/src](./src) folder contains the source code.
...
+│   │   ├── index.tsx              # Shared page component
+│   │   ├── components/            # Shared-specific components
+│   │   │   ├── SharedFilesList.tsx
+│   │   │   └── ShareDialog.tsx
+│   │   ├── hooks/                 # Custom hooks for sharing
+│   │   │   └── useSharedFiles.ts
+│   │   ├── services/              # API calls for sharing
+│   │   │   └── shareService.ts
+│   │   ├── store/                 # Redux slice for shared files
+│   │   │   └── sharedSlice.ts
```
xabg2: Should we do shared.slice.ts as well? 🤔

README.md
CandelR: Yes, this is just an example for the directory structure; I didn't mean that all the names should be changed. I'll change it anyway so that it matches the current one

README.md
CandelR: same as above comment, I'll change it

README.md
CandelR: This is simply an example of the structure; it does not mean that there must be a slice that is `sharedSlice`. There will be as many as there are now with current names. 
In any case, it seems better to change it to match the names as they are now in the project, as it could cause confusion.
Thanks @xabg2

PR #1703, opened 2025-10-27
[PB-4373]: feat/decrypt stream using offset byte

xabg2:
## Description

This PR enables decrypting encrypted streams from any byte position instead of requiring decryption from the beginning. The implementation adjusts the AES-CTR initialization vector based on the block number and skips intra-block bytes to align the decipher with the requested offset.

## Checklist

- [ ] QA Passed

## Additional Notes

This approach is based on the existing [CLI implementation](https://github.com/internxt/cli/blob/bd91485a7f51be32270e2d9e5abe585e08169399/src/services/crypto.service.ts#L122).

### DIFF
--- a/src/app/core/services/stream.service.test.ts
+++ b/src/app/core/services/stream.service.test.ts
@@ -16,7 +16,7 @@ vi.mock('app/network/download', () => ({
 
 import { decryptStream } from './stream.service';
 import { getDecryptedStream } from 'app/network/download';
-import crypto from 'crypto';
+import { createDecipheriv } from 'crypto';
 
 describe('Stream service', () => {
   let mockDecipher: any;
@@ -31,7 +31,7 @@ describe('Stream service', () => {
     mockDecipher = {
       update: vi.fn(),
     };
-    vi.mocked(crypto.createDecipheriv).mockReturnValue(mockDecipher);
+    (createDecipheriv as any).mockReturnValue(mockDecipher);
     mockInputSlices = [
       new ReadableStream({
         start(controller) {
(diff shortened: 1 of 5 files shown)


src/app/core/services/stream.service.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      new ReadableStream({
+        start(controller) {
+          controller.close();
+        },
+      }),
+    ];
+  });
+
+  describe('Decrypt Stream', () => {
+    test('When there is no offset, then should create a new cipher', () => {
```
TamaraFinogina: A bit misleading test description. Maybe something like 'When there is no offset, then should proceed with standard decryption' is better

xabg2: Changing it! 🚀

TamaraFinogina (approved):

TamaraFinogina:
LGTM, it's the same as in cli: https://github.com/internxt/cli/blob/ff528c4ae70339543a366603ab36b0ccc95042b6/src/services/crypto.service.ts#L118

P.S. Tests are failing due to the ~missing~ extra default in mocks

xabg2:
> LGTM, it's the same as in cli: https://github.com/internxt/cli/blob/ff528c4ae70339543a366603ab36b0ccc95042b6/src/services/crypto.service.ts#L118
> 
> P.S. Tests are failing due to the missing default in mocks

Yep, I don't know why because in local it works and does not throw the error. Fixing it...

sg-gs:
Please ping me on the associate task on Jira to check that the ACs are strict enough as this involves cryptography changes @xabg2

sg-gs (approved):

src/app/core/services/stream.service.test.ts
```diff
@@ -0,0 +1,70 @@
+import { describe, it, expect, vi, beforeEach, test } from 'vitest';
+
+vi.mock('crypto', () => ({
+  createDecipheriv: vi.fn(),
```
TamaraFinogina: btw, it's better not to mock crypto module. SpyOn should work too

xabg2: Yep. I used crypto directly in the service so I can use `crypto.createDecipheriv` and then spy on it in the test 👍🏽

PR #1712, opened 2025-11-03
[PB-4373]: feature/implementing multipart download for big single files

xabg2:
## Description

This PR implements the multipart download files for big single files. This is the following criteria:
- The file is bigger than or equal to 2GB - Multipart download
- The file is smaller than 2GB - Normal download (sequential)

## Checklist

- [ ] QA Passed

## Testing Process
Open the preview link:
- Download a file smaller than 2GB - See in the network tab (DevTools) that there is only one call downloading the file pointing to the S3 bucket.
- Download a big file bigger than or equal to 2GB - See in the network tab (DevTools) that there aire several calls downloading the file pointing to the S3 bucket.

P.d. After each try, check the file integrity by using checksum.


src/app/workers/downloadWorker.ts
```diff
@@ -1,6 +1,23 @@
 import { binaryStreamToBlob } from 'app/core/services/stream.service';
 import createFileDownloadStream from 'app/drive/services/download.service/createFileDownloadStream';
+import createMultipartFileDownloadStream from 'app/drive/services/download.service/createMultipartDownloadStream';
 import { DriveFileData } from 'app/drive/types';
+import { MIN_DOWNLOAD_MULTIPART_SIZE } from 'app/network/networkConstants';
+
+export interface DownloadFilePayload {
```
TamaraFinogina: Can we move it to some type file? Because importing `DownloadFilePayload` in another file can create vitest problems due to dependencies

src/app/workers/downloadWorker.ts
```diff
@@ -36,19 +38,21 @@ export class DownloadWorker {
...
       });
 
-      const downloadedFile = await createFileDownloadStream(
-        file,
-        isWorkspace,
-        callbacks.onProgress,
-        this.abortController,
-        credentials,
-      );
+      const downloadedFile: ReadableStream<Uint8Array<ArrayBufferLike>> = await getFileReadableStream();
```
TamaraFinogina: Is it the same as `ReadableStream<Uint8Array>`?

src/app/network/download/v2.ts
```diff
@@ -88,6 +89,38 @@ const downloadOwnFile: DownloadOwnFileFunction = async (params) => {
...
+  const networkFacade = new NetworkFacade(
+    Network.client(
+      envService.getVariable('storjBridge'),
+      {
+        clientName: 'drive-web',
+        clientVersion: '1.0',
+      },
+      {
+        bridgeUser: params.creds ? auth.username : '',
+        userId: params.creds ? auth.password : '',
```
TamaraFinogina: why`params.creds`? I think in all other cases we use `auth.password` because it's hashed version

src/app/drive/services/download.service/createMultipartDownloadStream.ts
```diff
@@ -0,0 +1,23 @@
+import { multipartDownloadFile } from 'app/network/download';
+import { DriveFileData } from '../../types';
+
+export default async function createMultipartFileDownloadStream(
+  itemData: DriveFileData,
+  updateProgressCallback: (progress: number) => void,
+  sharingOptions: { credentials: { user: string; pass: string }; mnemonic: string },
+  abortController?: AbortController,
+): Promise<ReadableStream<Uint8Array<ArrayBufferLike>>> {
```
TamaraFinogina: Do we need `<ArrayBufferLike>` here?

src/app/workers/downloadWorker.ts
```diff
@@ -1,6 +1,23 @@
...
+  isWorkspace: boolean;
+  isBrave: boolean;
+  credentials: any;
+}
+
+interface DownloadFileCallback {
+  onProgress: (progress: number) => void;
+  onSuccess: (fileId: string) => void;
+  onError: (error: any) => void;
+  onBlob: (blob: Blob) => void;
```
TamaraFinogina: onBlob is ever used?

src/app/workers/downloadWorker.ts
xabg2: Yep, it is the same.

src/app/drive/services/download.service/createMultipartDownloadStream.ts
xabg2: We don’t really need `<ArrayBufferLike>` here — `Uint8Array` already defaults to `ArrayBufferLike` internally, so it’s redundant. We can remove it if it causes confusion, but this works correctly.

src/app/workers/downloadWorker.ts
xabg2: Yes, it is used when the browser is Brave.

src/app/network/download/v2.ts
xabg2: `params.creds` is not used here explicitly. We use `params.creds` to check that we, in fact, have the credentials. We use `auth.password` (the hashed one) if the credentials are set, if not, we pass an empty stream.

I can refactor this part if it seems confusing.

TamaraFinogina: @xabg2 I see, it's ok

TamaraFinogina (approved):

src/app/network/download/v2.ts
```diff
@@ -88,6 +89,38 @@ const downloadOwnFile: DownloadOwnFileFunction = async (params) => {
...
+    ),
+  );
+
+  const multipartDownload = new MultipartDownload(networkFacade);
+
+  return multipartDownload.downloadFile({
+    bucketId,
+    fileId,
+    mnemonic,
+    fileSize,
```
CandelR, line 116: what happens if the `filesize` is wrong?

src/app/workers/downloadWorker.test.ts
```diff
@@ -27,7 +34,7 @@ describe('Download Worker', () => {
     isWorkspace: false,
     isBrave: false,
     credentials: { user: 'test', pass: 'test' },
-  } as any;
+  } as DownloadFilePayload;
```
CandelR, line 37: 🥳

src/app/network/download/v2.ts
xabg2: We need the file size to split the file in different chunks. If the file size is wrong, then the file will not be download correctly. Any recommendation?

CandelR: We could add a fallback to the normal download, in case there are any files with the wrong file size. We would also need to find a way to verify that the file is not correct, if we had access to the original hash. But anyway, this can be added in another task to improve multipart download management

CandelR (approved):

sg-gs (approved):
I am running a QA with 2 & 2.5GB size files, apart from the typical QA run by @ValentynaBlahodyrQA: 

- [x] Downloading - an uploaded file in production of <2GB - from this version does not corrupt the content. 
- [x] Downloading - an uploaded file in production of >2GB - from this version does not corrupt the content.

PR #1723, opened 2025-11-10
[PB-4435]: Refactor reorganize login module structure and consolidate auth services

terrerox:
## Description

- Move auth services (auth, user, vpnAuth) from src/app/auth/services to src/services
- Relocate login views from src/app/auth/views to src/views/Login
- Move login components from src/app/auth/components to src/views/Login/components
- Relocate shared input components (PasswordInput, TextInput) to src/common/components
- Create barrel exports for src/views/Login and src/views/Login/components
- Update all import paths across the codebase to reflect new structure
- Add error reporting with errorService in useLoginRedirections and ChangePassword
- Remove redundant type annotations (| undefined) from input component interfaces
- Clean up unused files (ActivateTwoFactorAuthSkeleton, auth types, DownloadBackupKey)
- Fix test mocks to use correct service paths

## Checklist

- [ ] QA Passed


terrerox:
Hey @CandelR ! I split this ticket into two more PRs: [#1725](https://github.com/internxt/drive-web/pull/1725) and [#1726](https://github.com/internxt/drive-web/pull/1726). The tests are in the last one. Once the other two get approved, I’ll merge them and fix the conflicts

CandelR (commented):
There are many conflicts and the build is failing. When you resolve them, I will review it

terrerox:
> There are many conflicts and the build is failing. When you resolve them, I will review it

@CandelR  I’ve resolved the conflicts. The build is failing because I still need to merge this PR: https://github.com/internxt/drive-web/pull/1725. I’d prefer to wait for your review on this one first to avoid mixing changes you’ve already reviewed

src/views/Login/UniversalLinkSuccessView.tsx
```diff
@@ -48,7 +52,7 @@
   };
 
   const handleGoToUniversalLinkUrl = () => {
-    window.location.href = getUniversalLinkAuthUrl(user);
+    globalThis.location.href = getUniversalLinkAuthUrl(user);
```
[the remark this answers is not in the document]
terrerox: I think this suggested change might affect the expected behavior of that logic. What do you guys think?

src/app/auth/components/PasswordInput/PasswordInput.tsx
CandelR, line 28: Why have changed the component name to `FormPasswordInput`?

src/app/auth/components/TextInput/TextInput.tsx
CandelR, line 25: Same for this component, why the name change?

src/views/Login/components/SuccessRedirectView.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface SuccessRedirectViewProps {
+  title: string;
+  description: string;
+  redirectSeconds?: number;
+  redirectPath?: string;
+  buttonText: string;
+}
+
+export default function SuccessRedirectView({
```
CandelR, line 15: What is this new view for? Where is it used?

vitest.shared.ts
```diff
@@ -12,6 +12,8 @@ export default {
       use_cases: path.resolve(__dirname, 'src/use_cases'),
       hooks: path.resolve(__dirname, 'src/hooks'),
       views: path.resolve(__dirname, './src/views'),
+      services: path.resolve(__dirname, './src/services'),
+      components: path.resolve(__dirname, './src/components'),
```
CandelR, line 16: Is it necessary to add aliases to the `shared.ts` configuration file if they are already in the `config.ts` file?

CandelR (commented):
These changes could affect login/sign up on mobile app, macOS desktop app, meet web and CLI, so it is mandatory to test them and record the results, either in the Jira task or here

src/views/Login/UniversalLinkSuccessView.tsx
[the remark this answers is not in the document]
CandelR: It should be implemented to avoid security issues, but not in this task, which is already very large. In a separate one with its own task in Jira

src/views/Login/UniversalLinkSuccessView.tsx
```diff
@@ -32,7 +33,10 @@ export default function UniversalLinkView(): JSX.Element {
 
     let baseURL = DEEPLINK_SUCCESS_REDIRECT_BASE;
     if (redirectUri) {
-      baseURL = Buffer.from(redirectUri, 'base64').toString();
+      const decodedUri = Buffer.from(redirectUri, 'base64').toString();
+      if (decodedUri.startsWith(ALLOWED_REDIRECT_SCHEME)) {
```
larryrider: This change will break the current CLI login implementation.

The `redirectUri` parameter is fully controlled by the CLI and must be accepted exactly as it is provided, without filtering or restricting the scheme. In the CLI flow, the redirect URI is not an Internxt deep-link (`internxt://`), but a plain HTTP callback pointing to a local temporary server (e.g., `http://127.0.0.1:45832/callback` or even a different IP if the browser is opened on another machine).

By enforcing a scheme like `internxt://` or checking against `ALLOWED_REDIRECT_SCHEME`, the login flow will fail because the decoded redirect URI will not match that scheme. The CLI relies on being able to dynamically generate a redirect URI that matches the local environment of the user, and this may be any valid URL: a local IP, LAN IP, localhost, or a random port assigned by the OS.

The `internxt://` scheme is only intended for desktop apps, not the CLI. For the CLI, the redirect URI must remain fully flexible so the authentication response can be sent back to the local server the CLI starts during the login process.

For this reason, we need to avoid adding scheme validation here or the CLI login will stop working.

larryrider (changes requested):

src/views/Login/UniversalLinkSuccessView.tsx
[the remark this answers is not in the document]
larryrider: The CodeQL/Copilot suggestion for the decodedUri param here is a false positive for our specific use case.

In our CLI login architecture, we must continue allowing arbitrary values such as:
- `http://localhost:<port>`
- `http://127.0.0.1:<port>`
- any local/private IP address
- any port dynamically assigned by the OS

Because this is exactly how the CLI login flow works.

The only "allowed" pattern is essentially "any HTTP callback pointing to the user's own machine". We cannot restrict the scheme to `internxt://` (which is only for desktop apps), nor can we enforce strict equality or a fixed allowlist without completely breaking the CLI.

So even though CodeQL correctly detects that the parameter is externally modifiable, the suggested fix does not apply to our use case. There is simply no viable secure mitigation based on strict filtering, because the dynamic redirect is a fundamental part of how the CLI flow functions.

For these reasons, this is effectively a false positive in our context, and the change should be reverted. We should keep the current behavior and perhaps add a code comment explaining why redirect filtering cannot be tightened in the CLI flow right now.

src/app/auth/components/PasswordInput/PasswordInput.tsx
terrerox: Modified because two components had the same name despite containing different content; the old component has now been removed, leaving only one in use

vitest.shared.ts
terrerox: Yes, because the workspace test projects extend from vitest.shared.ts directly, not from vitest.config.ts. This means they don't inherit the aliases defined in the main config

terrerox:
> These changes could affect login/sign up on mobile app, macOS desktop app, meet web and CLI, so it is mandatory to test them and record the results, either in the Jira task or here

Since i reverted the logic that would affect the CLI login implementation. These are the only test scenarios:
https://github.com/user-attachments/assets/0f160497-6095-4927-b506-95c0b8cf4441

larryrider (approved):
IT looks good to me but please, **don't merge** this PR until tests and build works, and QA passes

src/views/Login/components/SuccessRedirectView.tsx
terrerox: It’s used on the ChangePassword and RestartAccount pages. It’s the component that shows a countdown and then redirects

PR #1724, opened 2025-11-11
[PB-5095]: feat/add stream saver library manually

xabg2:
## Description

This PR implements the StreamSaver library locally so that we can maintain our own version and ensure compatibility with modern browser features, especially for Safari.

## Checklist

- [ ] QA Passed


xabg2:
Hey @CandelR , take a look to this whenever you can! 😉

CandelR (commented):
Have you considered using the [vite-pwa plugin](https://vite-pwa-org.netlify.app/guide/register-service-worker.html)?  
I mention this because it will help us to structure this better, with TypeScript instead of vanilla JavaScript, and we will save part of the code implemented in this PR for the use of the service worker (probably what is in public/streamsaver would not be necessary, nor would the iframe trick).
I would need to review it in more depth as I have never used it, but I mention it because if there is time, it seems like a good option that would make this simpler and easier to maintain :)

public/streamsaver/mitm.html
```diff
@@ -0,0 +1,146 @@
...
+        if (data.streamSaverVersion === '1.2.0') {
+          console.warn('[StreamSaver] Please update StreamSaver');
+        }
+
+        if (!data.headers) {
+          console.warn(
+            "[StreamSaver] Pass `data.headers` that you would like to pass along to the service worker\nIt should be a 2D array or a key/val object that fetch's Headers API accepts",
+          );
+        } else {
+          new Headers(data.headers);
```
larryrider: I think this new Headers code is not used

larryrider (commented):
check sonarcloud issues

larryrider (approved):

PR #1727, opened 2025-11-12
[PB-5116] feature/Add login for meet

CandelR:
## Description

- Added new url for authetincation from other intenrxt apps using postMessage for secure communication between popup and parent window

## Checklist

- [x] QA Passed

## Testing Process

- Check that other login/signup methods works correctly (from desktop, mobile, web)


src/app/auth/hooks/useOAuthFlow.test.tsx
```diff
@@ -0,0 +1,163 @@
...
+  registerCompleted: true,
+  hasReferralsProgram: true,
+  createdAt: new Date('2023-06-01T12:00:00.000Z'),
+  avatar: null,
+  emailVerified: true,
+};
+
+vi.mock('../services/oauth.service');
+
+import * as oauthService from '../services/oauth.service';
```
larryrider: this import is after a mock and a const declarations, thats a bit rare, right?

src/app/auth/hooks/useOAuthFlow.test.tsx
xabg2, line 1: I’d suggest making the test descriptions less technical by avoiding the use of variable names. Writing tests in a more descriptive, non-technical way makes them easier to understand.
Also, if variable names change or are removed in the future, it can lead to confusion.

src/views/Signup/components/SignupForm.tsx
```diff
@@ -204,6 +214,20 @@ function SignUpForm(): JSX.Element {
 
     handleVPNAuth(isVPNAuth, xNewToken);
 
+    if (isOAuthFlow) {
+      const user = localStorageService.getUser();
+      if (user) {
+        const success = handleOAuthSuccess(user, xToken, xNewToken);
```
xabg2: if `xToken` is not longer used (I think we just use `xNewToken` as far as I know), maybe we can remove it.

larryrider: @xabg2 I wrote the same but I have removed my comment because:
- this pr is on draft
- the xToken removal is already planned for drive-web

src/app/auth/hooks/useOAuthFlow.test.tsx
CandelR: As a personal preference, I don't mind having variable names in the test description, because if the variable changes, the test will also change if that is what is being tested. 
I'm not entirely convinced by the reason for the change because it's not something that changes very often, and looking at the test description, I already know what it tests. However, if it's all natural language and not technical, I don't need to know which part of the code is being tested, I have to go and look at the test no matter what. 
Give me some examples with this test file and we'll see how it looks, whether it's easier to understand or not

src/views/Signup/components/SignupForm.tsx
CandelR: I'll remove it from here, guys, as I hadn't considered that it's no longer in use. Good catch :)

src/app/auth/hooks/useOAuthFlow.test.tsx
xabg2: I’d suggest structuring the tests based on behavior, rather than variables or internal code details.
For example, you could start with:
`describe('OAuth custom hook', ...)`
Then:
```
describe('Auth origin', ...)
it('When an auth origin is provided, then the auth process begins', ...) // You can base this on "on component mount" to verify `isOauthFlow` and the purpose of that variable
it('When an auth origin is not provided, then nothing happens', ...)
```
And later:
```
describe('Handling successful OAuth', ...)
it('When the OAuth process completes successfully, then <expected behaviour>', ...)
it('When the OAuth process fails or is not completed, then <expected behaviour>', ...)
```
This way, tests are easier to understand and focused on the expected outcomes rather than the implementation details.

CandelR: Okay, I'll add it in that way

xabg2 (approved):

src/app/auth/components/LogIn/LogIn.tsx
```diff
@@ -147,14 +152,28 @@ export default function LogIn(): JSX.Element {
   };
 
   const handleSuccessfulAuth = (token: string, user: UserSettings, mnemonic: string): void => {
+    const newToken = localStorageService.get('xNewToken');
+
+    if (newToken) {
```
larryrider: Wouldn't it better to check here if (newToken && isOauthFlow) instead of checking isOauthFlow after calling to `handleOAuthSuccess` ?

That way we are always calling `handleOAuthSuccess` even if we are not in the oauth flow 🤔

CandelR: Wow, yes, at first I added the check in the function itself, and when I made changes, I forgot to rectify this

larryrider (approved):

PR #1728, opened 2025-11-13
[MKT-648]: update/features for plans

jaaaaavier:
## Description
In this PR we update the product features text for each plan, changes:
- Update JSON text of productFeatures in every language according to the features every plan needs.

The feature descriptions were outdated and didn't reflect the current plan capabilities.

## Checklist

- [ ] QA Passed

## Additional Notes
- Only text content was changed; JSON structure remains unchanged
- No code modifications were required, minimizing risk of bugs
- Maybe in the future we can make a general features or something similar if you want, we are pending to also update the business features, then we can change


xabg2 (approved):
Check carefully that this does not breaks either the product cards or the checkout.

jaaaaavier:
> Check carefully that this does not breaks either the product cards or the checkout.

Yes, I have tested it the way we usually do, by running both projects locally and linking them with the redirection. No issues found. Also tested that the panel doesn't generate any problems in any language.

PR #1730, opened 2025-11-14
[PB-4444]: Add unit tests for Checkout services and store reducer

terrerox:
## Description

  Added comprehensive test coverage for:
  - authCheckoutService payment methods and subscription handling
  - paymentService checkout sessions, subscriptions, and trial creation
  - checkoutReducer state management for coupons, seats, and currency

## Checklist

- [ ] QA Passed

### DIFF
--- a/src/views/Checkout/store/checkoutReducer.test.ts
+++ b/src/views/Checkout/store/checkoutReducer.test.ts
@@ -1,0 +1,183 @@
+import { describe, it, expect } from 'vitest';
+import { checkoutReducer, initialStateForCheckout } from './checkoutReducer';
+import { State } from './types';
+import { PriceWithTax } from '@internxt/sdk/dist/payments/types';
+import { DisplayPrice, UserType } from '@internxt/sdk/dist/drive/payments/types/types';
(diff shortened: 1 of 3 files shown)


src/views/Checkout/store/checkoutReducer.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  describe('SET_CURRENT_PLAN_SELECTED', () => {
+    it('updates currently viewing plan', () => {
+      const mockPlan = createMockPlan('456', 'eur', 1999, UserType.Business);
+      const result = checkoutReducer(initialStateForCheckout, { type: 'SET_CURRENT_PLAN_SELECTED', payload: mockPlan });
+
+      expect(result.currentSelectedPlan).toEqual(mockPlan);
+    });
+  });
```
CandelR: Hey @terrerox, check these tests. There's no need to reach 100% coverage only adding tests without good checks. 
For example, these tests only check that the state changes, but they should also check that the rest of the state is maintained, because otherwise we could make each action delete the entire reducer and none of these tests would return an error

CandelR (approved):

PR #1750, opened 2025-11-25
[MKT-664]:feat/Meta Pixel Implementation for Lead and Purchase

jaaaaavier:
## Description
This PR implements Meta Ads tracking microstrategy to capture key business conversions with the highest possible data quality.

MetaPixelService is introduced to centralize tracking logic, and ensure all events send high-quality user parameters necessary.

Functionality implemented:
- Lead Event: Fires upon successful user registration.
- Purchase Event: Fires upon successful payment confirmation.

Also we removed one GTM container that we are not longer using

## Checklist

- [ ] QA Passed

## Testing Process
Following the environment setup and ensuring all necessary configurations were complete in GTM, we proceeded to use GTM’s native Preview Mode for validation. We confirmed that the custom-built tags fire as intended: the Lead tag fires correctly upon user registration, and the Purchase tag fires successfully after a completed sale. Furthermore, both tags accurately receive all the necessary enriched values and parameters. We can therefore confirm the implementation is fully functional. Should visual proof be required, please let me know.

## Additional Notes

I will also test using the specific Preview URL  from this PR to ensure all environment variables are correctly being picked up by the application.

### DIFF
--- a/src/app/analytics/meta.service.test.ts
+++ b/src/app/analytics/meta.service.test.ts
@@ -1,0 +1,176 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { trackLead, trackPurchase } from './meta.service';
+import localStorageService from 'app/core/services/local-storage.service';
+
+vi.mock('app/core/services/local-storage.service', () => ({
+  default: {
+    get: vi.fn(),
+  },
+}));
+
+describe('Meta Tracking Service', () => {
+  let mockDataLayer: any[];
+  let mockFbq: vi.Mock;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    mockDataLayer = [];
+    mockFbq = vi.fn();
+
+    Object.defineProperty(window, 'dataLayer', {
+      value: mockDataLayer,
+      writable: true,
+      configurable: true,
+    });
+
+    Object.defineProperty(window, 'fbq', {
+      value: mockFbq,
+      writable: true,
+      configurable: true,
+    });
+  });
+
+  describe('trackLead', () => {
+    it('When valid data is provided, then lead event is pushed to dataLayer', () => {
+      const email = 'test@example.com';
+      const userID = 'user123';
+
+      trackLead(email, userID);
+
+      expect(mockDataLayer).toHaveLength(1);
+      expect(mockDataLayer[0]).toMatchObject({
+        event: 'leadSuccessful',
+        eventCategory: 'User',
+        eventAction: 'registration_complete',
+        userEmail: email,
+        userID: userID,
+      });
+    });
+
+    it('When dataLayer is not available, then no event is pushed', () => {
+      Object.defineProperty(window, 'dataLayer', {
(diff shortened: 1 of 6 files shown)


src/app/analytics/meta.service.test.ts
```diff
@@ -0,0 +1,176 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { trackLead, trackPurchase } from './meta.service';
+import localStorageService from 'app/core/services/local-storage.service';
+
+vi.mock('app/core/services/local-storage.service', () => ({
```
xabg2: instead of using mock, use spyOn.

src/app/analytics/meta.service.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      const initialLength = mockDataLayer.length;
+      trackLead('test@example.com', 'user123');
+
+      expect(mockDataLayer).toHaveLength(initialLength);
+    });
+  });
+
+  describe('trackPurchase', () => {
+    beforeEach(() => {
```
xabg2, line 73: Same, use spyOn in each case you need to mock the returned value.

src/app/analytics/meta.service.ts
```diff
@@ -0,0 +1,44 @@
+import localStorageService from 'app/core/services/local-storage.service';
+
+const canTrack = () => {
```
xabg2, line 3: It can be directly a variable instead of a function, isn't it? Something like `const isTrackAllowed = typeof window !== 'undefined' && (window as any).dataLayer && (window as any).fbq;` or there is any issue with that?

src/app/analytics/meta.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  return typeof window !== 'undefined' && (window as any).dataLayer && (window as any).fbq;
+};
+
+export const trackLead = (email: string, userID: string) => {
+  if (!canTrack()) {
+    return;
+  }
+
+  (window as any).dataLayer.push({
```
xabg2: Define `dataLayer` in [this file](https://github.com/internxt/drive-web/blob/master/src/react-app-env.d.ts#L25) so you can use it without `(window as any)`

src/app/analytics/meta.service.ts
jaaaaavier: We need a function because we call the tracking in runtime, after the external scripts GTM and Facebook Pixel have loaded, If we use a variable, it's evaluated once when the module loads, before the tracking scripts exist, so it stays false. I investigated what we could do if we use a variable, and it's complicated to manage because we wouldn't know when to update it because the scripts are external.

Maybe you have other ideas?? I think this is the easiest one

src/app/analytics/meta.service.ts
jaaaaavier: Done

xabg2:
Check the sonar issues too @jaaaaavier

xabg2 (approved):

PR #1753, opened 2025-11-26
[_]: chore/remove Sentry form the project

xabg2:
## Description

Removing Sentry from the project as it is no longer used anymore.

## Checklist

- [ ] QA Passed


src/services/error.service.ts
```diff
@@ -20,15 +17,8 @@ const errorService = {
...
-    if (envService.getVariable('nodeEnv') === 'development') {
-      console.error('[ERROR_CATCHED]: This error has been catched and is being reported to Sentry', exception);
-    }
-    captureException(exception, context);
-  },
-
-  addBreadcrumb(breadcrumbProps: Breadcrumb): void {
-    addBreadcrumb(breadcrumbProps);
+  reportError(exception: unknown): void {
+    console.error('[ERROR]: There was an error: ', exception);
```
larryrider, line 24: the console.error was only happening on development, maybe we should keep that condition 🤔

xabg2, line 24: I left this function in case we want to add Sentry again, so we can implement it easier.

sg-gs (approved):

src/app/network/DownloadManager.ts
larryrider: remove this "orphan" comments too

src/app/network/UploadManager.ts
```diff
@@ -320,19 +307,6 @@ class UploadManager {
   }) {
     const file = fileData.filecontent;
```
larryrider: it seems that this file is not being used neither

src/services/error.service.ts
xabg2: Good idea. I'll keep that condition.

larryrider (approved):

PR #1761, opened 2025-12-03
[PB-5409]: feature/video streaming

xabg2:
## Description

Implementing the video streaming so the users can preview the videos without downloading them before.

## Checklist

- [ ] QA Passed


src/utils/initializeServiceWorkers.utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      await navigator.serviceWorker.ready;
+      console.log('[ServiceWorkers] All service workers ready');
+
+      console.log('[ServiceWorkers] All workers initialized successfully');
+    } catch (error) {
+      console.error('[ServiceWorkers] Failed to initialize service workers:', error);
+    }
+  } else {
+    console.warn('[ServiceWorkers] Service Workers not supported in this browser');
```
larryrider, line 24: Do we have a fallback for when service workers are not available? 
If not, maybe it would be a good idea to create a future task to allow blob preview as before if its not available. If it works right now like that, then i have not said anything 👀

src/app/drive/types/file-types.ts
```diff
@@ -218,7 +218,7 @@ export const videoExtensions: FileExtensionMap = {
   flv: ['flv', 'f4v', 'f4p', 'f4a', 'f4b'],
 };
 
-const previewableVideoExtensionsGroup: string[] = ['webm', 'mkv', 'quicktime', 'mpeg-4', '3gpp'];
+const previewableVideoExtensionsGroup: string[] = Object.keys(videoExtensions);
```
larryrider, line 221: Is this working good? All video extensions are supported and working good? 😲 🕵️

xabg2: I've tested some of them and they worked well (less avi for example). But as we only fetch the first chunk and some extensions are supported by the browser, then we will try to preview it and, if the browser says no, then we will update the viewer state accordingly.

src/utils/initializeServiceWorkers.utils.ts
xabg2: Hmm, no, we don't have an alternative, but yes, we should add an alternative and use the current preview of the blob with file size limitations. We can do it in another task.

larryrider (approved):
Good job!! 🚀

PR #1771, opened 2025-12-09
  [PB-5465] Implement file version history actions

terrerox:
## Description

  - Add version download, delete, and restore functionality
  - Implement file version service with API integration
  - Add version action dialogs for delete and restore confirmations
  - Add loading skeleton component for version history sidebar
  - Replace mock data with real API integration
  - Add UI state management for version dialogs
  - Pass userName prop to version components to avoid duplication
  - Update name collision dialog to use replaceFile endpoint instead of delete-and-reupload
  - Add replace-file service for efficient file replacement
  - Upgrade @internxt/sdk from 1.11.17 to 1.11.18
  
## Checklist

- [ ] QA Passed


src/app/i18n/locales/en.json
```diff
@@ -892,7 +892,20 @@
       "autosaveVersions": "{{count}}/{{total}} autosave versions",
       "restoreVersion": "Restore version",
       "downloadVersion": "Download version",
-      "deleteVersion": "Delete version"
+      "deleteVersion": "Delete version",
```
terrerox, line 895: The remaining translations are included in this PR: https://github.com/internxt/drive-web/pull/1773

src/views/Drive/components/VersionHistory/services/file-version.service.ts
```diff
@@ -0,0 +1,61 @@
...
+  version: FileVersion,
+  fileName: string,
+  bucketId: string,
+  updateProgressCallback?: (progress: number) => void,
+  abortController?: AbortController,
+): Promise<void> {
+  const user = localStorageService.getUser();
+  if (!user) throw new Error('User not found');
+
+  const fileStream = await downloadFile({
```
xabg2: Can we use the multipart download for this? Imagine the file has 5GB or more. It would be better to download it in parallel-small chunks instead of the full file at once. You can do it by using the DownloadManager.downloadItem if I'm not mistaken.

Check that as we need to download the items using the actual method since it solves a lot of speed/memory issues.

src/views/Drive/components/VersionHistory/services/fileVersion.service.ts
xabg2, line 1: Remember to add tests to this service.

src/views/Drive/components/VersionHistory/services/file-version.service.ts
```diff
@@ -0,0 +1,61 @@
...
+      notifyProgress: (totalBytes, downloadedBytes) => {
+        if (updateProgressCallback) {
+          updateProgressCallback(downloadedBytes / totalBytes);
+        }
+      },
+      abortController,
+    },
+  });
+
+  const blob = await binaryStreamToBlob(fileStream);
```
xabg2: Downloading it as blob and then save it will causes memory issues, specially for Firefox. Also, it won't work for big files in Safari - the file will be downloaded but only 700 bytes (or even 0) will be saved to the disk as Safari does not allow saving big blobs. 

Try the method I mentioned above as it already handles all these things, such as saving the downloaded + decrypted chunk directly to the user disk instead of saving it in memory

src/views/Drive/services/replaceFile.service.ts
xabg2, line 1: Why is this function in this service? It is a part of file versioning, isn't it? Also, remember to add tests to this service.

xabg2 (changes requested):

src/views/Drive/services/replaceFile.service.ts
terrerox: This service is used in the name-collision modal, which belongs to Drive and isn’t necessarily part of the file-version history flow. That’s why I placed it inside the Drive module. It could live under Version History as well, but keeping it in Drive makes more sense in this context. What do you think?

The tests will be added in a separate PR.

xabg2 (approved):

yarn.lock
```diff
@@ -1906,10 +1906,10 @@
   version "1.0.2"
   resolved "https://codeload.github.com/internxt/prettier-config/tar.gz/9fa74e9a2805e1538b50c3809324f1c9d0f3e4f9"
 
-"@internxt/sdk@=1.11.17":
-  version "1.11.17"
-  resolved "https://registry.yarnpkg.com/@internxt/sdk/-/sdk-1.11.17.tgz#2f5bdada5d3cbf5cfc685a21c24b5df3ff51d8c8"
-  integrity sha512-91iEUvZizlwX6KBEFJ3JdFiGrhMBQ9R54sTc3Pei9QtV2FYTU8nTVEPYAg39tLOGzT/kVuplYOtBxfk6wFtSDA==
+"@internxt/sdk@1.11.19":
+  version "1.11.19"
+  resolved "https://npm.pkg.github.com/download/@internxt/sdk/1.11.19/f9033c8bf0849d414a3865806bacd79026a38769#f9033c8bf0849d414a3865806bacd79026a38769"
```
larryrider: This change is breaking your cloudflare build, and it also breaking the PR's that depend on this one. You are using an .npmrc file so you are changing the internxt registry from yarnpkg to github.

You should use this one:

`registry=https://registry.yarnpkg.com/`
`@internxt:registry=https://registry.yarnpkg.com/`

larryrider (changes requested):

yarn.lock
terrerox: Thanks! Since this is an old PR, I made the change in this one instead: https://github.com/internxt/drive-web/pull/1779

src/app/core/layouts/HeaderAndSidenavLayout/HeaderAndSidenavLayout.tsx
```diff
@@ -58,7 +58,9 @@ export default function HeaderAndSidenavLayout(props: HeaderAndSidenavLayoutProp
 
             <VersionHistorySidebar />
           </div>
-          <TaskLogger />
+          <div className="absolute bottom-0 right-0 z-50 w-80">
```
CandelR, line 61: this div wrapping the task logger in necessary? I think could cause wrong behaviours

src/app/drive/components/NameCollisionDialog/NameCollisionContainer.tsx
```diff
@@ -146,19 +149,14 @@ const NameCollisionContainer: FC<NameCollisionContainerProps> = ({
...
-            options: {
-              disableDuplicatedNamesCheck: true,
-            },
-          }),
-        ).then(() => {
-          dispatch(fetchSortedFolderContentThunk(folderId));
+        const file = itemToUpload as File;
+        await replaceFileService.replaceFile(itemToReplace.uuid, {
+          fileId: itemToReplace.fileId,
+          size: file.size,
```
CandelR: I don't quite understand this change. The file data is replaced, but when is the new one uploaded?
Maybe is in another PR?d

src/views/Drive/components/VersionHistory/hooks/useVersionItemActions.ts
```diff
@@ -40,7 +65,7 @@ export const useVersionItemActions = ({ version, onDropdownClose }: UseVersionIt
     {
       name: translate('modals.versionHistory.deleteVersion'),
       icon: Trash,
-      action: handleDelete,
+      action: handleDeleteClick,
```
CandelR, line 68: move the hooks and services outside the components folder, add it to Drive services and hooks. 
Otherwise, it would be crazy to maintain this. The components would be full of hooks and services, and components should not have business logic or API services

src/app/drive/components/NameCollisionDialog/NameCollisionContainer.tsx
terrerox: The backend has that upload new version logic

src/app/core/layouts/HeaderAndSidenavLayout/HeaderAndSidenavLayout.tsx
terrerox: When downloading a version, the task logger was located under the File Versions sidebar, making it difficult to cancel the download.

src/views/Drive/components/VersionHistory/hooks/useVersionItemActions.ts
terrerox: To avoid conflicts, im going to merge all these approved PRs and then work this comment in a new [PR](https://github.com/internxt/drive-web/pull/1808)

PR #1772, opened 2025-12-09
[PB-5525] Remove getAesInitFromEnv and its react variables

TamaraFinogina:
## Description

This PR removes magic iv and salt as everything can work perfectly without them. 

## Checklist

- [ ] QA Passed

## Testing Process

Unit tests


src/app/crypto/services/keys.service.ts
```diff
@@ -3,14 +3,13 @@ import { Keys } from '@internxt/sdk';
 import { getOpenpgp, generateNewKeys, compareKeyPairIDs } from './pgp.service';
 import { isValid } from './utilspgp';
 import { UserSettings } from '@internxt/sdk/dist/shared/types/userSettings';
-import envService from 'services/env.service';
 const MINIMAL_ENCRYPTED_KEY_LEN = 129;
 
 export async function getKeys(password: string): Promise<Keys> {
   const { privateKeyArmored, publicKeyArmored, revocationCertificate, publicKyberKeyBase64, privateKyberKeyBase64 } =
     await generateNewKeys();
-  const encPrivateKey = aes.encrypt(privateKeyArmored, password, getAesInitFromEnv());
```
xabg2, line 12: I guess that `getAesInitFromEnv` is an optional parameter, right? @TamaraFinogina

xabg2 (approved):

src/app/crypto/services/keys.service.ts
TamaraFinogina: yes, if salt and iv are not given, they are sampled at random (better than the same for all). And since it's not secret and stored as part of the ciphertext, nothing will change:
`return Buffer.concat([salt, iv, tag, encrypted]).toString('base64');`

sg-gs (approved):
Nice removal @TamaraFinogina. Be aware of considering the login after the signup with this method during the QA as a malformation here affects sharing and the entire b2b workflow.

PR #1777, opened 2025-12-15
[PB-4447]: refactor/create custom hooks to fetch products and initialize checkout

xabg2:
## Description

New custom hooks to initialize the checkout and fetch the products (price and coupon code).

## Checklist

- [ ] QA Passed

## Additional Notes

There are more changes to do. They will be done in another PR.


src/views/Checkout/hooks/useInitializeCheckout.ts
```diff
@@ -0,0 +1,124 @@
...
+  useEffect(() => {
+    if (stripeSdk && price) {
+      loadStripeAndCrypto();
+    }
+  }, [stripeSdk, price?.price?.id]);
+
+  const initCheckout = async () => {
+    try {
+      await initializeStripe();
+      await fetchUserLocationAndStore();
```
larryrider: Maybe could be a good idea to wrap this two promises in a Promise.all so they are executed on parallel. I know that they are only two but maybe for future versions is better

larryrider: just only if they are independent from each other (as it seems to be)

src/views/Checkout/hooks/useInitializeCheckout.ts
```diff
@@ -0,0 +1,124 @@
...
+    } catch {
+      redirectToFallbackPage();
+    }
+  };
+
+  const loadStripeAndCrypto = async () => {
+    try {
+      await loadStripeData();
+      await loadCryptoCurrencies();
```
larryrider: same here

src/views/Checkout/hooks/useInitializeCheckout.ts
xabg2: Yep, they are. They are loading different things

larryrider (approved):

CandelR (approved):

PR #1792, opened 2025-12-24
[PB-5239]: add confirmation dialog for deleting backups

terrerox:
## Description

Introduces a confirmation dialog before deleting items in the backups view to prevent accidental deletions. Refactors delete logic to handle both file viewer and list contexts.

## Checklist

- [ ] QA Passed


src/views/Backups/BackupsView.tsx
```diff
@@ -80,46 +83,84 @@ export default function BackupsView(): JSX.Element {
...
         text: translate('notificationMessages.errorDeletingItems'),
         type: ToastType.Error,
       });
+      return false;
+    } finally {
+      setIsDeletingItems(false);
     }
   };
 
+  const onDeleteSelectedItems = async () => {
```
larryrider, line 137: I think this function could be a little "weird" with the return inside the if 

Maybe its better this way wdyt?:

if (isFileViewerOpen) {
      await performDeleteItems(selectedItems);
} else {
    openDeleteItemsDialog(selectedItems);
}

src/views/Backups/BackupsView.tsx
```diff
@@ -80,46 +83,83 @@ export default function BackupsView(): JSX.Element {
     }
   };
 
-  const onDeleteSelectedItems = async () => {
-    const selectedItemsIDs = new Set(selectedItems.map((item) => item.id));
-    const filteredCurrentItems = currentItems.filter((item) => !selectedItemsIDs.has(item.id));
+  const openDeleteItemsDialog = (items: DriveItemData[]) => {
+    if (!items.length || isFileViewerOpen) {
```
CandelR: why if the file viewer is opened we skip the dialog and we delete the items directly? 
I think that we should it display always

CandelR (approved):

PR #1797, opened 2026-01-02
[MKT-701]:feat:Add first upload condition to FeaturesBanner display

jaaaaavier:
## Description
Added logic to show the FeaturesBanner only after the user has uploaded their first file and 90 seconds. This ensures new users see the banner at the appropriate time.

## Checklist

- [ ] QA Passed

## Testing Process
Created a new account, and uploaded File, then checked the banner displays after 90 seconds, uploaded a video on the Jira task


src/app/banners/BannerWrapper.tsx
```diff
@@ -15,17 +15,26 @@ const BannerWrapper = (): JSX.Element => {
...
 
   useEffect(() => {
-    const timeout = setTimeout(() => setShowDelayedBanner(true), TIMEOUT);
-    return () => clearTimeout(timeout);
+    newStorageService.hasUploadedFiles().then(({ hasUploadedFiles }) => {
+      if (hasUploadedFiles) {
+        setHasUploadedFirstFile(true);
+      }
+    });
   }, []);
```
CandelR, line 29: this logic only checks the upload status when mount hte banner, are this the behaviour that you want?
Add the related tests :)

jaaaaavier: Yes, that's the intended behavior. I've used the same approach we have for the tutorial. I've tested it and confirmed that the banner doesn't appear after 90 seconds just from being on the homepage, and that it correctly appears 90 seconds after the first file upload. Adding the tests now.

larryrider (approved):

CandelR (approved):

PR #1803, opened 2026-01-08
[MKT-690]: feat/implement incomplete checkout tracking logic

jaaaaavier:
## Description

This PR implements the logic to track incomplete checkouts. It adds a new trackIncompleteCheckout method to the checkoutService and integrates it into the checkout view.

Key changes:
Updated SDK version

Service: Added trackIncompleteCheckout to checkoutService, which constructs the checkout URL and calls the users SDK.

View: Integrated the tracking call within a useEffect in the checkout component.

Logic: Implemented a useRef (hasTrackedRef) to ensure the tracking event is triggered only once per session/render to prevent duplicate events.

Tests: Added unit tests for the new service method to verify URL construction and error handling.

## Checklist

- [ ] QA Passed

## Additional Notes
Video test on the [task](https://inxt.atlassian.net/jira/software/c/projects/MKT/boards/81?selectedIssue=MKT-690)

### DIFF
--- a/src/views/Checkout/services/checkout.service.ts
+++ b/src/views/Checkout/services/checkout.service.ts
@@ -246,101 +246,27 @@ const loadStripeElements = async (
 };
 
 const trackIncompleteCheckout = async (selectedPlan: PriceWithTax | undefined, price?: number): Promise<void> => {
-  console.log('[trackIncompleteCheckout] Starting...', {
-    selectedPlan,
-    price,
-    hasPriceId: !!selectedPlan?.price?.id,
-  });
-
   try {
-    // Validación 1: selectedPlan.price.id
-    if (!selectedPlan?.price?.id) {
-      console.warn('[trackIncompleteCheckout] Early return: missing selectedPlan.price.id', {
-        selectedPlan,
-      });
-      return;
-    }
-
-    console.log('[trackIncompleteCheckout] selectedPlan.price.id exists:', selectedPlan.price.id);
-
-    // Validación 2: bytes e interval
-    if (!selectedPlan.price.bytes) {
-      console.warn('[trackIncompleteCheckout] Missing selectedPlan.price.bytes', {
-        bytes: selectedPlan.price.bytes,
-      });
-    }
-
-    if (!selectedPlan.price.interval) {
-      console.warn('[trackIncompleteCheckout] Missing selectedPlan.price.interval', {
-        interval: selectedPlan.price.interval,
-      });
-    }
-
-    // Validación 3: price
-    if (price === undefined || price === null) {
-      console.warn('[trackIncompleteCheckout] price is undefined or null', { price });
-    }
+    if (!selectedPlan?.price?.id) return;
 
     const priceId = selectedPlan.price.id;
-
-    console.log('[trackIncompleteCheckout] Converting bytes to string...', {
-      bytes: selectedPlan.price.bytes,
-    });
-
-    const bytesStr = bytesToString(selectedPlan.price.bytes);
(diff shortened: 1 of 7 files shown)


src/views/Checkout/views/CheckoutViewWrapper.tsx
```diff
@@ -201,6 +203,14 @@ const CheckoutViewWrapper = () => {
     }
   }, [isCheckoutReady]);
 
+  useEffect(() => {
+    if (selectedPlan?.price && isAuthenticated && !hasTrackedRef.current) {
+      hasTrackedRef.current = true;
+      const planPrice = selectedPlan.taxes?.amountWithTax || selectedPlan.price.amount;
+      checkoutService.trackIncompleteCheckout(selectedPlan, planPrice);
+    }
+  }, [selectedPlan, isAuthenticated]);
```
xabg2, line 212: How does it work internally? Because we send the track as soon as the page loads, not when it leaves for example. How do we know that the user hasn't purchased the plan?

Also, consider adding a check when it comes to the development environment so that the tracks are not sent.

jaaaaavier: We use a Klaviyo metric called 'IncompleteCheckout.' Since tracking only those who exit was difficult, I used Klaviyo's filters and our Stripe integration to target users who haven't paid. The flow works as follows: 
1. It triggers when a checkout starts. 
2. It waits one hour to check for a payment. 
3. If no purchase is detected after that hour, it sends the recovery email.

src/views/Checkout/services/checkout.service.ts
```diff
@@ -244,6 +245,28 @@ const loadStripeElements = async (
...
 
+const trackIncompleteCheckout = async (selectedPlan: PriceWithTax | undefined, price?: number): Promise<void> => {
+  try {
+    if (!selectedPlan?.price?.id) return;
+
+    const priceId = selectedPlan.price.id;
+    const planName = bytesToString(selectedPlan.price.bytes) + ' ' + selectedPlan.price.interval;
+    const currentParams = new URLSearchParams(globalThis.location.search);
+    currentParams.set('planId', priceId);
+    const completeCheckoutUrl = `https://drive.internxt.com/checkout?${currentParams.toString()}`;
```
xabg2: Do not hardcode the hostname. You can use [this](https://github.com/internxt/drive-web/blob/master/src/services/env.service.ts#L29) variable to get the host. You can use the function `getVariable(...)`.

src/views/Checkout/services/checkout.service.ts
```diff
@@ -244,6 +245,28 @@ const loadStripeElements = async (
...
+
+    const usersClient = SdkFactory.getNewApiInstance().createUsersClient();
+    await usersClient.handleIncompleteCheckout({
+      completeCheckoutUrl,
+      planName,
+      price,
+    });
```
xabg2: Extract these functions [here](https://github.com/internxt/drive-web/blob/master/src/services/user.service.ts).

src/views/Checkout/services/checkout.service.ts
```diff
@@ -244,6 +245,28 @@ const loadStripeElements = async (
...
+
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
xabg2: You're sure this is needed?

xabg2 (changes requested):

src/views/Checkout/services/checkout.service.ts
jaaaaavier: For now removed, we will check if it works on production

xabg2 (approved):

PR #1805, opened 2026-01-08
[PB-5655] feat(recovery): send backup public keys for validation during account

douglas-xt:
## Description

This PR adds support for sending public keys during account recovery. When users export their backup file, it now includes both private and public keys. During account recovery, these public keys are extracted from the backup file and forwarded to the backend, which validates that they match the account being recovered.

## Checklist

- [ ] QA Passed

## Additional Notes

Updated @internxt/sdk from v1.11.17 to v1.12.0. This version includes a breaking change in changePasswordWithLinkV2 the keys parameter changed from PrivateKeys to RecoveryKeys ({ private?, public? }), which allows sending public keys alongside private keys during recovery.


larryrider (dismissed):

src/utils/backupKeyUtils.ts
```diff
@@ -49,6 +56,10 @@ export function handleExportBackupKey(translate) {
         ecc: user.keys?.ecc?.privateKey || user.privateKey,
         kyber: user.keys?.kyber?.privateKey || '',
       },
+      publicKeys: {
+        ecc: user.keys?.ecc?.publicKey || '',
+        kyber: user.keys?.kyber?.publicKey || '',
```
CandelR: I see that if for some reason the public keys are not there, empty strings will be exported. 
When attempting to send this empty string, do we notify the user?
Shouldn't we notify them if empty strings are going to be exported?

src/services/auth.service.test.ts
```diff
@@ -687,7 +687,7 @@ describe('updateCredentialsWithToken', () => {
     expect(keys).toBeUndefined();
   });
 
-  it('should successfully update credentials with token and with backup data (ECC only)', async () => {
+  it('should not send keys when backup data has no publicKeys (legacy backup)', async () => {
```
CandelR: if doesn't mind, change the test that are modifying to "when X, then Y" format

src/utils/backupKeyUtils.ts
douglas-xt: If both ecc and kyber public keys are not present, publicKeys shouldn't be exported at all. In order to keep backward compatibility, I've changed that.

CandelR (dismissed):

xabg2 (approved):

PR #1807, opened 2026-01-09
[PB-1053]: fix/prevent default extension for files that does not have extension

xabg2:
## Description

Fixed an issue where browsers were adding a random file extension when downloading files without an extension. For StreamSaver-based downloads, the `DefaultType: none` header is now used, and for Blob-based downloads, the Blob is created with type: `application/octet-stream`.

## Checklist

- [x] QA Passed

### DIFF
--- a/public/streamsaver/stream-saver.js
+++ b/public/streamsaver/stream-saver.js
@@ -80,7 +80,6 @@ self.onfetch = (event) => {
 
   const responseHeaders = new Headers({
     'Content-Type': 'application/octet-stream; charset=utf-8',
-    DefaultType: 'None',
     'Content-Security-Policy': "default-src 'none'",
     'X-Content-Security-Policy': "default-src 'none'",
     'X-WebKit-CSP': "default-src 'none'",
--- a/public/streamsaver/stream-saver.js
+++ b/public/streamsaver/stream-saver.js
@@ -80,6 +80,7 @@ self.onfetch = (event) => {
 
   const responseHeaders = new Headers({
     'Content-Type': 'application/octet-stream; charset=utf-8',
+    DefaultType: 'None',
     'Content-Security-Policy': "default-src 'none'",
     'X-Content-Security-Policy': "default-src 'none'",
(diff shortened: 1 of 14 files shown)


public/streamsaver/stream-saver.js
```diff
@@ -80,6 +80,7 @@ self.onfetch = (event) => {
 
   const responseHeaders = new Headers({
     'Content-Type': 'application/octet-stream; charset=utf-8',
+    DefaultType: 'None',
```
CandelR: I think that this header does not exists https://developer.mozilla.org/en-US/docs/Glossary/Response_header

CandelR (changes requested):

src/services/stream.service.ts
```diff
@@ -19,7 +21,7 @@ export async function binaryStreamToBlob(stream: BinaryStream, mimeType?: string
     finish = done;
   }
 
-  return new Blob(slices as BlobPart[], mimeType ? { type: mimeType } : {});
+  return new Blob(slices as BlobPart[], mimeType ? { type: mimeType } : { type: DEFAULT_BLOB_MIME_TYPE });
```
larryrider: Maybe its more readable this way:
`return new Blob(slices as BlobPart[], { type: mimeType ?? DEFAULT_BLOB_MIME_TYPE });`

wdyt?

public/streamsaver/stream-saver.js
xabg2: Indeed... hahahaha. it is Friday man... nothing to say but I'm gonna remove it.

src/services/stream.service.ts
xabg2: Mmm yep, better! 🚀

public/streamsaver/stream-saver.js
CandelR: 😆 😆 😆

CandelR (approved):

larryrider (approved):

PR #1830, opened 2026-01-27
[PB-5357]: fix/update usage using the socket event

xabg2:
## Description

The user’s storage limit is now updated whenever an event is received from the backend.

When a user purchases a plan, the backend sends a socket event called `PLAN_UPDATED` containing the new storage (in bytes). This value is now used to update the user’s storage limit directly, without requiring an additional fetch to retrieve the updated limit.

## Checklist

- [x] QA Passed


src/services/sockets/socket.service.test.ts
```diff
@@ -76,32 +74,32 @@ describe('RealtimeService', () => {
...
-    it('establishes a secure connection with authentication when initialized', () => {
+    test('When init is called, then it establishes a secure connection with authentication', () => {
       service.init();
 
       expect(ioMock).toHaveBeenCalledWith('https://notifications.example.com', {
         auth: { token: 'mock-token-123' },
-        reconnection: false,
-        withCredentials: true,
+        reconnection: true,
+        withCredentials: false,
```
CandelR: connection with autentication withCredentials set to false?

src/services/sockets/socket.service.ts
```diff
@@ -27,18 +26,31 @@ export default class RealtimeService {
       auth: {
         token: getToken(),
       },
-      reconnection: isProduction(),
-      withCredentials: true,
+      reconnection: true,
+      withCredentials: isProduction(),
```
CandelR, line 30: need to check this, thats because the test that I left a comments is weird

src/services/socket.service.ts
```diff
@@ -52,37 +64,40 @@ export default class RealtimeService {
...
-      return false;
+  onEvent(cb: (data: any) => void): () => void {
+    if (!isProduction()) {
+      console.log('[REALTIME] Registering event handler. Total handlers:', this.eventHandlers.size + 1);
     }
 
-    this.socket?.on('event', (data) => {
+    this.eventHandlers.add(cb);
+
+    // Return cleanup function
```
CandelR: no need this comment :)

src/views/Drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -308,20 +309,23 @@ const DriveExplorer = (props: DriveExplorerProps): JSX.Element => {
...
+              folderIds: [item.folderUuid],
+              items: [item],
+            }),
+          );
+        }
       }
-    }
-  };
+    },
+    [currentFolderId, dispatch],
```
CandelR: maybe if we move the functions that handles the events to one place we could reuse it if it is necessary and will be easy to mantain :)

src/App.tsx
```diff
@@ -88,6 +90,30 @@ const App = (props: AppProps): JSX.Element => {
     i18next.changeLanguage();
   }, []);
 
+  const handlePlanUpdatedEvent = useCallback(
+    (data: EventData) => {
+      if (data.event === SOCKET_EVENTS.PLAN_UPDATED) {
+        const newLimit = data.payload?.maxSpaceBytes;
+
+        if (newLimit) {
+          dispatch(planActions.updatePlanLimitFromSocket(Number(newLimit)));
```
CandelR: If the limit does not come in the event data, perhaps we can do a refetch of the limit, so we can control that case as well to update it

src/services/sockets/types/socket.types.ts
```diff
@@ -0,0 +1,27 @@
...
+} as const;
+
+interface BaseEventData {
+  email: UserSettings['email'];
+  clientId: string;
+  userId: UserSettings['userId'];
+}
+
+interface FileCreatedEvent extends BaseEventData {
+  event: 'FILE_CREATED';
```
CandelR: why not use the SOCKET_EVENTS constants? :)

larryrider (approved):

CandelR (approved):

PR #1833, opened 2026-01-29
[PB-5493]: feature/show the downloaded bytes when downloading items

xabg2:
Display the downloaded bytes progress when downloading folders/items.

## Checklist

- [ ] QA Passed


sg-gs (commented):
Does this keep the items count progress also? We added that because for cases with a lot of small files, the download bytes was not the most realistic way of reporting the download progress

xabg2:
> Does this keep the items count progress also? We added that because for cases with a lot of small files, the download bytes was not the most realistic way of reporting the download progress

No, it doesn't. Because the count + download progress don't fit in the task log. Also, Paula and I discussed it yesterday and agreed that it's better to show the user that something is happening rather than the items, because if there is a heavy item, the count doesn't add up, but the download progress does.

sg-gs (approved):

larryrider (approved):

PR #1834, opened 2026-02-03
[BR-1794]: Add rate limit retry mechanism with exponential backoff

terrerox:
## Description

  Implement retry utility to handle 429 rate limit errors with configurable backoff strategy, extract rate limit info
   from response headers, and apply it to public shared folder content requests to improve reliability under rate
  limiting conditions.

## Checklist

- [ ] QA Passed


src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+  maxRetries: 5,
+  maxDelay: 60000,
+  onRetry: () => {},
+};
+
+function isErrorWithStatus(error: unknown): error is ErrorWithStatus {
+  return typeof error === 'object' && error !== null;
+}
+
+function isRateLimitError(error: unknown): boolean {
```
CandelR: Use arrow functions to be consistent with rest of the codebase

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (!isErrorWithStatus(error)) {
+    return false;
+  }
+
+  const message = typeof error.message === 'string' ? error.message.toLowerCase() : '';
+
+  return (
+    error.status === 429 ||
+    error.statusCode === 429 ||
+    error.response?.status === 429 ||
```
CandelR: If I remember correctly, there are status codes, use them or add this one if it is not there

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+    message.includes('429') ||
+    message.includes('too many requests')
```
CandelR: I do not think this check is necessary, in some extreme cases, it could produce a false positive

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+}
+
+function extractRateLimitInfo(error: unknown): RateLimitInfo | undefined {
+  if (!isErrorWithStatus(error) || !error.headers) {
+    return undefined;
+  }
+
+  const headers = error.headers;
```
CandelR, line 26: I think that headers in Axios comes in `error.reponse.headers`: https://axios-http.com/docs/handling_errors. Maybe should add a generic ways to obtain the headers

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  if (!headers['x-ratelimit-reset']) {
+    return undefined;
+  }
+
+  const resetValueMs = Number.parseInt(headers['x-ratelimit-reset'], 10);
```
CandelR: the unit that return the server is in miliseconds?

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+      const rateLimitInfo = extractRateLimitInfo(error);
+      const delayMs = rateLimitInfo ? rateLimitInfo.retryAfter : opts.maxDelay;
+
+      opts.onRetry(attempt + 1, delayMs, rateLimitInfo);
+
+      await wait(delayMs);
+    }
+  }
+
+  throw new Error('Maximum retries exceeded');
```
CandelR: Is this error reachable? Check it with tests

src/app/share/services/share.service.ts
```diff
@@ -123,12 +124,21 @@ export function getPublicSharedFolderContent(
...
-      throw error;
-    });
+  return retryWithBackoff(
+    async () => {
+      const shareClient = SdkFactory.getNewApiInstance().createShareClient();
+      return shareClient.getPublicSharedFolderContent(sharedFolderId, type, token, page, perPage, code, orderBy);
+    },
+    {
+      maxRetries: 5,
+      maxDelay: 60000,
```
CandelR: these are the default values, i think that not need to put it

src/app/share/services/share.service.ts
```diff
@@ -123,12 +124,21 @@ export function getPublicSharedFolderContent(
...
+      onRetry: (attempt, delay) => {
+        console.log(
```
CandelR: maybe better use console warn if want to log the retry for when user records the console and send to us

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+    error.response?.status === 429 ||
+    message.includes('429') ||
+    message.includes('too many requests')
+  );
+}
+
+function extractRateLimitInfo(error: unknown): RateLimitInfo | undefined {
+  if (!isErrorWithStatus(error) || !error.headers) {
+    return undefined;
+  }
```
CandelR: this is already checked in `isRateLimitError`, I think that never will reach this code

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+    : undefined;
+
+  return {
+    retryAfter: resetValueMs,
+    limit,
+    remaining,
+  };
+}
+
+export async function retryWithBackoff<T>(fn: () => Promise<T>, options: RetryOptions = {}): Promise<T> {
```
CandelR: add jsdoc to this exported function

CandelR (changes requested):
Is this accompanied by some kind of visual information for the user? I don't see the point of this PR from the user's point of view, as they will be waiting at least 60 seconds without knowing what is happening, which is an enormous amount of time.

src/app/share/services/share.service.ts
sg-gs: I find 60s a pretty hard value. The Retry-After [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) should be the reference, not random fixed values

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (attempt === opts.maxRetries) {
+        throw error;
+      }
+
+      if (!isRateLimitError(error)) {
+        throw error;
+      }
+
+      const rateLimitInfo = extractRateLimitInfo(error);
+      const delayMs = rateLimitInfo ? rateLimitInfo.retryAfter : opts.maxDelay;
```
sg-gs: If it is a rate limit error, this header should be present. If not, then it is not a rate limit error

src/app/network/retry-with-rate-limit.ts
sg-gs: I agree with @CandelR

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,104 @@
...
+    return false;
+  }
+
+  const message = typeof error.message === 'string' ? error.message.toLowerCase() : '';
+
+  return (
+    error.status === 429 ||
+    error.statusCode === 429 ||
+    error.response?.status === 429 ||
+    message.includes('429') ||
```
sg-gs: Same thing here, if the status is not 429, this check is just a patch
cc @CandelR

sg-gs (changes requested):

src/app/network/retry-with-rate-limit.ts
terrerox: yes

src/app/network/retry-with-rate-limit.ts
terrerox: Done! I created this PR https://github.com/internxt/sdk/pull/360 with that change. Once is merged i will implement it and set this PR ready for review @CandelR

src/app/network/retry-with-rate-limit.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      const retryAfter = extractRetryAfter(error as ErrorWithStatus);
+
+      if (!retryAfter) {
+        throw error;
+      }
+
+      if (!hasShownRateLimitToast) {
+        hasShownRateLimitToast = true;
+        notificationsService.show({
```
sg-gs: I would not mix UI related things with pure logic like this function. Return something that allows the function consumer to determine if showing this notification or not proceeds

src/app/i18n/locales/de.json
```diff
@@ -1150,7 +1150,8 @@
       "copy-to-clipboard": "Link in die Zwischenablage kopiert",
       "link-updated": "Linkeinstellungen aktualisiert",
       "link-deleted": "Link gelöscht",
-      "links-deleted": "Link(s) gelöscht"
+      "links-deleted": "Link(s) gelöscht",
```
CandelR, line 1153: Seems that here is missing `error-deleting-links` translation, can you add it? :)

src/app/share/services/share.service.ts
```diff
@@ -123,12 +126,27 @@ export function getPublicSharedFolderContent(
...
+        console.warn(
+          `[PUBLIC-SHARED-${type.toUpperCase()}] Retry attempt ${attempt} after ${delay}ms for folder ${sharedFolderId}`,
+        );
+        if (!hasShownRateLimitNotification) {
+          hasShownRateLimitNotification = true;
+          notificationsService.show({
+            text: t('shared-links.toast.rate-limit-retry'),
+            type: ToastType.Warning,
+            duration: Infinity,
+          });
```
CandelR: Move this out of the service, return the error, and handle it in the manager component of the UI to decide what to display. This way, we try to keep it clean. Even if there is a function with the `notificationService` in this service, we will try not to mix it up any further :)

sg-gs (approved):

CandelR (approved):

terrerox:
PR closed due to the retry mechanism being applied globally https://github.com/internxt/sdk/pull/370

PR #1838, opened 2026-02-05
[PB-5715] fix(thumbnails): use workspace bucket for workspace uploads

douglas-xt:
## Description

When users upload files from desktop clients to their workspaces, both the file and thumbnail are sent to the correct workspace bucket using the appropriate context. However, the web application always uses the personal storage context when downloading thumbnails, regardless of where they were actually stored. This causes thumbnails from workspace uploads to fail to display. The issue only appears with desktop uploads because when uploading directly from web, it uses the same incorrect personal context for both upload and download, so they accidentally match and work.

The file upload flow was corrected by adjusting how we determine whether a file belongs to a workspace or personal storage. Instead of using a hardcoded value, the upload context is now evaluated dynamically to retrieve the appropriate credentials from the environment. This ensures that when uploading a file to a workspace, both the file and its thumbnail are stored in the same workspace storage location, guaranteeing that files and their thumbnails always reside together in the correct storage space.

Additionally, backward compatibility was implemented for existing thumbnails that were previously stored in the wrong location. When downloading a thumbnail,  the actual storage location is verified and the appropriate credentials are used to retrieve it. This means users with older thumbnails that were incorrectly stored in personal buckets can still view them seamlessly, while all new uploads follow the corrected behavior. This approach ensures a smooth transition without requiring migration of existing thumbnails or disrupting the user experience.

## Checklist

- [ ] QA Passed

## Testing Process

1. From the Windows desktop client, upload an image file to a workspace                       
2. Open the web app and navigate to that workspace — the thumbnail should be visible in the file list without opening the file viewer
3. From User A, upload an image to a workspace
4. Log in as User B (with access to the same workspace) and navigate to it — the thumbnail should be visible in the file list without opening the file viewer
5. For existing files uploaded before this fix, thumbnails stored in the personal bucket should still be visible 

## Additional Notes


src/app/network/UploadManager.ts
```diff
@@ -178,7 +178,7 @@ class UploadManager {
             }
           },
           {
-            isTeam: false,
+            isTeam: !!this.options?.ownerUserAuthenticationData?.workspaceId,
```
sg-gs, line 181: Is this flag the workspace correct flag or a deprecated flag of the previous deprecated version? Provide a GitHub ref demonstrating so pls

douglas-xt: isTeam is the name of an external boolean variable ([ref](https://github.com/internxt/drive-web/blob/6ec72f8356cab37dd7ecd95f798d34ec7942883b/src/app/drive/services/thumbnail.service.ts#L159)) that's passed as an argument to get environment config ([ref](https://github.com/internxt/drive-web/blob/6ec72f8356cab37dd7ecd95f798d34ec7942883b/src/app/drive/services/network.service/index.ts#L89)). The confusion  likely comes from outdated JSDoc ([ref](https://github.com/internxt/drive-web/blob/6ec72f8356cab37dd7ecd95f798d34ec7942883b/src/app/drive/services/network.service/index.ts#L86)) that  references isTeam, but the actual parameter is isWorkspace.

CandelR (approved):

CandelR (approved):
@douglas-xt Remember to fill the testing process :). Good job 🚀

PR #1859, opened 2026-02-13
[PB-5882]: feat/show request Id if exists when an error occurs

xabg2:
## Description

When a request fails and the x-request-id header is available in the response, it is included in the error toast notification.

## Checklist

- [ ] QA Passed


sg-gs (approved):
Just a thing that could be more maintainable if possible: could be better to decorate the error inside the SDK, where the client centralises the error handling rather than expanding the cast function around this app? This + showing the requestId on the toast if it is present could be easier to manage, considering that this could be extended to other apps like Meet and other platforms like Mobile

xabg2:
> Just a thing that could be more maintainable if possible: could be better to decorate the error inside the SDK, where the client centralises the error handling rather than expanding the cast function around this app? This + showing the requestId on the toast if it is present could be easier to manage, considering that this could be extended to other apps like Meet and other platforms like Mobile


Okay, now we [return](https://github.com/internxt/sdk/pull/369) tthe `requestId` in the error if it is present in the response headers. 
The castError function now simply returns the error as-is when it is an instance of `AppError`. Otherwise, it extracts the headers from a generic Error (for example, errors automatically cast by async functions) to retrieve the `requestId` when available. We do not fetch the `requestId` manually anymore.

I will push the changes once the SDK PR is merged and new release is done.

src/views/Login/components/ChangePassword.tsx
```diff
@@ -32,20 +32,21 @@ export default function ChangePassword(props: Readonly<ChangePasswordProps>): JS
 
       if (backupData.mnemonic && validateMnemonic(backupData.mnemonic)) {
         setBackupKeyContent(uploadedBackupKeyContent);
-        return;
```
CandelR, line 35: why have removed this return?

src/app/notifications/components/NotificationToast/LongNotificationToast.tsx
```diff
@@ -10,10 +10,18 @@ const LongNotificationToast = ({
...
   onClose,
 }: Omit<ToastShowProps, 'duration'> & { visible: boolean; onClose: () => void }): JSX.Element => {
   let Icon: typeof CheckCircle | undefined;
   let IconColor: string | undefined;
+
+  const handleCopyRequestId = () => {
+    if (requestId) {
+      navigator.clipboard.writeText(requestId);
+    }
+  };
```
CandelR, line 27: WWhen the user presses copy, some visual feedback would be great, otherwise they won't know if it has been copied to the clipboard or not

src/views/Login/components/ChangePassword.tsx
xabg2: Because it is no longer needed. I added the error notification in the catch, not outside.

src/app/notifications/components/NotificationToast/LongNotificationToast.tsx
xabg2: Indeed

src/views/Login/components/ChangePassword.tsx
CandelR: ah right! 👌 
But be careful, because now if there is no mnemonic in backupData, the error message does not appear, whereas before it did

xabg2: Lol. Let me change it then

CandelR (approved):

PR #1865, opened 2026-02-23
[PB-5816]: refactor/video streaming

xabg2:
## Description

Refactored the video streaming architecture to scope the Service Worker under `video-stream`, isolating streaming traffic and simplifying client communication.

The updated flow works as follows:

- The Service Worker is now registered under the video-stream scope, meaning it only intercepts network requests within that path. This ensures streaming logic is fully isolated from the rest of the application.

- Introduced a dedicated player.html file that embeds the <video> element and coordinates communication with the Service Worker. The <video> element still performs native HTTP Range requests, which are transparently intercepted and fulfilled by the Service Worker under the video-stream scope. It is embedded to the client by using an iFrame.

- The client-side video streaming service/session logic has been refactored and simplified. It now communicates directly with the embedded HTML player.

## Checklist

- [x] QA Passed


src/app/drive/components/FileViewer/viewers/FileVideoViewer/FileVideoViewer.tsx
```diff
@@ -77,11 +73,12 @@ const FileVideoViewer = ({
 
     if (!bridgeUser || !userId || !mnemonic) {
       console.error('[FileVideoViewer] Missing credentials');
+      handleOnError('Missing credentials');
       return;
     }
 
     const session = new VideoStreamingSession({
-      fileId: file.fileId,
+      fileId: file.fileId ?? '',
```
larryrider: If fileId is nullish or empty, then it means it is an empty file so it cannot be played. Maybe its a good idea to disable the preview when that happens or just throw an error if this code is not really reachable
wdyt? 🤔

xabg2: Yep, we can do that. I just leave it as it is because the flow will throw an error when trying to download the requested chunk, but yeah, let's throw an early error.

src/app/drive/services/video-streaming.service/VideoStreamingSession.ts
```diff
@@ -15,126 +13,168 @@ export interface VideoStreamingSessionConfig {
...
-      if (this.chunkCache.size > CACHE_SIZE_LIMIT) {
-        const firstKey = this.chunkCache.keys().next().value;
-        if (firstKey) {
-          this.chunkCache.delete(firstKey);
-        }
-      }
+    if (this.iframe) {
+      this.iframe.remove();
+      this.iframe = null;
+      this.sendToIframe('DESTROY', {});
```
CandelR: I believe this will never do anything because the iframe is null and the condition will never be satisfied

src/app/drive/services/video-streaming.service/VideoStreamingSession.ts
```diff
@@ -15,126 +13,168 @@ export interface VideoStreamingSessionConfig {
...
 
-  getVideoUrl(): string | null {
-    return this.videoUrl;
-  }
+  private generateIframe(container: HTMLElement): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (document.getElementById('video-iframe')) {
+        this.iframe = document.getElementById('video-iframe') as HTMLIFrameElement;
+        resolve();
+      }
```
CandelR, line 44: without a return after the resolve the function will continue executing and add second iframe

src/app/drive/components/FileViewer/viewers/FileVideoViewer/FileVideoViewer.tsx
```diff
@@ -91,84 +88,57 @@ const FileVideoViewer = ({
...
-        generateThumbnailBlob(video, async (blob) => {
-          if (!blob) return;
-          await handlersForSpecialItems?.handleUpdateThumbnail(file, blob);
-        });
-      }
-    };
-
-    video.addEventListener('error', handleError);
-    video.addEventListener('canplay', handleCanPlay);
+    video.addEventListener('error', (error) => handleOnError(error.message));
```
CandelR: the `removeEventListener` will not work because it uses a different anonymous function.
`removeEventListener` compares by reference, so the arrow function of add and remove are two different objects in memory

xabg2: Indeed. IT has been updated.

src/app/drive/services/video-streaming.service/VideoStreamingSession.ts
xabg2: Yep. We should to emit the event before removing the iframe and updating the reference to null

larryrider (approved):

CandelR (approved):

PR #1866, opened 2026-02-24
[PB-5908]: add global retry with rate-limit notifications for SDK clients

terrerox:
## Description

Upgrade @internxt/sdk to 1.15.0 and introduce retry strategies for HTTP requests. Silent retries (2 attempts) are enabled globally, while Storage and Share clients use a user-facing notification strategy (5 attempts) with a cooldown-based warning toast. Adds hasElapsed utility to date.service and i18n translations for the rate-limit toast in all supported languages

## Checklist

- [ ] QA Passed


.github/workflows/playwright.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: '20.11.1'
+          node-version: '20.19.0'
```
terrerox, line 20: Bumped from 20.11.1 to 20.19.0 to fix failing Playwright CI job.                                               
@noble/hashes@2.0.1 (transitive dep of @internxt/sdk@1.15.0) requires Node >= 20.19.0,
which caused `yarn install` to fail with "Found incompatible module".

src/app/core/factory/sdk/retryStrategies.ts
```diff
@@ -0,0 +1,39 @@
...
+const RATE_LIMIT_TOAST_COOLDOWN_MINUTES = 1;
+let lastRateLimitToastShownAt: Dayjs | null = null;
+
+export const resetToastCooldown = (): void => {
+  lastRateLimitToastShownAt = null;
+};
+
+export const retryStrategies = {
+  silent: (label = 'Global'): RetryOptions => ({
+    maxRetries: 2,
```
CandelR: avoid magic numbers, add it to a descriptive constants

src/app/core/factory/sdk/index.test.ts
```diff
@@ -66,6 +77,14 @@ describe('SdkFactory', () => {
     SdkFactory.initialize(mockDispatch, mockLocalStorage);
   });
 
+  describe('initialize', () => {
+    it('When initialized, then the global retry is set to silent', () => {
```
CandelR: I would add another test to check that it is not enabled if it has not been set :)

src/app/core/factory/sdk/retryStrategies.ts
```diff
@@ -0,0 +1,39 @@
...
+      console.warn(`[SDK] ${label} rate limited. Retry attempt ${attempt}, waiting ${delay}ms`);
+      const isToastOnCooldown =
+        lastRateLimitToastShownAt &&
+        !hasElapsed(lastRateLimitToastShownAt, RATE_LIMIT_TOAST_COOLDOWN_MINUTES, 'minute');
+      if (!isToastOnCooldown) {
+        lastRateLimitToastShownAt = dayjs();
+        notificationsService.show({
+          text: t('sdk.rateLimitToast'),
+          type: ToastType.Warning,
+          duration: 60000,
```
CandelR: I just realised that there is nothing to indicate that the SDK retries only run with 429s. If retries are ever added for a 409, for example, and the SDK is updated, it will still show notifications related to 429s, right?

CandelR: A retry strategy should not need to know how to display toasts or translate strings.
I suggest injecting the notification behaviour as a callback 🐊

terrerox: You're right, but if we ever change it to a 409 or any retryable error, I think the notification message ("This is taking longer than expected. Please wait...") will fit them all.

src/services/date.service.ts
```diff
@@ -31,13 +31,17 @@ export const getDaysUntilExpiration = (expiresAt: Date | string): number => {
   return Math.max(0, Math.ceil(diffInDays));
 };
 
+export const hasElapsed = (since: Dayjs, amount: number, unit: dayjs.ManipulateType): boolean =>
```
CandelR: I don't think it's necessary to export the function if it's already exported through the dataService. I would remove the export

src/app/core/factory/sdk/index.test.ts
```diff
@@ -66,6 +111,20 @@ describe('SdkFactory', () => {
...
+      expect(HttpClient.enableGlobalRetry).toHaveBeenCalledWith(
+        expect.objectContaining({ maxRetries: SILENT_MAX_RETRIES, onRetry: expect.any(Function) }),
+      );
+    });
+
+    it('When not initialized, then the global retry is not enabled', () => {
+      vi.clearAllMocks();
+
+      expect(HttpClient.enableGlobalRetry).not.toHaveBeenCalled();
+    });
```
CandelR: This test does not make sense. If no code is executed, it is logical that it has never been called 😵‍💫

terrerox: What did you mean by, “I would add another test to check that it is not enabled if it has not been set :)”?
I think I misunderstood.

CandelR (dismissed):

src/app/core/factory/sdk/index.ts
```diff
@@ -11,6 +11,34 @@ import { Checkout } from '@internxt/sdk/dist/payments';
...
 import { STORAGE_KEYS } from 'services/storage-keys';
 import { Location } from '@internxt/sdk';
+import { HttpClient } from '@internxt/sdk/dist/shared/http/client';
+import dayjs, { Dayjs } from 'dayjs';
+import dateService from 'services/date.service';
+import notificationsService, { ToastType } from 'app/notifications/services/notifications.service';
+import { t } from 'i18next';
+import { retryStrategies, NotifyUserCallback } from './retryStrategies';
+
+const RETRY_TOAST_DURATION_MS = 60000;
```
TamaraFinogina: Maybe we should move all constants in src/app/core/constants.ts so that they are all in one place. @CandelR do we have a file with constants specific to sdk or is the one in core the only one?

sg-gs (dismissed):

src/services/date.service.ts
```diff
@@ -31,13 +31,17 @@ export const getDaysUntilExpiration = (expiresAt: Date | string): number => {
   return Math.max(0, Math.ceil(diffInDays));
 };
 
+const hasElapsed = (since: Dayjs, amount: number, unit: dayjs.ManipulateType): boolean =>
```
TamaraFinogina, line 39: Do we need a function for this? It's called in only one place and it's a one-line function

src/app/core/factory/sdk/retryStrategies.ts
```diff
@@ -0,0 +1,23 @@
+import { RetryOptions } from '@internxt/sdk/dist/shared';
+
+export const SILENT_MAX_RETRIES = 2;
```
TamaraFinogina, line 7: Same with those constants

src/services/date.service.ts
larryrider: For me its better that way: 
- Readability: hasElapsed(since, 5, 'minute') reads closer to plain English than dayjs().diff(since, 'minute') >= 5, which requires _'knowing'_ the dayjs API
- Testability: a named function is easier to unit test in isolation
- Future reuse: "called in only one place now" doesn't mean it won't be reused later 🤔 
- Abstraction: it hides the dayjs dependency, so if you ever swap date libraries, there's one place to change

wdyt? @TamaraFinogina

larryrider (dismissed):

TamaraFinogina (dismissed):

sg-gs (approved):
Don't want to block this, myself, dismiss the approval @TamaraFinogina if affects your comments

src/app/core/factory/sdk/corsErrorInterceptor.ts
```diff
@@ -0,0 +1,57 @@
...
+export const corsMaskedErrorInterceptor: CustomInterceptor = {
+  response: {
+    onRejected: (error: unknown) => {
+      const axiosError = error as AxiosError;
+
+      if (isCORSMaskedError(axiosError)) {
+        const headers = { 'retry-after': RETRY_AFTER_SECONDS };
+
+        axiosError.response = {
+          status: 429,
```
TamaraFinogina, line 43: ideally it should be a constant

TamaraFinogina (approved):

src/app/core/factory/sdk/corsErrorInterceptor.ts
terrerox: Will add it in another PR

PR #1891, opened 2026-03-20
[PB-5298]: Add unit tests for referral service and loadExternalScript utility

terrerox:
## Checklist

- [ ] QA Passed

### DIFF
--- a/src/services/referral.service.test.ts
+++ b/src/services/referral.service.test.ts
@@ -1,12 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
-vi.mock('./local-storage.service', () => ({
-  default: {
-    get: vi.fn(),
-    set: vi.fn(),
-  },
-}));
-
 vi.mock('./env.service', () => ({
   default: {
     getVariable: vi.fn(),
@@ -33,11 +26,10 @@ vi.mock('utils/loadExternalScript', () => ({
   loadExternalScript: vi.fn().mockResolvedValue(undefined),
 }));
 
-import localStorageService from './local-storage.service';
 import envService from './env.service';
 import dateService from './date.service';
 import { loadExternalScript } from 'utils/loadExternalScript';
-import referralService from './referral.service';
+import type referralServiceType from './referral.service';
 
 const BANNER_STATE_KEY = 'referral_banner_state';
 const UCC_STORAGE_KEY = 'cello_ucc';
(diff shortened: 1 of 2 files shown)


src/services/referral.service.test.ts
```diff
@@ -0,0 +1,394 @@
...
+      push: (fn: (cello: { boot: typeof mockBoot }) => void) => fn({ boot: mockBoot }),
+    };
+    globalThis.cello = { cmd: autoExecCmd as unknown as typeof globalThis.cello.cmd };
+  });
+
+  return { mockBoot };
+};
+
+describe('referralService', () => {
+  beforeEach(() => {
```
CandelR: if no reset the module as in `loadExternalScript` tests I think that bootPromise won't be reseted between tests. Can you confirm it?

src/services/referral.service.test.ts
```diff
@@ -0,0 +1,394 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('./local-storage.service', () => ({
+  default: {
+    get: vi.fn(),
+    set: vi.fn(),
+  },
+}));
```
CandelR: as localStorageService is a passthrough interface with no logic of its own regarding localStorage, I think that is not necessary to mock local-storage with vitest, it should work without mocking it.
https://vitest.dev/guide/environment

src/services/referral.service.test.ts
terrerox: Thats right, i applied the same approach as `loadExternalScript.test.ts` to fix that

src/services/referral.service.test.ts
```diff
@@ -0,0 +1,398 @@
...
+  SdkFactory: {
+    getNewApiInstance: vi.fn(() => ({
+      createReferralsClient: vi.fn(() => ({
+        createReferralToken: vi.fn().mockResolvedValue({ token: 'mock-token' }),
+      })),
+    })),
+  },
+}));
+
+vi.mock('utils/loadExternalScript', () => ({
```
larryrider, line 29: for me its more readable if the imports are the first thing in the file, and then the mocks, the constants, etc.

right now its all mixed and seems a little messy

wdyt?

CandelR (commented):
The tests are failing 🙃

CandelR (approved):

larryrider (approved):

PR #1903, opened 2026-03-26
[_] feature/Added typescript check

CandelR:
## Description

- Added typescript commando to check type errors
- The command has been added to run in the precommit and action stages.
- Fixed related type errors

## Checklist

- [x] QA Passed

### File Download
- [ ] Download a single file from Drive
- [ ] Download multiple files simultaneously
- [ ] Cancel an progress download
- [ ] Download a file using Firefox
- [ ] Download a folder as ZIP (from Drive and Backups)

### PDF Viewer
- [ ] Open a valid PDF in the FileViewer, should render correctly
- [ ] Open a corrupted or empty PDF, should show "preview unavailable" without crashing

### Image Viewer
- [ ] Open a JPEG/PNG image, should display correctly
- [ ] Open a HEIC image, should convert and display
- [ ] Open an image with a broken URL, should show "preview unavailable"

### Version History
- [ ] Open the version history sidebar for a file

### PGP Key Generation / Sign Up
- [ ] Create a new account,  keys should be generated without errors (now uses `ed25519Legacy`)
- [ ] Log in with an existing account, keys load correctly

### Credential Update (not tested in local because CORS error)
- [ ] Password update flow completes without errors
- [ ] Verify the `keys.private.{ ecc, kyber }` payload is sent correctly to the backend


src/app/crypto/services/pgp.service.ts
```diff
@@ -32,7 +34,7 @@ export async function generateNewKeys(): Promise<{
 
   const { privateKey, publicKey, revocationCertificate } = await openpgp.generateKey({
     userIDs: [{ email: 'inxt@inxt.com' }],
-    curve: 'ed25519',
+    curve: 'ed25519Legacy',
```
sg-gs, line 37: Do you approve this @TamaraFinogina ?

src/app/drive/components/FileViewer/viewers/FileImageViewer/FileImageViewer.tsx
```diff
@@ -25,10 +25,14 @@ const FileImageViewer = ({
           const updatedFile = { ...file };
           handlersForSpecialItems?.handleUpdateProgress(PROGRESS_BAR_STATUS.PENDING);
 
+          if (!blob) {
```
sg-gs: Is this an expected behaviour @CandelR ? This seems to me than an error to handle

.github/workflows/ci.yml
```diff
@@ -53,6 +53,26 @@ jobs:
...
         run: yarn test
+  typecheck:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        node-version: [20.x]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
```
xabg2: `actions/setup-node@v1` is outdated. Maybe we can upgrade to `actions/setup-node@v4` or, ideally, the latest version `actions/setup-node@v6`.

Same for `actions/checkout@v3`, should be updated to the latest version, `actions/checkout@v6`.

src/libs/streamSaver/index.ts
```diff
@@ -40,7 +40,8 @@ export class StreamSaver {
     const transport: MitmTransporter = {
       frame: iframe,
       loaded: false,
-      postMessage: (...args: any[]) => iframe.contentWindow?.postMessage(...args),
+      postMessage: (message: ServiceWorkerMessage, targetOrigin: string, transfer?: Transferable[]) =>
+        iframe.contentWindow?.postMessage(message, targetOrigin, transfer),
```
sg-gs, line 44: WDYT @xabg2 ?

src/views/Drive/components/VersionHistory/Sidebar.tsx
```diff
@@ -88,7 +88,7 @@ const Sidebar = () => {
   useEffect(() => {
     if (item) {
       setCurrentVersion({
-        id: item.fileId,
+        id: item.uuid,
```
sg-gs: Is this expected @CandelR ?

src/utils/isFileEmpty.test.ts
```diff
@@ -9,6 +9,12 @@ describe('Is file empty check', () => {
     expect(isFileEmpty(emptyFile)).toBe(true);
   });
 
+  test('When the file has no fileId, then it should indicate it is empty', () => {
+    const emptyFile = { size: 0, fileId: null } as DriveFileData;
+
+    expect(isFileEmpty(emptyFile)).toBe(true);
```
xabg2, line 15: nit: consider using `toBeTruthy()`

.github/workflows/ci.yml
```diff
@@ -53,6 +53,26 @@ jobs:
...
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: 'https://npm.pkg.github.com'
+      - run: echo "registry=https://registry.yarnpkg.com/" > .npmrc
```
sg-gs: Is this still needed?

sg-gs (commented):
For all the reviewers here, be aware that although this PR is for enforcing types, there are logic changes that although could appear small or non-relevant, are still logic changes so they can trigger unintended effects, requiring a regression test. 

cc @ValentynaBlahodyrQA

xabg2 (dismissed):

larryrider (dismissed):

src/app/crypto/services/pgp.service.ts
CandelR: https://github.com/openpgpjs/openpgpjs/wiki/v6-Changelog
This change is because the update to v6.
```
Rename EdDSA-related enums following the standardization of new key formats:
- Rename string value of enums.curve.ed25519Legacy to 'ed25519Legacy' (was: 'ed25519')
```

src/libs/streamSaver/index.ts
xabg2: These are the params we pass to the SW. Also, the video streaming works on the preview link, so yeah, that's correct 👍🏽 .

These are the [args](https://github.com/internxt/drive-web/blob/master/src/libs/streamSaver/index.ts#L135). And we use this postMessage [here](https://github.com/internxt/drive-web/blob/master/src/libs/streamSaver/index.ts#L182) (mitmTransporter).

src/views/Drive/components/VersionHistory/Sidebar.tsx
CandelR: Yes, that's fine, it's only used as a key. Here, the `fileId` is not used for downloading, as it may be null if an empty file has been uploaded, I have modified it so that it always uses a unique string value

larryrider: then lets change it also to uuid so its consistent here:
https://github.com/internxt/drive-web/blob/9391e8488024e140998ccc96b93f4ba35f89558c/src/views/Drive/components/VersionHistory/Sidebar.tsx#L197

src/app/drive/components/FileViewer/viewers/FileImageViewer/FileImageViewer.tsx
CandelR: It’s mainly a defensive guard, it should never actually reach null, as the parent handles it. I’ve left it as it is to avoid making further changes to the other components of the PR prevent introducing any bugs since the main aim of the task was to add the type check to trigger warnings.
But it doesn’t seem like the best solution to leave it like this, so I’m going to review the parent class to see if a more appropriate solution can be implemented. Thanks for point it :)

src/app/crypto/services/pgp.service.ts
TamaraFinogina: @sg-gs, @CandelR, we can use both. This change only affects newly generated keys (account creation, account reset). The decryption/encryption algorithms of OpenPGP get information about the curve from the key, so they would know what to do. 

During the OpenPGP update to v6 (done on 25 Feb 2026, passed QA), I checked sharings between new accounts (with ed25519) keys and old ones (with keys that are now ed25519Legacy), all worked.

So, we can just choose which new keys we want to generate. Since 25 Feb 2026, we have used ed25519 for new keys (when OpenPGP was upgraded to v6). We can change to ed25519Legacy or keep ed25519. Sharing will work with both.

```
  const encryptedMessage = await openpgp.encrypt({
    message: await openpgp.createMessage({ text: message }),
    encryptionKeys: publicKey,
  });
```

larryrider: @TamaraFinogina In OpenPGP >v6 we can only use ed25519Legacy as the Curve name:

`type EllipticCurveName = 'ed25519Legacy' | 'curve25519Legacy' | 'nistP256' | 'nistP384' | 'nistP521' | 'secp256k1' | 'brainpoolP256r1' | 'brainpoolP384r1' | 'brainpoolP512r1';`

As Ramon has said, they just changed it from 'ed25519' to 'ed25519Legacy'

TamaraFinogina: @CandelR You are righ, ed25519 is set equal to ed25519Legacy

Screenshot 2026-03-26 at 12 53 08
t

@sg-gs It's the same, all good
@CandelR all good, thanks!

.github/workflows/ci.yml
CandelR: I’ve removed it as part of some other changes I’ve made, as it’s no longer needed. I’ll create tasks to update the actions, as some versions need updating, and the tokens aren’t necessary since the libraries were already on npm
cc: @larryrider

TamaraFinogina (approved):

larryrider (approved):

sg-gs (approved):

PR #1912, opened 2026-03-30
[BR-1933]: fix/plan size handling and comparison when updating plan

xabg2:
## Description

When a user already has a lifetime plan and attempts to purchase an additional one, the current warning message can be misleading. For example, if the user has a total storage limit of 10TB with only 8.1GB used and tries to purchase an additional 5TB plan, the alert incorrectly indicates that the user has 8TB and is about to buy 5TB. This happens because the system is not properly accounting for the existing lifetime plan. Instead, when a user with an active lifetime plan purchases another lifetime plan, the calculation should be based on the sum of the current storage limit and the new storage being added, rather than considering only the new plan in isolation.

## Checklist

- [ ] QA Passed

## Testing Process
- Log in with a user that already has an active lifetime plan with a known storage limit (e.g., 10TB) and usage (e.g., 8TB used) > Navigate to the Plan tab (Preferences) and select an additional lifetime plan (e.g., +5TB) > Verify that the warning message shown before completing the purchase does not appear.

### DIFF
--- a/src/views/NewSettings/components/Sections/Account/Plans/components/ChangePlanDialog.tsx
+++ b/src/views/NewSettings/components/Sections/Account/Plans/components/ChangePlanDialog.tsx
@@ -45,10 +45,9 @@ const ChangePlanDialog = ({
   const selectedPlanSize = priceSelected?.bytes;
   const selectedPlanAmount = priceSelected?.amount;
   const selectedPlanInterval = priceSelected?.interval;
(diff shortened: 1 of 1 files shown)


CandelR (commented):
Please fill in the fields requested in the PR description. 
The checklist and instructions on _Testing Process_ are missing.

CandelR (approved):

PR #1929, opened 2026-04-10
[_] feature/Updated dependencies and updated related tests

CandelR:
## Description

- Update dependencies and fix related security vulnerabilities                                                                                                        
- Update sdk
- Upload was broken locally with a `window is undefined` error. `uploadWorkerUtils` imported `notifyUserWithCooldown` directly as a static dependency, causing it to be evaluated inside the worker context where `window` and other browser APIs (`notificationsService`, `i18next`) are not available. Refactored to receive `onRateLimited` as an injected callback from `network.service`, which runs in the main thread where those APIs work correctly.  
## Checklist

- [x] QA Passed

## Testing Process
- [ ] Login with wrong credentials shows backend error message                                              
- [ ] Password protected shared link - wrong password shows inline error, correct password loads content
- [ ] Invalid/expired shared link - shows "link expired" screen        
- [ ] Rate limit display notification
- [ ] Upload items working as expected


.github/workflows/ci.yml
```diff
@@ -49,10 +49,12 @@ jobs:
       - run: echo "always-auth=true" >> .npmrc
       - name: Install
         run: yarn
+      - name: Clear Vite / Vitest cache
+        run: rm -rf node_modules/.vite .vite
```
xabg2: Why? Is there a problem with cache or something like that and this fix it? 🤔

src/app/drive/services/download.service/createFileDownloadStream.test.ts
```diff
@@ -40,22 +36,23 @@ describe('createFileDownloadStream', () => {
     expect(result).toBe(mockReadableStream);
   });
 
+  test('When an abortController is provided without sharing options, then it is passed through to fetchFileStream', async () => {
```
xabg2: Try to avoid using technical descriptions for test. if the function name changes, the test does not make sense anymore.

src/services/error.service.test.ts
```diff
@@ -39,16 +55,108 @@ describe('Error Service', () => {
   });
 
   describe('castError', () => {
+    describe('Null and undefined inputs', () => {
+      it('returns Unknown error for null', () => {
```
xabg2: Try to follow the same way to describe tests as you did before (When..., then... pattern).

src/views/PublicShared/components/ShareItemPwdView.tsx
```diff
@@ -40,17 +40,16 @@ const ShareItemPwdView = (props: ShareItemPwdViewProps) => {
...
-        } else {
-          errorService.reportError(error);
-          notificationsService.show({
-            text: errorService.castError(error).message,
-            type: ToastType.Warning,
-            duration: 50000,
-          });
-        }
+      const appErr = errorService.castError(error);
+      if (appErr.status === 403) {
```
xabg2: Can we use `HTTP_CODES` to match the codes?

src/services/auth.service.ts
```diff
@@ -422,10 +421,11 @@ export const changePassword = async (newPassword: string, currentPassword: strin
       if (newToken) localStorageService.set(LocalStorageItem.NewToken, newToken);
     })
     .catch((error) => {
-      if (error.status === 500) {
+      const appErr = errorService.castError(error);
+      if (appErr.status === 500) {
```
xabg2: Can we use `HTTP_CODES` to match the codes?

.github/workflows/ci.yml
CandelR: I was running some tests because the CI tests were failing a lot, but this isn't necessary, it was sorted out in the Vitest settings. Thanks for asking, I hadn't realised :)

xabg2 (dismissed):

xabg2 (approved):

PR #1934, opened 2026-04-17
[_] Hide terms link in mobile view

CandelR:
## Description

- Hide terms link on mobile view to comply with Google Play requirements 

## Checklist

- [ ] QA Passed

## Testing Process

- Check that only on mobile views cannot acces to TC


xabg2 (commented):
nit: would be better to extract all the repeated URLs to a constant instead. Then, if we need two update it, we only need to update 1 link instead of 10

larryrider (approved):

PR #1935, opened 2026-04-21
[MKT-822]:feat/add isFirstPruchaseCheck to the impact services tracking

jaaaaavier:
## Description

Implemented logic to distinguish between new sales and returning customers during checkout to optimize Impact conversion tracking.

Key Changes:
- checkIsFirstPurchase: Added a new function in useUserPayment.ts that retrieves the user's invoice history to determine if the purchase is their first. The isFirstPurchase state is now stored in localStorage during the payment process for retrieval on the success page.

- Impact Service: Updated trackPaymentConversion to ensure conversion events are only sent to Impact if:
The purchase is confirmed as the user's first.

## Checklist

- [ ] QA Passed

## Testing Process

Simulated the checkout flow in development environment using new and existing user accounts.

Scenarios Tested:
- New Customer (First Purchase):
       - Step: Logged in with a user who has no previous invoice history.
       - Observed:
          - The invoices API call returned an empty array.
          - isFirstPurchase was correctly set to true in localStorage after clicking "Pay".

- Returning Customer (Repeat Purchase):
       - Step: Logged in with a user who already has a paid invoice.
       - Observed:
           - The invoices API call returned at least one record.
           - isFirstPurchase was correctly set to false in localStorage.
## Additional Notes
Video attached of the testing process attached on the [task](https://inxt.atlassian.net/jira/software/c/projects/MKT/boards/81?selectedIssue=MKT-822)


src/app/analytics/impact.service.ts
xabg2, line 36: Would be better to pass an object instead of function parameter, so you know what props you have.

src/views/Checkout/hooks/useUserPayment.ts
```diff
@@ -257,6 +281,7 @@ export const useUserPayment = () => {
     confirmSetupIntent,
   }: UseUserPaymentPayload) => {
     const planInterval = selectedPlan.price.interval;
+    const isFirstPurchase = await checkIsFirstPurchase();
```
xabg2, line 284: What if this fails here? Will the payments not go through? Is that the expected outcome? Unless it’s absolutely necessary, this shouldn’t interfere with the payment process. In other words, whether it works or fails, I understand that it shouldn’t affect the payment flow.

jaaaaavier: The checkIsFirstPurchase function contains an internal try/catch block. If the API call fails, it catches the error, logs it, and simply returns false. It should not throw an error, so the payment flow is never interrupted. However, if you think this is risky, I can look for an alternative.

xabg2: It's ok for me then! 🚀

xabg2 (approved):

PR #1941, opened 2026-04-28
[PB-5743]: feat/show dialog to download backup keys

xabg2:
## Description

After login or registration, a modal will appear reminding the user to download their backup key. The behavior is:

- If the user has never seen the dialog, it opens automatically.
- The user can download the backup key at any time using the download button.
- Clicking "Remind me later" dismisses the dialog and will not show it again for 14 days or he is authenticated again.
- Once the user downloads the key and clicks "I've saved my key", the dialog will never be shown again.
- If the user logs in with a different account, the dialog will appear again for that account (state is stored per user).
- The dialog will not appear again after logout and login (with the same account) if the user already acknowledged it.
- While the dialog is open, promotional banners should be hidden.

## Checklist

- [ ] QA Passed


src/services/local-storage.service.ts
```diff
@@ -11,6 +11,37 @@ function set(key: string, value: string): void {
   return localStorage.setItem(key, value);
 }
 
+function getBackupKeyStorageKeys() {
+  const user = getUser();
+  const userId = user?.uuid ?? 'anonymous';
```
larryrider: are we sure about saving "anonymous" if we dont have the user uuid? 🤔

src/utils/backupKeyUtils.ts
```diff
@@ -46,9 +49,12 @@ export function handleExportBackupKey(translate) {
     const backupData: BackupData = {
       mnemonic,
       privateKey: user.privateKey,
+      publicKey: user.publicKey,
       keys: {
         ecc: user.keys?.ecc?.privateKey || user.privateKey,
+        eccPublicKey: user.keys?.ecc?.publicKey || user.publicKey,
```
larryrider: preffer `??` over `||`

(same at all properties)

src/services/local-storage.service.ts
xabg2: I can use the userId or any other filed, but the user.uuid should exist always. Otherwise, something is going wrong. Let me take a look if we can use another user field.

src/utils/backupKeyUtils.ts
xabg2: Indeed 👍🏽

src/services/local-storage.service.ts
xabg2: What about passing `user?.uuid` directly? WDYT? @larryrider

larryrider: its more or less the same but yeah, it could be better this way 👍

larryrider (dismissed):

larryrider (approved):

PR #1942, opened 2026-04-29
[PB-6311]: feat/get correct user limit after purchase

xabg2:
## Description

Refactored the Realtime service (socket connection) by simplifying its initialization: it is now set up when the app mounts and continuously listens for events.

Additionally, a temporary polling mechanism (every 5 secs during 30 secs) has been introduced after a user purchases a plan to ensure their storage is updated correctly.

## Checklist

- [ ] QA Passed


src/services/sockets/socket.service.ts
```diff
@@ -33,22 +37,21 @@ export default class RealtimeService {
     });
 
     this.socket.on('connect', () => {
-      if (!this.isProduction) {
-        console.log('[REALTIME]: CONNECTED WITH ID', this.socket?.id);
-      }
+      console.log('[REALTIME]: CONNECTED WITH ID', this.socket?.id);
```
CandelR, line 39: I suppose this is done on purpose so that it appears in the production environment, and it’s not that you’ve forgotten to put it back :)
I’m just pointing it, just in case

src/services/sockets/socket.service.ts
```diff
@@ -33,22 +37,21 @@ export default class RealtimeService {
...
-        }
-      });
+      switch (data.event) {
+        case SOCKET_EVENTS.PLAN_UPDATED:
+          eventHandler.onPlanUpdated(data);
+          break;
+        case SOCKET_EVENTS.FILE_CREATED:
+          eventHandler.onFileCreated(data);
+          break;
+      }
```
CandelR: I think that the before implementation was good, now for add more events we have to modify the socket service and not only the event handler :( 
Maybe I'm missing something :O

src/services/sockets/socket.service.ts
```diff
@@ -58,45 +61,20 @@ export default class RealtimeService {
     });
 
     this.socket.on('connect_error', (error) => {
+      if (error.message === 'Session ID unknown') {
```
CandelR: This might be too susceptible to changes in the message. Does it return any code we can use? Although the message doesn't necessarily have to change, the code is usually less susceptible to changes than the message

src/services/sockets/socket.service.ts
xabg2: I can send the data.event directly to the eventHandler and handle there the events (using the switch). This way, we handle all in the same place, not instantiating it everywhere and using Sets...

src/services/sockets/socket.service.ts
xabg2: Yep, so we have visibility about the socket connection.

src/services/sockets/socket.service.ts
xabg2: Mmm seems like socket io does not throw a code within the error. I can use includes instead (`ID unknown` for example).

CandelR (approved):

PR #1943, opened 2026-04-29
  [MKT-834]:feat: account popover new design

jaaaaavier:
## Description
In this PR, we’ve updated the design of the popover that appears when you click on AccountPopover. Changes:
- We’ve changed the separator; it’s no longer a barely visible border but a div no wider than 1px, which helps us better distinguish between sections
- We’ve slightly rearranged the position of the elements and added two new ones: “Account” and “Give Feedback.” The first opens the preferences panel to the account section, just as it does in the settings. The second, meanwhile, opens a new tab to the page where users can provide feedback on the product; to do this, we’ve added the URL in the same way links to “meet” or “send” are managed
- Update all necessary content in the JSONs
- We've also slightly adjusted the width of the popover and the upgrade button to prevent the line break that occurred in some languages, which looked pretty ugly. 
## Checklist

- [ ] QA Passed

## Testing Process
Oppened the popover on all languages and checked the upgrade section shows as expected and all other buttons works as before

### DIFF
--- a/src/app/i18n/locales/zh.json
+++ b/src/app/i18n/locales/zh.json
@@ -428,7 +428,9 @@
         "logout": "登出",
         "referAndEarn": "推荐赚取",
         "earnReferral": "赚取 €100",
-        "spaceUsed": "已使用储存空间的{{space}}%"
+        "spaceUsed": "已使用储存空间的{{space}}%",
+        "account": "账户",
+        "giveFeedback": "提供反馈"
       },
       "tabs": {
         "account": {
--- a/src/app/i18n/locales/zh.json
+++ b/src/app/i18n/locales/zh.json
@@ -428,7 +428,7 @@
         "logout": "登出",
         "referAndEarn": "推荐赚取",
         "earnReferral": "赚取 €100",
-        "spaceUsed": "已使用储存空间的{{space}}%",
+        "spaceUsed": "已用存储空间{{space}}%",
         "account": "账户",
         "giveFeedback": "提供反馈"
       },
--- a/src/views/Home/components/AccountPopover.tsx
+++ b/src/views/Home/components/AccountPopover.tsx
@@ -78,7 +78,7 @@ export default function AccountPopover({ className = '', user, plan }: Readonly<
         </p>
 
         <button
-          className="w-min cursor-pointer  text-sm font-medium text-primary no-underline"
+          className="w-full cursor-pointer  text-sm font-medium text-primary no-underline"
           onClick={() => {
             navigationService.openPreferencesDialog({
               section: 'account',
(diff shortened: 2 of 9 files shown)


src/views/Home/components/AccountPopover.tsx
```diff
@@ -74,27 +76,36 @@ export default function AccountPopover({ className = '', user, plan }: Readonly<
...
-                workspaceUuid: selectedWorkspace?.workspaceUser.workspaceId,
-              });
-              dispatch(uiActions.setIsPreferencesDialogOpen(true));
-            }}
-          >
-            {translate('actions.upgrade')}
-          </button>
-        )}
+
+        <button
```
CandelR: The upgrade button will now always be visible, including for accounts with a subscription, is that correct?

src/app/i18n/locales/zh.json
```diff
@@ -428,7 +428,9 @@
         "logout": "登出",
         "referAndEarn": "推荐赚取",
         "earnReferral": "赚取 €100",
-        "spaceUsed": "已使用储存空间的{{space}}%"
+        "spaceUsed": "已用存储空间{{space}}%",
```
CandelR, line 431: The symbols for ‘spaceUsed’ are no longer the same as before, is this change correct?

src/views/Home/components/AccountPopover.tsx
```diff
@@ -74,27 +76,36 @@ export default function AccountPopover({ className = '', user, plan }: Readonly<
...
+            section: 'account',
+            subsection: 'account',
+            workspaceUuid: selectedWorkspace?.workspaceUser.workspaceId,
+          });
+          dispatch(uiActions.setIsPreferencesDialogOpen(true));
+        }}
+      >
+        <User size={20} />
+        <p className="ml-3">{translate('views.account.popover.account')}</p>
+      </button>
```
CandelR: I think the `Item` component could work for this, and while we’re at it, we could also change the one at the bottom that takes user to general section of `PrefencesDialog`. What do you think?

CandelR (commented):
Good @jaaaaavier , just a few details to mention :)

src/app/i18n/locales/zh.json
jaaaaavier: I updated the text because it was too long and threw off the layout quite a bit there was a line break that caused the {{space}} text to appear below it, making it look a bit odd. Everything was too cramped. The literal translations change is “Used storage space” -> “Used storage”

src/views/Home/components/AccountPopover.tsx
jaaaaavier: I forgot❌

larryrider (dismissed):

src/views/Home/components/AccountPopover.tsx
jaaaaavier: Replaced!

CandelR (approved):

PR #1947, opened 2026-05-05
[MKT-841]:feat/checkout flow improves for new annual plans

jaaaaavier:
## Description
In this PR, we’re updating a couple of things in the checkout flow in connection with the recent deployment of new annual plans. Key changes:
- We have replaced the content of `renewalTitle.monthly` (using `monthly` since that’s what is coming from the EP and it’s the simplest solution) with the text we need now. We might be able to implement some different logic, but this seems like a straightforward solution that will work for the moment.
- Added a new entry, annualBillingTemplate, which informs the user about how the plan they are about to subscribe to works; this text has been added to the checkout component
- Fixed a small bug that displayed an incorrect price in one of our dialogs
- Updated the billing type in the plan card header

## Checklist

- [ ] QA Passed

## Testing Process
Nothing out of the ordinary—just making sure everything is working as expected and everything is displayed correctly.


xabg2 (approved):
We need to improve this, as using monthly checks for annual texts is confusing.

PR #1950, opened 2026-05-07
[MKT-842]:feat/add default fallback for non special plans

jaaaaavier:
## Description
In this PR, we resolved an issue related to the plans we offer that aren’t displayed on the website and are used for other purposes. The point is that these plans have their own IDs and could appear at checkout, so they must be accounted for. Previously, we assumed that if a plan wasn’t one of the three we have on website, it would be the free plan and we used the 1GB path, this is no longer the case. To address this:
- We have modified the JSONs so that it is no longer the 1GB plan but the default one, we dynamically pass the size in bytes to this JSON
- We’ve also updated the checkoutProductCard so it can recognize when a plan has specific features (Array.isArray—this is done because if that’s not the case, it literally returns the path) or should use the default features. The approach for the title is the same

## Checklist

- [ ] QA Passed

## Testing Process
We have verified that the standard plans continue to display as expected and that these special plans are now being displayed. Examples:
- planId=price_1RjEJPFAOdcgaBMQifUoiJgL
- planId=price_1RBxrfFAOdcgaBMQxbEQlLsI


CandelR (changes requested):
The web crashes when enter to plans view `?preferences=open&section=account&subsection=plans` in settings 🙃.

jaaaaavier:
> The web crashes when enter to plans view `?preferences=open&section=account&subsection=plans` in settings 🙃.

It was a problem with the features, fixed!

src/app/i18n/locales/de.json
```diff
@@ -2051,9 +2051,10 @@
             ],
             "comingSoonFeatures": ["Mail", "Fotos"]
           },
-          "1GB": {
-            "title": "Kostenlos",
-            "features": ["1GB verschlüsselter Cloud-Speicher", "Verschlüsseltes VPN"]
+          "default": {
+            "freeTitle": "Kostenlos",
```
CandelR: `freeTitle` translation is used?

src/views/Checkout/components/CheckoutProductCard.tsx
```diff
@@ -79,11 +79,18 @@ export const CheckoutProductCard = ({
...
   const planType = isBusiness ? 'businessPlanFeaturesList' : 'planFeaturesList';
 
-  const productLabel = translate(`preferences.account.plans.${planType}.${bytes}.title`) ?? bytes;
-  const featureKeys =
-    translateList(`preferences.account.plans.${planType}.${bytes ?? 'freeFeatures'}.features`, {
-      returnObjects: true,
-    }) ?? translateList('preferences.account.plans.planFeaturesList.1GB.features');
+  const specificTitleKey = `preferences.account.plans.${planType}.${bytes}.title`;
+  const productLabel =
+    translate(specificTitleKey) === specificTitleKey
```
CandelR: Can assign the result of `translate(specificTitleKey)` to a const and use it in the ‘else’ branch of the condition, that way we avoid calling it twice

CandelR (dismissed):
It looks fine, ready for QA. Just two notes on the code :)

src/app/i18n/locales/de.json
jaaaaavier: No!

src/views/Checkout/components/CheckoutProductCard.tsx
jaaaaavier: Oh perfect

CandelR (approved):

PR #1954, opened 2026-05-11
[PB-5247] remove old token

TamaraFinogina:
Reverts internxt/drive-web#1953 (revert of the original PR, yep)

This PR removes the old token from drive-web. 

BLOCKED until the mobile stops expecting the old token


TamaraFinogina:
It's ready for merge, but we are waiting for mobile to do a new release with fixed login

PR #1955, opened 2026-05-11
[PB-5819]: feature/auto-restore items

xabg2:
## Description

Adding the logic to auto restore items if the parent folder exists. If to, then the "Move items" dialog is opened to select the destination of the items.

## Checklist

- [ ] QA Passed


src/hooks/moveItems/useMoveItems.ts
```diff
@@ -0,0 +1,137 @@
...
+    withErrorHandler(async () => {
+      const restorable = items.filter((item) => item.parent?.status === FileStatus.EXISTS);
+      const needsDestination = items.filter((item) => item.parent?.status !== FileStatus.EXISTS);
+
+      await Promise.all(
+        restorable.map((item) => {
+          dispatch(storageActions.setMoveDestinationFolderId(item.parent?.uuid as string));
+          return processMove({ finalDestinationId: item.parent!.uuid, items: [item] });
+        }),
+      );
```
CandelR: there is a race condition here as `setMoveDestinationFolderId` is overwritten in every call, anyway, as processMove receives de `finalDestinationId` is no need to set setMoveDestinationFolderId i think. 
Maybe we can add a test for this case :)

src/hooks/moveItems/useMoveItems.ts
```diff
@@ -0,0 +1,137 @@
...
+  finalDestinationId: string;
+  items: DriveItemData[];
+  displayTaskLogger?: boolean;
+}
+
+export const useMoveItems = () => {
+  const dispatch = useAppDispatch();
+  const itemsToMove: DriveItemData[] = useSelector((state: RootState) => state.storage.itemsToMove);
+  const processMove = async ({ finalDestinationId, items, displayTaskLogger }: ProcessMoveProps) => {
+    const processItems = items ?? itemsToMove;
```
CandelR: as `items` is a required field is no need `itemsToMove` fallback

CandelR (dismissed):

src/hooks/moveItems/useMoveItems.ts
xabg2: Yep. But I noticed that the `moveDestinationFolderId` is used when there are conflicts with the file in the folder we want to move in. `destinationFolderId` is used for "Name Collision". 

Let me think another way so we can avoid the race condition here.

src/hooks/moveItems/useMoveItems.ts
xabg2: Indeed

larryrider (dismissed):

larryrider (approved):

PR #1957, opened 2026-05-14
[PB-6328/PB-6333]: feat/validate file size before upload

xabg2:
## Description

Implemented a pre-validation step before starting uploads to filter out files that exceed the maximum upload size allowed for the user’s current tier.

If a file is larger than the allowed limit, a dialog is displayed informing the user about:

* their current upload size limit
* why the upload cannot proceed
* the next available tier required to upload larger files

This prevents invalid uploads from starting and provides clearer upgrade guidance to the user.

## Checklist

- [ ] QA Passed

## Additional Notes

This can be safely merged. If the server does not provide a `maxFileSizeLimit`, we allow the upload. The backend will check the file size limit if needed. So, this not affects to the current flow fn until we activate the limit in the backend.


src/views/Drive/components/DriveExplorer/DriveExplorer.tsx
```diff
@@ -818,6 +820,7 @@ export default connect((state: RootState) => {
 
   return {
     isAuthenticated: state.user.isAuthenticated,
+    maxUploadFileSize: state.fileVersions.limits?.maxUploadFileSize ?? undefined,
```
xabg2: I know it looks a bit weird to get the max upload file size from `fileVersions.limit`. The issue is that this value actually represents the user's Drive limits based on their tier, but when it was originally introduced it was named `fileVersions.limits` instead of `driveLimits` or smth like that. We should eventually refactor this and rename it to something more generic, like `limits`, to better reflect its real purpose.

It will be done in another PR so we don't go out of scope.

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -461,23 +497,35 @@ export const uploadItemsParallelThunk = createAsyncThunk<void, UploadItemsPayloa
         }),
       );
 
+    const openReachedPlanLimitDialog = () => {
```
TamaraFinogina: Can we make it a common constant for both?

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
```diff
@@ -144,9 +149,7 @@ export const uploadFolderThunk = createAsyncThunk<void, UploadFolderThunkPayload
         //Added wait in order to allow enough time for the server to create the folder
         await wait(500);
 
-        if (!rootFolderItem) {
-          rootFolderItem = createdFolder;
-        }
+        rootFolderItem ??= createdFolder;
 
         if (!rootFolderData) {
           rootFolderData = createdFolder;
```
TamaraFinogina, line 156: we already did rootFolderItem ??= createdFolder; just a line before, this if won't help

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        fullPathEdited: 'path',
+      },
+      currentFolderId: 'parent-uuid',
+      options: { taskId },
+    })(dispatch, buildGetState(100), {});
+
+    expect(uploadItemsParallelThunk).toHaveBeenCalledWith(expect.objectContaining({ files: [smallFile, bigFile] }));
+  });
+
+  test('When no size limit is configured, then all files are uploaded regardless of size', async () => {
```
TamaraFinogina: Don't we have some max limit for all to use instead of infinite?

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
xabg2: `rootFolderItem` and `rootFolderData` are different

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.test.ts
xabg2: Yeah, I can limit it to 40GB as default (the max size we currently have right now in prod).

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.ts
TamaraFinogina: ops, you are right, it's another one

src/app/store/slices/storage/storage.thunks/uploadFolderThunk.test.ts
```diff
@@ -178,7 +185,7 @@ describe('Upload Folder Thunk', () => {
     expect(uploadItemsParallelThunk).toHaveBeenCalledWith(expect.objectContaining({ files: [smallFile, bigFile] }));
   });
 
-  test('When no size limit is configured, then all files are uploaded regardless of size', async () => {
+  test('When no size limit is configured, then all files above the default size limit are uploaded', async () => {
```
TamaraFinogina: maybe below and not above?

src/app/store/slices/storage/storage.thunks/uploadItemsThunk.ts
```diff
@@ -459,7 +461,7 @@ export const uploadItemsParallelThunk = createAsyncThunk<void, UploadItemsPayloa
       };
     }
 
-    const allowedFilesToUpload = validateFileSize(dispatch, files, maxFileSize);
+    const allowedFilesToUpload = validateFileSize(dispatch, files, maxFileSize as number);
```
TamaraFinogina: Sonar says no need for 'as number'

larryrider (approved):

PR #1962, opened 2026-05-19
[PB-6824]:feat/new dynamic login desgin

jaaaaavier:
## Description
This PR introduces a new login design, with the following key changes:
- New animated background, featuring four colored patches that rotate gradually to create a dynamic effect
- New design for the input fields: some text has been updated, and a background color and shadow have been added to give them a raised appearance
- It is now clearer when a user has entered incorrect credentials
- The color of the text for terms and conditions, etc., has been changed, and the logo at the top has been centered

Everything has been implemented in both light and dark modes

The task dont say anything about the "new" page

## Checklist

- [ ] QA Passed

## Testing Process
Just been in login and check that everything is working as expected


src/views/Login/SignInView.tsx
```diff
@@ -11,19 +11,157 @@ export default function SignInView(props: Readonly<SignInProps>): JSX.Element {
   const { translate } = useTranslationContext();
 
   return (
-    <div className={`flex h-full w-full flex-col bg-login-gradient ${props.displayIframe ? '' : 'overflow-auto'}`}>
+    <div
+      className={`relative flex h-full w-full flex-col dark:bg-[#0A0F1C] ${props.displayIframe ? '' : 'overflow-hidden'}`}
+    >
+      <style>{`
+        @keyframes orbTL {
+          0%, 100% { transform: translate(0, 0) rotate(-164.506deg); }
```
CandelR: Move all this animation into a separate component, that will make it easier to reuse and maintain.
I can’t see this new background on the sign up screen.
Do we only want it on the login screen? It looks a bit odd.
Have you checked mobile login and sign up screens?

src/views/Login/components/LogIn.tsx
```diff
@@ -271,7 +277,7 @@ export default function LogIn(): JSX.Element {
             loading={isLoggingIn}
             buttonDataCy="loginButton"
             variant="primary"
-            disabled={isLoggingIn}
+            className="disabled:!bg-primary"
```
CandelR: remove disabled while logging is a specificaction of the task?

CandelR (dismissed):

src/views/Login/SignInView.tsx
jaaaaavier: Moved the animation, also checked with design if we need to update the Sign In and yes!

src/views/Login/components/LogIn.tsx
jaaaaavier: No the idea was to not having in gray, updated

CandelR (commented):
Remain modify the login and signup forms to both follow the same structure and review the Playwright tests that are failing :)

src/app/i18n/locales/en.json
```diff
@@ -98,14 +98,16 @@
     "environment": "World Environment Day"
   },
   "auth": {
-    "email": "Email",
+    "emailFloatingLabel": "Email address",
+    "email": "Email address",
```
larryrider, line 102: why having two differents translations for what it seems to be the same value? Is it really going to be different in the future?

"emailFloatingLabel" & "email"
"passwordFloatingLabel" & "password"

🤔

larryrider (dismissed):

src/app/i18n/locales/en.json
jaaaaavier: I think it’s better to keep the tags separate, it makes them easier to manage in case you want to change some text or something similar in the future. In any case, if it’s a problem or if it’s better to reuse them, it can be changed.

jaaaaavier:
In QA, we've found a couple more pages where we need to add a background

larryrider (dismissed):

CandelR:
> In QA, we've found a couple more pages where we need to add a background

Okay, just request us review when it’s ready :)

larryrider (dismissed):

jaaaaavier:
QA passed!

larryrider (approved):

larryrider (approved):

PR #1964, opened 2026-05-20
[MKT-850]:feat/new tracking for impact in checkout

jaaaaavier:
## Description

When traffic arrives directly at the checkout page via a DTC affiliate link (e.g. drive.internxt.com/checkout?irclickid=...), Impact tracking was not working*. The impactSource and impactAnonymousId cookies were never set because the website's handleImpact() never ran, so trackPaymentConversion() silently skipped sending the conversion event. As a result, subid1 and shareID were not populating in Impact's dashboard and sales were not attributed to affiliates.

The fix replicates the same tracking logic the website runs on the homepage flow, but now directly in the checkout page:

- If irclickid is present in the URL on load, handleImpactDTCCheckout sets the impactSource, impactAnonymousId, and impactClickId cookies on internxt.com domain.
- It also fires a page-view event to the Impact API with the irclickid and anonymousId, which is the step that allows Impact to later link that anonymous ID to the affiliate click when the conversion event arrives.
## Checklist

- [ ] QA Passed

### DIFF
--- a/src/app/analytics/impact.service.test.ts
+++ b/src/app/analytics/impact.service.test.ts
@@ -34,6 +34,7 @@ vi.mock('./utils', () => ({
     if (key === 'gclid') return '';
     return '';
   }),
+  setImpactCookies: vi.fn(),
 }));
 
 vi.mock('./addShoppers.services', () => ({
@@ -450,7 +451,7 @@ describe('handleImpactDTCCheckout', () => {
   const irclickid = 'TZr0kB1hUxyZWqOxHoVulWsMUkuRX82pgw77Sk0';
   const utmMedium = '312695';
 
-  it('When called with irclickid, then it sends a page view event to the Impact API with the correct payload', async () => {
+  it('When tracking an affiliate click, then it reports the event to Impact analytics with the click identifier', async () => {
     const getCookieMock = await import('./utils');
     vi.mocked(getCookieMock.getCookie).mockImplementation((key) => {
       if (key === 'impactAnonymousId') return 'anon_id';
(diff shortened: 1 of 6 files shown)


src/app/analytics/impact.service.test.ts
xabg2, line 1: Avoid using technical words. Describe what the test is, in fact, testing. Not the variables that are updated or used.

src/app/analytics/impact.service.ts
```diff
@@ -70,6 +70,58 @@ export function savePaymentDataInLocalStorage({
...
+
+    const anonymousExpiration = new Date();
+    anonymousExpiration.setFullYear(anonymousExpiration.getFullYear() + 10);
+
+    const trackingExpiration = new Date();
+    trackingExpiration.setDate(trackingExpiration.getDate() + 30);
+
+    document.cookie = `impactSource=Impact;expires=${sourceExpiration.toUTCString()};domain=${cookieDomain};Path=/`;
+    document.cookie = `impactAnonymousId=${anonymousId};expires=${anonymousExpiration.toUTCString()};domain=${cookieDomain};Path=/`;
+    document.cookie = `impactClickId=${irclickid};expires=${trackingExpiration.toUTCString()};domain=${cookieDomain};Path=/`;
```
xabg2: I think we already have something like `setKookie(...)`

src/app/analytics/impact.service.test.ts
jaaaaavier: Okay

src/app/analytics/impact.service.ts
jaaaaavier: I decided to create a new setImpactCookies, to be more specific

xabg2 (approved):

src/app/analytics/impact.service.ts
xabg2: It does not matter. `setcookies` is used to add, in fact, cookies. It does not matter how you use it.
Reference: https://github.com/internxt/drive-web/blob/master/src/app/analytics/utils.ts#L1

xabg2: instead to do document.cookie..., you just do `setCookie(impactSource, Impact, expDays)`

PR #1966, opened 2026-05-22
[PB-6433]: feature/design for request access to shared item flow

xabg2:
## Description

Implementing the necessary components and modifying the existing ones to allow the users to request access to a restricted shared items. The new components for this flow has been added but they are been commented so it does not infer with the correct flow until we add the logic.

Maybe the components are modified in the future, but these are the initial/basic ones.

## Checklist

- [ ] QA Passed


src/app/drive/views/RequestAccess/UserCheapCard.tsx
```diff
@@ -0,0 +1,36 @@
+import { UserSettings } from '@internxt/sdk/dist/shared/types/userSettings';
+import { useTranslationContext } from 'app/i18n/provider/TranslationProvider';
+import { Button } from '@internxt/ui';
+
+interface RequestAccessUserCardProps {
+  user: UserSettings | undefined;
+  onChangeAccount: () => void;
+}
+
+const TEXTAREA_MAX_LENGTH = 1000;
```
CandelR: this constant is not used in the component

src/app/i18n/locales/en.json
```diff
@@ -1077,6 +1083,9 @@
         }
       }
     },
+    "accessRequests": {
+      "pending": ""
+    },
```
CandelR: This translation is used? I don't find it. Also this is the only translation file that it has it

src/app/drive/components/ShareDialog/components/AccessRequests/list/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+    </div>
+  );
+};
+
+const AccessRequestsList = ({ accessRequestList }: { accessRequestList: AccessRequestItem[] }) => {
+  return (
+    <div className="flex flex-col w-full">
+      {accessRequestList.map((request, index) => (
+        <RequestCheap key={index} {...request} />
```
CandelR, line 51: Have some sonar cloud warnings

xabg2: Yes. The design/components are added but not used (so we do not break the current flow for now). This way, we just need to add the logic and this design later and that's it.

They will be fixed later when completing the task.

CandelR (approved):

PR #1976, opened 2026-05-29
[PB-6439]: open referral widget via ?referral=open URL param

terrerox:
## Description

Add a URL contract so the Desktop app can deep-link users directly into the referral experience on Drive instead of only landing on the main page.

Introduce the useReferralParamsChange hook, which reads the ?referral=open query param and opens the Cello referral widget via referralService.openPanel once the user and eligibility have loaded. The widget only opens for eligible users, and the param is stripped from the URL afterwards so refresh/back navigation does not reopen it. The hook is wired into SidenavWrapper, the existing referral entry point.

## Checklist

- [ ] QA Passed


src/views/Drive/hooks/useReferralParamsChange.tsx
```diff
@@ -0,0 +1,38 @@
...
+  useEffect(() => {
+    if (!shouldOpenReferral || hasOpened.current) return;
+    if (!user || !isReferralEligible) return;
+
+    hasOpened.current = true;
+
+    referralService.openPanel(
+      { name: user.name, lastname: user.lastname, email: user.email, emailVerified: user.emailVerified },
+      i18n.language,
+    );
```
CandelR: If this fails, should we notify the user in some way?

It would also be a good idea to add some tests to the hook :)

CandelR (approved):

PR #1996, opened 2026-06-17
[PB-6542] Add new sharings

TamaraFinogina:
## Description

This PR switches to new public sharing version

## Related Pull Requests

[Server change](https://github.com/internxt/drive-server-wip/pull/1079)

## Checklist

- [x] QA Passed

## Testing Process

1. When a file or a folder that has never been shared previously is publicly shared on this environment, then it can be viewed and downloaded on production. The sharing version should be inxt-v3.
2. When a file or a folder that has been shared previously is publicly shared on this environment, then it still can be viewed and downloaded on production. The sharing verision should be inxt-v2.
3. When a file or a folder is shared with a non-user via user-invitation, then it work as before.
4. When a file or a folder is shared with existing user, then it work as before. 
5. When user is invited to a workspace, then it works as before.


larryrider (dismissed):

CandelR (commented):
waiting server changes to test it ⚠️

TamaraFinogina:
@CandelR Server is ready, we can test it

src/app/share/services/share.service.ts
```diff
@@ -38,12 +38,14 @@ import { domainManager } from './DomainManager';
...
 import referralService from 'services/referral.service';
+import { generateFileBucketKey } from 'app/network/crypto';
+import { LocalStorageItem } from 'app/core/types';
 
 interface CreateShareResponse {
   created: boolean;
   token: string;
 }
-
+const NEW_SHARING_VERSION = 'inxt-v3';
```
CandelR, line 47: nit: It would be better if both the server and the website had a single source of truth, because if it changes again one day and this logic isn’t taken into account, it could break

src/app/share/services/share.service.ts
```diff
@@ -317,19 +329,13 @@ export const getPublicShareLink = async (
...
   try {
-    const publicSharingItemData = await createPublicShareFromOwnerUser(uuid, itemType);
-    const { id: sharingId, encryptedCode: encryptedCodeFromResponse } = publicSharingItemData;
-    const isUserInvited = publicSharingItemData.ownerId !== user.uuid;
-
-    if (isUserInvited && encriptedMnemonic) {
+    if (encriptedMnemonic) {
       const ownerMnemonic = await decryptMnemonic(encriptedMnemonic);
-      if (ownerMnemonic) mnemonic = ownerMnemonic;
+      if (ownerMnemonic) localStorageService.set(LocalStorageItem.UserMnemonic, ownerMnemonic);
```
CandelR: rewrite own mnemonic with another users one not seems good idea. This can lead in decript or encription errors if in some point we use this one. 
Maybe I’m getting this wrong, could you confirm it?

src/app/share/services/share.service.ts
TamaraFinogina: Yes, but I'm not sure where to put it. SDK types?

src/app/share/services/share.service.ts
TamaraFinogina: Old logic was:

Do sharing (which requires a mnemonic, so without a mnemonic it does something strange), then check if the encrypted mnemonic is given and the user is invited; if so, use it to decrypt the code. 

Now:

 I moved code decryption inside the function, so I need to have the mnemonic in advance. The only place that calls getPublicShareLink with an encrypted mnemonic is onCopyLink in case isAdvancedShareItem exists. I think it's sharing with an invitation. So, it should be the same. But I'm a bit confused about how it was working before: if the mnemonic was undefined prior to setting it -  it should have failed share creation; if the mnemonic of the invited user was already set at that point -  why double setting? I'm lost.

TamaraFinogina: @CandelR Fixed, should be exactly as before

src/app/share/services/share.service.test.ts
```diff
@@ -309,3 +311,97 @@ describe('Get shared link', () => {
...
+
+    const { createPublicShareFromOwnerUser } = await import('./share.service');
+    const { publicSharingItemData, plainCode } = await createPublicShareFromOwnerUser('uuid', 'file');
+
+    expect(plainCode).toBeDefined();
+    expect(publicSharingItemData).toBeDefined();
+    expect(spyDecrypt).not.toHaveBeenCalled();
+  });
+
+  test('When encrypted code changes, decrypt code', async () => {
```
sg-gs, line 500: When does this happen? I am not sure

src/app/share/services/share.service.ts
sg-gs: It should be the backend, the SDK is only a reflect of it @TamaraFinogina

src/app/share/services/share.service.test.ts
TamaraFinogina: @sg-gs, it's when we share the same file again. If the file was already shared, the server will not change sharings, and to access the file via the link, we need the same code as in the server records, so we decrypt it.

sg-gs (dismissed):
LGTM

sg-gs (dismissed):
LGTM. Remember to test already existing and new sharings @ValentynaBlahodyrQA

TamaraFinogina:
@larryrider @sg-gs The PR passed QA, but I couldn't merge it without 80% coverage. So, I added more tests. Can you please re-check?

WARNING: I changed the code a bit: [see last commit](https://github.com/internxt/drive-web/pull/1996/changes/2881295a922d7a4e9722345f1ef1a34cf4523ffb#diff-603350e72e8de56ac836180c5821e49ae80369662305b76e4b0f6c112948d739.) The change is minor, and in "if condition" we could never test. It seems to be dead code (old workspace functionality), but I was not able to confirm it.

sg-gs (approved):

larryrider (approved):

PR #1997, opened 2026-06-17
[PB-6509]: feat/verify-integrity-on-download

sg-gs:
## Description

Validate file's integrity by calculating the hash on download to get the HMAC of the downloaded file to compare against the network one. 

## Related Issues

None

## Related Pull Requests

None

## Checklist

- [X] Changes have been tested locally.
- [X] Unit tests have been written or updated as necessary.
- [X] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [X] No new warnings or errors have been introduced.
- [x] SonarCloud issues have been reviewed and addressed.
- [x] QA Passed

## Testing Process
- When uploading 10MB, 100MB and 1GB from production and downloading from this environment, the file is downloaded correctly (check integrity with hashes)
- When uploading a 1GB file on this environment, the RAM usage stays almost the same as in the production version

## Additional Notes
None

PR #2000, opened 2026-06-19
[PB-6509]: feat/store-hmac-on-upload

sg-gs:
### What
- Adds `hmacSha512` (proper HMAC-SHA512 via [RFC 2104](https://datatracker.ietf.org/doc/html/rfc2104)) to `utils.ts` and updates `getFileHmacFromShardHashes` to use it instead of the previous `SHA512(key || data)` concatenation
- Derives a dedicated HMAC key via `HKDF(fileKey, info="for hmac", 512 bits)` to enforce cryptographic key separation between the AES-256-CTR encryption key and the HMAC-SHA512 authentication key
- Wires `computeHmac` into `NetworkFacade`'s `cryptoLib` so the SDK computes and sends an HMAC-SHA512 in the `finishUpload` payload during every upload
- Fixes the download verification path to convert SHA256 shard hashes to RIPEMD160 before computing the HMAC, matching the shard hash format stored by the bridge
- Bumps `@internxt/sdk` to `1.17.9` which exposes `computeHmac` in the `Crypto` interface and threads it through `finishUpload`/`finishMultipartUpload`

### Why 
Files uploaded to the bridge are encrypted client-side. Without an HMAC, a compromised server could substitute shard content and the client would silently decrypt garbage. The HMAC binds the encrypted shard (via its RIPEMD160(SHA256) hash, the same identifier the bridge stores) to the file key, so only a client that holds the key could have produced it. Using proper HMAC-SHA512 instead of `SHA512(key || data)` eliminates the length extension vulnerability present in the previous construction.

Regarding the HMAC key: using the same key for both encryption and authentication violates the cryptographic principle of key separation; HKDF ensures the two keys are independent even though they share the same origin.

## Related Issues
None

## Related Pull Requests
- https://github.com/internxt/drive-web/pull/1997

## Checklist

- [ ] QA Passed

## Testing Process
Specified on the Jira related ticket. 

## Additional Notes
None

### DIFF
--- a/src/app/crypto/services/utils.ts
+++ b/src/app/crypto/services/utils.ts
@@ -2,7 +2,7 @@ import { aes, items as itemUtils } from '@internxt/lib';
 import envService from 'services/env.service';
 import { Buffer } from 'buffer';
 import CryptoJS from 'crypto-js';
-import { blake3, createSHA256, createSHA512, ripemd160, sha256 } from 'hash-wasm';
+import { blake3, createHMAC, createSHA256, createSHA512, ripemd160, sha256 } from 'hash-wasm';
 import { DriveItemData } from 'app/drive/types';
 import { AdvancedSharedItem } from '../../share/types';
 
@@ -15,6 +15,11 @@ import { AdvancedSharedItem } from '../../share/types';
 async function getSha512Combined(key: Buffer, data: Buffer): Promise<string> {
   const hash = await createSHA512();
   return hash.init().update(key).update(data).digest();
+}
+
+async function hmacSha512(key: Buffer, data: Buffer): Promise<string> {
+  const hmac = await createHMAC(createSHA512(), key);
+  return hmac.init().update(data).digest();
 }
 
 interface PassObjectInterface {
@@ -115,7 +120,7 @@ const getItemPlainName = (item: DriveItemData | AdvancedSharedItem) => {
 };
 
 async function getFileHmacFromShardHashes(fileKey: Buffer, shardHashes: string[]): Promise<string> {
-  return getSha512Combined(fileKey, Buffer.from(shardHashes.join(''), 'hex'));
+  return hmacSha512(fileKey, Buffer.from(shardHashes.join(''), 'hex'));
 }
 
 export {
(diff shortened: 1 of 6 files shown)


larryrider (dismissed):

CandelR (dismissed):

src/app/network/NetworkFacade.ts
```diff
@@ -297,9 +299,10 @@ export class NetworkFacade {
         );
 
         fileStream = buildProgressStream(
-          decryptedStream, 
+          decryptedStream,
           (readBytes) => options?.downloadingCallback?.(fileSize, readBytes),
           async () => {
+            const shardHashes = await Promise.all(sha256Hashes.map(getRipemd160FromHex));
             const hmac = await getFileHmacFromShardHashes(key as Buffer, shardHashes);
```
TamaraFinogina, line 306: Can we here derive a different key for HMAC here to make sure that encryption and HMAC do not share the same key?

Something like:

import { blake3 } from '@noble/hashes/blake3.js';
const keyHmac = blake3(key as Buffer, { context: txt.encode('hmac-key-for-file') });

or 

const keyHmac = await hmacSha512(key, Buffer.from('hmac-key-for-file'));

sg-gs: Done

src/app/crypto/services/utils.ts
```diff
@@ -115,7 +130,8 @@ const getItemPlainName = (item: DriveItemData | AdvancedSharedItem) => {
 };
 
 async function getFileHmacFromShardHashes(fileKey: Buffer, shardHashes: string[]): Promise<string> {
-  return getSha512Combined(fileKey, Buffer.from(shardHashes.join(''), 'hex'));
+  const hmacKey = await deriveHmacKey(fileKey);
```
TamaraFinogina, line 133: Great! It's not going to be derived every time, right? Just once?

TamaraFinogina (approved):

src/app/crypto/services/utils.ts
sg-gs: Yes, just once

larryrider (approved):

PR #2005, opened 2026-06-23
[PB-6529]:feat/adding .txt files previewable view

jaaaaavier:
## Description
This PR adds preview functionality for .txt files that users may have; the requirements are that the text be rendered and be readable. There are also requests for features such as support for large files: up to ~5–10 MB (lazy load beyond that). This is being addressed; we will load in 2MB increments, and when there are 300px left to reach the end, I have tried to use something similar to how it’s done in the XLSX preview. 

## Checklist

- [ ] QA Passed

## Testing Process
I uploaded a small file and checked that it renders correctly; I also uploaded a 60MB text file and verified that it renders correctly, the load times are fine.
Captura de pantalla 2026-06-23 a las 10 58 56


src/app/drive/components/FileViewer/viewers/FileTxtViewer/FileTxtViewer.tsx
```diff
@@ -0,0 +1,61 @@
...
+    readChunk(0);
+  }, [blob]);
+
+  const handleScroll = (event: React.UIEvent<HTMLDivElement>) => {
+    if (!hasMore || loadingRef.current) return;
+    const { scrollTop, clientHeight, scrollHeight } = event.currentTarget;
+    if (scrollTop + clientHeight >= scrollHeight - SCROLL_THRESHOLD_PX) {
+      readChunk(offset);
+    }
+  };
```
xabg2: What about using the `IntersectionObserver` instead?
Something like:
```
useEffect(() => {
    const sentinel = sentinelRef.current;
    if (!sentinel) return;
    const observer = new IntersectionObserver(
      (entries) => {
        if (entries[0].isIntersecting) readChunk();
      },
      { rootMargin: '300px' },
    );
    observer.observe(sentinel);
    return () => observer.disconnect();
  }, [readChunk]);
```

And then adding a `<div ref={sentinelRef} />`

I did not test the code but, maybe it can work and is cleaner, WDYT?

jaaaaavier: Yeah, that's a better idea to be honest, I didn't remember about IntersectionObserver. Change applied.

xabg2 (dismissed):

xabg2 (dismissed):

jaaaaavier:
We updated the bg-color of the viewer for a correct vision in dark mode

xabg2 (dismissed):

src/app/drive/types/file-types.ts
```diff
@@ -191,6 +191,7 @@ const txtExtensions: FileExtensionMap = {
   txt: ['txt', 'text', 'conf', 'def', 'list', 'log', 'md', 'lock'],
 };
 
+const previewableTxtExtensionGroups: string[] = ['txt'];
```
larryrider: It would be nice to extend this to other common plain text file types as well, not just .txt. For example: .json, .md, .xml....

This could be a list that should work the same:
.txt
.md
.rst
.adoc

Makefile
Dockerfile
.gitignore
.gitattributes
.editorconfig
.npmrc
.yarnrc
.prettierrc
.eslintrc
.babelrc
.env
.env.example

.json
.jsonc
.yaml
.yml
.toml
.ini
.cfg
.conf
.properties
.xml
.csv
.tsv

.js
.ts
.jsx
.tsx
.mjs
.cjs
.py
.go
.rs
.java
.kt
.swift
.c
.cpp
.h
.hpp
.cs
.php
.rb
.sh
.bash
.zsh
.ps1
.bat
.cmd

.html
.htm
.css
.scss
.less
.sql

.log
.patch
.diff
.gitmodules
LICENSE
README
CHANGELOG
CONTRIBUTING
CODE_OF_CONDUCT
AUTHORS
NOTICE


what do you think???

src/app/drive/types/file-types.ts
```diff
@@ -191,6 +191,7 @@ const txtExtensions: FileExtensionMap = {
   txt: ['txt', 'text', 'conf', 'def', 'list', 'log', 'md', 'lock'],
```
larryrider: you even have some of them here 👀

larryrider (dismissed):

src/app/drive/types/file-types.ts
jaaaaavier: Yes thats a great idea and it fits with this PR. I'll prepare

jaaaaavier:
Added a lot more plainText type available for preview

larryrider (approved):

PR #2006, opened 2026-06-23
[PB-6532] fix pwd share key

TamaraFinogina:
## Description

Fix for public sharing with pwd 

## Checklist

- [ ] QA Passed

## Testing Process

QA in local

### DIFF
--- a/src/app/network/NetworkFacade.ts
+++ b/src/app/network/NetworkFacade.ts
@@ -13,7 +13,13 @@ import { buildProgressStream, decryptStream } from 'services/stream.service';
 import { WORKER_MESSAGE_STATES } from 'app/drive/services/worker.service/types/upload';
 import { waitForContinueUploadSignal } from 'app/drive/services/worker.service/uploadWorkerUtils';
 import { TaskStatus } from '../tasks/types';
-import { createSha256HashingStream, encryptStreamInParts, generateFileKey, getEncryptedFile, processEveryFileBlobReturnHash } from './crypto';
+import {
+  createSha256HashingStream,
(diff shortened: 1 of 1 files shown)


sg-gs:
> Testing Process
> QA in local

Hey @TamaraFinogina specify what are you testing and how should we. You can add those ACs to the Jira task directly, which is better

src/app/network/NetworkFacade.ts
```diff
@@ -303,9 +309,10 @@ export class NetworkFacade {
           (readBytes) => options?.downloadingCallback?.(fileSize, readBytes),
           async () => {
             const shardHashes = await Promise.all(sha256Hashes.map(getRipemd160FromHex));
-            const hmac = await getFileHmacFromShardHashes(key as Buffer, shardHashes);
-            if (fileInfoRef.hmac?.value && hmac !== fileInfoRef.hmac.value) throw new Error('File integrity check failed');
-          }
+            const hmac = await getFileHmacFromShardHashes(options?.key || (key as Buffer), shardHashes);
```
sg-gs, line 312: Why?

TamaraFinogina: Because it has to be the same key as for encryption. When we download a file shared with a password, then hmac fails because the key in the options and the key as a Buffer are different.

TamaraFinogina:
@sg-gs, I notice the issue in prod and put videos (before and after) in the Slack chat. Can create an issue too

sg-gs (approved):
It's ok @TamaraFinogina

PR #2009, opened 2026-06-25
[PB-6503] Add download new sharings

TamaraFinogina:
## Description

Add bucketKey option to download to support the new sharing version (downloading of folders shared with the public sharing version "inxt-v3").

Changes: 

- add downloadOwnFileWithBucketKey function and refactor params to fit the old model
- use NetworkCredentials from the same file instead of redefining it in multiple files
- change mnemonic to FileKey = {mnemonic or bucketKey}
- refactor NetworkFacade to reduce code repetitions

The majority of deletes are coming from the download test, which was defining the same input in each test instead of declaring it once

## Related Issues

[PB-6503](https://inxt.atlassian.net/browse/PB-6503)

## Related Pull Requests

- [Bult on top of add_new_sharings](https://github.com/internxt/drive-web/pull/1996)

## Checklist

- [ ] QA Passed

## Testing Process

- Upload a new file/folder (or find a file/folder that has not been previously shared), and share it via public sharing, then download. The download should finish without error, and the file/folder should be viewable.  
- Upload a new file/folder (or find a file/folder that has not been previously shared), and share it via public sharing with PWD, then download. The download should finish without error, and the file/folder should be viewable.  
- Previously shared file/folder (version "inxt-v2") should still be downloadable and viewable.
- File/folder downloading from the drive should not be affected.
- Sharing a file/folder with an existing drive user should not be affected.
- Invitations to the workspace should not be affected.
- Backup should not be affected.

**WARNING:** Must be tested for files and folders separately as they have different underlying workflows.

[PB-6503]: https://inxt.atlassian.net/browse/PB-6503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/network/download/v2.ts
```diff
@@ -68,8 +73,9 @@ async function getAuthFromCredentials(creds: NetworkCredentials): Promise<{ user
 }
 
 const downloadOwnFile: DownloadOwnFileFunction = async (params) => {
-  const { bucketId, fileId, mnemonic, options } = params;
+  const { bucketId, fileId, key, options } = params;
   const auth = await getAuthFromCredentials(params.creds);
+  const mnemonic = key.mnemonic ?? '';
```
sg-gs: This should never be empty, why is this possible?

src/app/network/download/v2.ts
```diff
@@ -89,9 +95,36 @@ const downloadOwnFile: DownloadOwnFileFunction = async (params) => {
   });
 };
 
+const downloadOwnFileWithBucketKey: DownloadOwnFileFunction = async (params) => {
+  const { bucketId, fileId, key, options } = params;
+  const auth = await getAuthFromCredentials(params.creds);
+  const bucketKey = key.bucketKey;
+  if (!bucketKey) {
+    throw new Error('DOWNLOAD ERRNO. Innvalid Bucket key');
```
sg-gs: Typo here, also, you should add a number to the error

src/app/network/download/v2.ts
```diff
@@ -89,9 +95,36 @@ const downloadOwnFile: DownloadOwnFileFunction = async (params) => {
...
+    downloadingCallback: options?.notifyProgress,
+    abortController: options?.abortController,
+  });
+};
+
 export async function multipartDownload(params: DownloadOwnFileParams & { fileSize: number }): Promise<FileStream> {
-  const { bucketId, fileId, mnemonic, fileSize, options } = params;
+  const { bucketId, fileId, key, fileSize, options } = params;
   const auth = await getAuthFromCredentials(params.creds);
+  const mnemonic = key.mnemonic ?? '';
```
sg-gs: Same here. Shouldn't we throw an error?

src/app/network/types/helper-types.ts
```diff
@@ -0,0 +1,9 @@
+export type FileKey = {
```
sg-gs, line 1: These are network-related types, they should be inside the `network` folder

src/app/network/download/v2.ts
TamaraFinogina: because I changed the mnemonic for FileKey, and it has a mnemonic as optional

src/app/network/types/helper-types.ts
TamaraFinogina: Done, moved to src/app/network/download/v2.ts

src/app/network/download/v2.ts
TamaraFinogina: Done, fixed with types

src/app/network/download/v2.ts
TamaraFinogina: Done, fixed with types

src/app/network/download/v2.ts
TamaraFinogina: Done, fixed with types

sg-gs (dismissed):
LGTM, as long as @ValentynaBlahodyrQA passes the integrity checks QA on the related Jira task for the sharing scenarios (retroactive sharings and the new ones, if apply)

TamaraFinogina:
@sg-gs We can merge with [add_new_sharings](https://github.com/internxt/drive-web/tree/add_new_sharings) branch and prepare QA there, to have a single PR

sg-gs:
Try to not be >500 LOCs per unique PR @TamaraFinogina. It's better to add downloads first (which do not break nothing) and do the biggest part of the QA on uploads (but still test that things here are not broken for current sharings)

TamaraFinogina:
@sg-gs, we cannot merge this PR first because it's on top of the sharings. And without sharing, it's not possible to check that the download path for bucketKey works

sg-gs:
It's not possible, but it's possible to review and check that actual things do not break with these changes. That's the point. The HMAC was introduced like that: you may have the code locally of the sharing creation to test this part but split those parts to avoid increasing the error surface without a reason. First introduce the download, which should not break anything of the current production environment. 

I mean, the code should be introduced in an acceptable sizing, even if it requires multiple steps. Thus, with this PR you get the half of the changes already in production and with the retroactive sharings working already tested. 

Then you introduce a PR with the sharing-creation changes, where a bigger QA may happen. That way you design and propose changes on a way that is manageable, faster and easier to review for everyone 👍🏽

TamaraFinogina:
@sg-gs Ok, I'm rebasing it on master then

auto-rebase failing, will try to resolve it

TamaraFinogina:
Changes are in another PR https://github.com/internxt/drive-web/pull/2010 that points to master

PR #2011, opened 2026-06-29
[PB-6503] Use localstorage functions to get and set token

TamaraFinogina:
## Description

This PR changes all calls for setting/getting a new user token to use local storage functions. It's a preparation for a fix for [PB-6503](https://inxt.atlassian.net/browse/PB-6503)

## Related Issues

Relates to [PB-6503](https://inxt.atlassian.net/browse/PB-6503)

## Checklist

- [x] Changes have been tested locally.
- [x] Unit tests have been written or updated as necessary.
- [ ] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [ ] SonarCloud issues have been reviewed and addressed.
- [ ] QA Passed

## Testing Process

Unit tests, small QA to ensure login, sign up and downlod keep working

[PB-6503]: https://inxt.atlassian.net/browse/PB-6503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2012, opened 2026-06-29
[PB-6505] Remove mnemonic from local storage service

TamaraFinogina:
## Description

Removes mnemonic from local storage. In almost all cases, we took the mnemonic from the user. 

## Related Issues

Relates to [PB-6503](https://inxt.atlassian.net/browse/PB-6503)

## Related Pull Requests

- [Use functions for get/set NewUserToken](https://github.com/internxt/drive-web/pull/2011)

## Checklist

- [ ] QA Passed

## Testing Process

QA to check login, sign up and back up

[PB-6503]: https://inxt.atlassian.net/browse/PB-6503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/views/Login/components/LogIn.tsx
```diff
@@ -46,7 +46,7 @@ export default function LogIn(): JSX.Element {
   const [showErrors, setShowErrors] = useState(false);
 
   const user = useSelector((state: RootState) => state.user.user) as UserSettings;
-  const mnemonic = localStorageService.get(LocalStorageItem.UserMnemonic);
+  const mnemonic = user?.mnemonic;
```
sg-gs, line 49: Is this value decrypted at this point? I recall seeing the mnemonic still encrypted in the xUser value (which is the user here, if I am not wrong)

src/views/Signup/hooks/useGuestSignupState.ts
```diff
@@ -20,7 +18,7 @@ export const useGuestSignupState = () => {
   const [showPasswordIndicator, setShowPasswordIndicator] = useState(false);
 
   const user = useSelector((state: RootState) => state.user.user) as UserSettings;
-  const mnemonic = localStorageService.get(LocalStorageItem.UserMnemonic);
+  const mnemonic = user?.mnemonic;
```
sg-gs, line 21: Same question here

src/utils/backupKeyUtils.ts
```diff
@@ -41,8 +40,8 @@ export interface BackupData {
  * @throws {Error} Implicitly throws if file saving fails
  */
 export function handleExportBackupKey(translate) {
-  const mnemonic = localStorageService.get(LocalStorageItem.UserMnemonic);
   const user = localStorageService.getUser();
+  const mnemonic = user?.mnemonic;
```
sg-gs, line 44: Same here

src/views/Login/components/LogIn.tsx
TamaraFinogina: Not defined:

Screenshot 2026-06-30 at 11 09 56
Screenshot 2026-06-30 at 11 10 31

src/views/Signup/hooks/useGuestSignupState.ts
TamaraFinogina: Also undefined:
Screenshot 2026-06-30 at 11 18 06
Screenshot 2026-06-30 at 11 18 16

src/utils/backupKeyUtils.ts
TamaraFinogina: It's the same:
Screenshot 2026-06-30 at 11 21 53
Screenshot 2026-06-30 at 11 22 03

src/views/Login/components/LogIn.tsx
sg-gs: If it is not defined at that moment, why are we using it?

src/views/Signup/hooks/useGuestSignupState.ts
sg-gs: If it is not defined at that moment, why are we using it?

TamaraFinogina: No idea, but it seems to be triggered when the login form appears. Maybe @larry-internxt o @xabg2 know

xabg2: This is used when the user already has an existing account. In the parent component (e.g. `ShareGuestSignupView`), we check whether the user is already authenticated and redirect them to the product, since it doesn't make sense to display the sign-up flow for someone who already has an active account.

More context: https://github.com/internxt/drive-web/blob/master/src/views/Signup/ShareGuestSignUpView.tsx#L78

TamaraFinogina: @sg-gs I checked, and if the user is already logged in, then those values are defined and identical

src/views/Login/components/LogIn.tsx
sg-gs: This point is still not clear @TamaraFinogina

TamaraFinogina: Same as for GuestSignUp. We use it to redirect users who are already logged in when they end up on the login page. In this case, mnemonic and user. mnemonic are identical

sg-gs (approved):

PR #2020, opened 2026-07-02
[PB-6622] Minimize secret data we store

TamaraFinogina:
## Description

This PR minimizes the data we store. We used to store the entire workspace (type WorkspaceData), but mostly used it for ID and mnemonic. Unfortunately, WorkspaceData has phoneNumber, email, and address - all sensitive fields we never use. 

It got highlighted by Git Security once before, but it got fooled by the switch from setItem to set. 

## Related Issues

Relates to [PB-6505](https://inxt.atlassian.net/browse/PB-6505)

## Related Pull Requests

- [Remove old token](https://github.com/internxt/drive-web/pull/2015)

## Checklist

- [x] Changes have been tested locally.
- [x] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [x] SonarCloud issues have been reviewed and addressed.
- [x] QA Passed

## Testing Process

- Ensure that the workspace is displayed correctly.
- Check that files can be uploaded to the business workspace and downloaded.
- Check that switching between individual and business workspace works.

## Additional Notes

The only line I'm not fully sure about is line 199 of src/app/store/slices/workspaces/workspacesStore.ts:

`   dispatch(workspacesActions.setSelectedWorkspace(selectedWorkspace));`

[PB-6505]: https://inxt.atlassian.net/browse/PB-6505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved B2B workspace handling so the app consistently remembers the selected workspace and related settings.
  * Fixed workspace configuration and trash views to use the correct workspace identifier, reducing errors when switching between personal and workspace contexts.
  * Updated workspace sign-out/reset behavior to fully clear stored workspace data.

PR #2023, opened 2026-07-06
[PB-6505] Add encryption functions for local-storage

TamaraFinogina:
## Description

This PR adds encryption/decryption methods with CryptoKey stored in IndexedDB. So far, the methods are not called.

## Related Issues

## Related Pull Requests

[Minimize workspace data](https://github.com/internxt/drive-web/pull/2020)

## Checklist

- [ ] QA Passed

## Testing Process

unit tests

## Additional Notes

Uint8Array.toBase64() and Uint8Array.fromBase64() required an update of typescript to a more modern version, which in turn required changes in config (update deprecated fields)

### DIFF
--- a/src/app/core/types.ts
+++ b/src/app/core/types.ts
@@ -161,6 +161,14 @@ export enum LocalStorageItem {
   AnniversaryTheme = 'anniversary_theme_enabled',
 }
 
+export enum LocalStorageProtectedItem {
+  User = 'xUser',
+  NewToken = 'xNewToken',
+  B2Bworkspace = 'b2bWorkspace',
+  WorkspaceCredentials = 'workspace_credentials',
+  FolderAccessToken = 'folderAccessToken',
+  FileAccessToken = 'fileAccessToken',
+}
 export enum OrderDirection {
   Asc = 'ASC',
   Desc = 'DESC',
--- a/src/services/local-storage-crypto.ts
+++ b/src/services/local-storage-crypto.ts
@@ -5,48 +5,19 @@ import {
   FailedToFindKey,
   KeyAlreadyExistsError,
 } from './local-storage-errors';
-import { DB_NAME, DB_VERSION, STORE, KEY_ID, KEY_LENGTH, IV_LENGTH, ALGORITHM } from './local-storage-constants';
-
-function openDb(): Promise<IDBDatabase> {
-  return new Promise((resolve, reject) => {
-    const req = indexedDB.open(DB_NAME, DB_VERSION);
-    req.onupgradeneeded = () => {
-      const db = req.result;
-
-      if (!db.objectStoreNames.contains(STORE)) {
-        db.createObjectStore(STORE);
-      }
-    };
-    req.onsuccess = () => resolve(req.result);
-    req.onerror = () => reject(req.error ?? new Error('Request to open database failed'));
-  });
-}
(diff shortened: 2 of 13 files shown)


tsconfig.json
```diff
@@ -11,13 +13,12 @@
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
sg-gs, line 16: Is this OK for you @CandelR @larry-internxt ? I do not know the collateral effects

src/app/core/types.ts
```diff
@@ -161,6 +161,14 @@ export enum LocalStorageItem {
   AnniversaryTheme = 'anniversary_theme_enabled',
 }
 
+export enum LocalStorageProtectedItem {
+  User = 'xUser',
+  NewToken = 'xNewToken',
+  B2Bworkspace = 'b2bWorkspace',
+  WorkspaceCredentials = 'workspace_credentials',
```
sg-gs, line 167: Use camelCase here, as you do with the rest of the keys

src/services/local-storage-crypto.ts
sg-gs, line 1: Why not using the already-existing LevelDB [layer](https://github.com/internxt/drive-web/blob/ae2ae979b19a1d5ca91be22e8b3a6ca72d5e92eb/src/app/database/services/database.service/indexed-db.service.ts#L1)? @TamaraFinogina

src/app/core/types.ts
TamaraFinogina: It's an already existing key. I just copied it to another type. Can change it, but it might affect already logged-in users.

tsconfig.json
larryrider: For me yes, this should only affect to the TypeScript's module resolution for type checking, not the actual build (Vite handles that). And bundler matches how Vite actually resolves modules, so maybe its actually more accurate than node. Any issue would surface as a compile time type error in CI, not at runtime

src/services/local-storage-crypto.ts
TamaraFinogina: @sg-gs Didn't see it. Switched to the existing layer.

src/services/local-storage-crypto.ts
```diff
@@ -0,0 +1,68 @@
...
+  return databaseService.delete(DatabaseCollection.CryptoKeys, KEY_ID);
+}
+
+export async function createNewKey(): Promise<CryptoKey> {
+  const existing = await getKey();
+  if (existing) {
+    throw new KeyAlreadyExistsError('A key already exists; clear storage before creating a new one');
+  }
+  try {
+    const key = await crypto.subtle.generateKey({ name: ALGORITHM, length: KEY_LENGTH }, false, ['encrypt', 'decrypt']);
```
sg-gs, line 25: Why `extractable` is not being used?

TamaraFinogina: It's used as a second parameter and set to false

generateKey(algorithm, extractable, keyUsages)

sg-gs: I wrongly thought that `false` meant not using it, but means the possibility of doing so, therefore, the desired value is `false`, as you proposed, my bad.

sg-gs (approved):

PR #2024, opened 2026-07-06
[PB-6505] Use set user

TamaraFinogina:
## Description

Removes the unused email field and switches to the setUser function instead of setting it directly in local storage.

## Related Issues

## Checklist

- [ ] QA Passed

## Testing Process

unit tests

## Additional Notes

setUser/getUser will be needed for adding encryption


src/services/local-storage.service.ts
```diff
@@ -57,6 +56,11 @@ function getUser(): UserSettings | null {
   return stringUser ? JSON.parse(stringUser) : null;
 }
 
+function setUser(user: UserSettings): void {
+  set(LocalStorageItem.UserUUID, user.uuid);
```
larryrider: Why saving the user uuid in a new localStorage var, if the user is still going to be saved, and the uuid is already there? 🤔

TamaraFinogina: Because the user will be encrypted, so, getting back up keys (getBackupKeyStorageKeys()) will require 1 decryption. Yet, it only needs the UUID, which is not secret and can be stored without any encryption at all.

TamaraFinogina: @larryrider I removed it at the end

larryrider (approved):

PR #2029, opened 2026-07-07
[PB-6505] make getToken async

TamaraFinogina:
## Description

This PR includes all changes related to making getToken and only getToken async. No encryption added, just the function is turned into async to match what the getToken with decryption will do.

## Related Issues

## Checklist

- [ ] QA Passed

## Testing Process

Check all main functionalities of the drive web as the token was everywhere.

## Additional Notes

I'm not sure this is the best way. Maybe we can compromise and store the token in the cache, sessionStorage, etc. Because there are lots of changes.

### DIFF
--- a/src/app/core/factory/sdk/index.ts
+++ b/src/app/core/factory/sdk/index.ts
@@ -47,78 +47,78 @@ export class SdkFactory {
     return this.sdk.newApiInstance;
   }
 
-  public createAuthClient(options?: { captchaToken?: string; unauthorizedCallback?: () => void }): Auth {
+  public async createAuthClient(options?: { captchaToken?: string; unauthorizedCallback?: () => void }): Promise<Auth> {
     const apiUrl = this.getApiUrl();
     const appDetails = this.getAppDetailsWithHeaders(options?.captchaToken);
-    const apiSecurity = this.getNewApiSecurity(options?.unauthorizedCallback);
+    const apiSecurity = await this.getNewApiSecurity(options?.unauthorizedCallback);
     return Auth.client(apiUrl, appDetails, apiSecurity);
   }
 
-  public createDesktopAuthClient(): Auth {
+  public async createDesktopAuthClient(): Promise<Auth> {
     const apiUrl = this.getApiUrl();
     const appDetails = SdkFactory.getDesktopAppDetails();
-    const apiSecurity = this.getNewApiSecurity();
+    const apiSecurity = await this.getNewApiSecurity();
     return Auth.client(apiUrl, appDetails, apiSecurity);
   }
 
-  public createNewStorageClient(): Storage {
+  public async createNewStorageClient(): Promise<Storage> {
     const apiUrl = this.getApiUrl();
     const appDetails = SdkFactory.getAppDetails();
-    const apiSecurity = this.getNewApiSecurity();
+    const apiSecurity = await this.getNewApiSecurity();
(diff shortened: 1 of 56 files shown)


src/services/referral.service.ts
```diff
@@ -301,7 +302,8 @@ const isEligibleForReferral = async (accountCreatedAt?: Date): Promise<boolean>
   }
 
   try {
-    const { isEnabled } = await getReferralsClient().isReferralEnabled();
+    const referalCklient = await getReferralsClient();
```
sg-gs: Typo here

sg-gs (commented):
You may need to test the OAuthFlow and the VPN Auth thing, which seem to be the affected parts in terms of functionality

src/services/referral.service.ts
TamaraFinogina: Fixed

src/services/local-storage.service.ts
```diff
@@ -71,8 +71,8 @@ function getUser(): UserSettings | null {
   return stringUser ? JSON.parse(stringUser) : null;
 }
 
-function getToken(): string | null {
-  return get(LocalStorageItem.NewToken);
+async function getToken(): Promise<string | undefined> {
+  return get(LocalStorageItem.NewToken) ?? undefined;
```
xabg2, line 75: Why?

src/app/core/factory/sdk/index.ts
```diff
@@ -177,8 +178,8 @@ export class SdkFactory {
     };
   }
 
-  private getNewToken(): Token {
-    return SdkFactory.sdk.localStorage.getToken() || '';
+  private async getNewToken(): Promise<Token> {
+    return (await SdkFactory.sdk.localStorage.getToken()) || '';
```
xabg2, line 182: Why? `localStorage` is not asynchronous and it is initialized when opening the app. Also, `localStorage.get(...)` is not a Promise. 🤔

xabg2: Maybe this can be async so you can decrypt/do whatever needs here.

CandelR (commented):
Perhaps, instead of doing it this way, I would create an additional layer of abstraction so as not to couple the crypto API with local storage. I would create a secure storage solution that encrypts and decrypts the values you want to store in encrypted form, so that the local storage service remains as it was. You can remove the sensitive keys and move them to the secure storage. 
Once you’ve created this extra layer, you can then delete the keys from the local storage service that need to be stored in the secure storage.
If I haven’t explained myself clearly, we can discuss it via DM :)

src/services/local-storage.service.ts
TamaraFinogina: Because when I switch to the encrypted version, I'll need this function to be async. Wanted to do everything in 1 go at first, but async alone brings way too many changes

src/app/core/factory/sdk/index.ts
TamaraFinogina: It will become async once we switch to an encrypted token and switch from get (sync) to getAndDecrypt (async). This PR is just to do the async change, because adding encryption here on top of all changes will be too much, and we won't be able to review it properly

src/services/local-storage.service.ts
xabg2: I agree with Ramon, I think we should have an extra layer for crypto stuff. Because we are mixing things in a service that only get/set data in local-storage.

PR #2030, opened 2026-07-08
[PB-6600]: decouple file upload manager from task tracking

jaaaaavier:
## Description

Inverts the dependency between the file upload manager and the tasks system:

- `UploadManager` now only handles data transfer. It emits neutral upload events
  (`onUploadStart`, `onUploadProgress`, `onUploadSuccess`, `onUploadError`,
  `onUploadAborted`) and no longer imports `tasksService`, `i18next` or
  `referralService`. It receives files with a pre-assigned `taskId` and treats it
  as an opaque correlation id.
- New tasks layer `uploadFilesWithTasks` creates (or resumes) one task per file,
  subscribes to the manager events and owns every `tasksService` interaction,
  including pause/cancel gating, folder-related progress, error subtitles and
  referral/network tracking.
- `uploadItemsThunk` now calls the tasks layer instead of the manager, so the
  flow is: component → thunk → tasks layer → upload manager.

## Checklist

- [ ] QA Passed

## Testing Process
I’ve tested the preview version for uploading new files and folders, and there were no problems.


src/app/tasks/upload/uploadFolderTaskLifecycle.ts
sg-gs, line 1: The network folder is for the network things (I am a code owner of it) therefore, these kind of things should be outside, I presume, in the corresponding service (I recall this project having folder/file services). Therefore, only if you change what is related to the network, I am requested as a reviewer.

As a rule of thumb: if the code contains tracking functions, logger-task-related functions and so on, it should be outside of here @jaaaaavier

jaaaaavier: Understood

src/app/network/UploadManager.ts
```diff
@@ -180,32 +169,20 @@ class UploadManager {
...
+            const isTaskCancelled = taskStatus === TaskStatus.Cancelled;
 
             if (!isTaskCancelled && !isTaskPaused) {
-              tasksService.updateTask({
-                taskId: taskId,
-                merge: {
-                  progress: uploadProgress,
-                },
-              });
+              this.taskLifecycle.updateTaskProgress(taskId, uploadProgress);
```
sg-gs: Again, it's not the responsibility of the upload manager to manage the tasks lifecycle. The upload manager may have events/callbacks like `onFileUploaded` or `onFolderUploaded` that could be used to manage tasks by injecting specific callbacks, but that should happen _outside_ of here. 

Right now, the control flow is
 
The component starts the upload
  -> The upload service (say folder or file, whatever)
       -> Upload Manager (handles the data transfer)
           -> Tasks Management (handles UI)

The control flow should be 

The component starts the upload
  -> The upload service (say folder or file, whatever)
       -> Tasks Management (handles UI)
       -> Upload Manager (handles the data transfer)

There is still a dependency between two unrelated things, making the code coupled and far less maintainable (and testable): the tasks management and the upload manager. 

Think it like this: if you had to change something related to the tasks management, does it have sense to go to the upload manager to make changes? It sounds like something strange is going on there. If you want to change something related to tasks, you go to the tasks management related-code only.

Consider this [reading](https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html) @jaaaaavier, as although for this specific PR and topic seems trivial, it is worth a lot in the medium and long term for our growing codebase

terrerox: @sg-gs I split this refactor into 5 stacked PRs:

1. This PR — file upload pipeline
2. #2061 — file tasks layer tests
3. #2062 — UploadManager tests
4. #2063 — folder upload pipeline
5. #2064 — folder tests

src/app/network/UploadManager.ts
```diff
@@ -131,17 +123,10 @@ class UploadManager {
...
-        taskId,
-        merge: {
-          status: TaskStatus.Encrypting,
-          stop: async () => {
-            if (this.abortController) this.abortController?.abort();
-            else fileData?.abortController?.abort();
-          },
-        },
+
+      this.events?.onUploadStart?.(fileData, async () => {
```
sg-gs, line 127: Why is this `async` and why if the abort controller exists, it is aborted, as this event is the start of an upload (?)

sg-gs (commented):
This code is far cleaner and respects responsibilities, I only have that doubt there @terrerox

src/app/network/UploadManager.ts
terrerox: That function isn't called here,  it's the stop callback we hand over to the tasks layer, which saves it as the task's `stop`. It only runs if the user cancels the upload later, nothing gets aborted at start.

The `async` is just to match the `stop` signature (`() => Promise<void>`), and the if/else covers the two cases: a shared abort controller for the whole batch, or one per file.

sg-gs (dismissed):

terrerox:
I merged the rest of the PRs and lost your approval, @sg-gs

sg-gs (approved):

PR #2031, opened 2026-07-08
[PB-6603] Encrypt token

TamaraFinogina:
## Description

Starts encrypting the token, separates local and encrypted storages 

## Related Issues

## Checklist

- [x] QA Passed

## Testing Process

Full QA of all main drive web functionalities, token change touches everything (should be tested for logged-out users and those who have an ongoing session and token in local storage)

## Additional Notes

Includes the migration step that should be removed once the encrypted token is in prod for more than its lifetime.


src/app/core/factory/sdk/index.ts
```diff
@@ -30,10 +32,15 @@ export class SdkFactory {
     this.apiUrl = apiUrl;
   }
 
-  public static initialize(dispatch: AppDispatch, localStorage: LocalStorageService): void {
+  public static initialize(
+    dispatch: AppDispatch,
+    localStorage: LocalStorageService,
+    encryptedStorage: EncryptedStorageService,
```
sg-gs, line 38: Q: is not cleaner and easier to modify the `getToken` method to decrypt and the `setToken` to encrypt the content, keeping the security layer at the storage level ? (the local storage)

TamaraFinogina: It was cleaner to separate. Encrypted storage has a token (and will have other sensitive variables) at the module level, so it's better to give it only to methods that truly need it.  Plus, after full migration, the SDK will only accept encrypted storage (so far, I changed only tokens, and workspace tokens are still taken from local storage).

CandelR: I suggested doing it this way so as not to tie the cryptography layer to a service that purely uses local storage. That way, we don’t alter the behaviour of the service, which is purely about accessing local storage, and we can work with the cryptography at a higher level of abstraction. This also allows us to keep storage requiring encryption separate from that which works fine as it is, so that, should the system grow, we can manage it in different locations and make it more maintainable

sg-gs (dismissed):

src/services/encrypted-storage.service.test.ts
```diff
@@ -0,0 +1,36 @@
...
+beforeEach(() => {
+  vi.restoreAllMocks();
+  localStorage.clear();
+});
+
+describe('Testing the local storage service', () => {
+  describe('Get and set encrypted values', () => {
+    const value = 'test-value';
+
+    it('When sets protected value, then the value is stored encrypted', async () => {
```
larryrider: use `test` instead of `it`

src/services/local-storage-crypto.ts
```diff
@@ -8,6 +8,16 @@ import {
...
 import databaseService, { DatabaseCollection } from 'app/database/services/database.service';
 
+async function ensureKey(): Promise<CryptoKey> {
+  const existing = await getKey();
+  if (existing) return existing;
+  else return createNewKey();
+}
+
+export async function ensureKeyExists(): Promise<void> {
+  await ensureKey();
```
larryrider, line 18: Why not using the `ensureKey` method from above directly? 🤔 

Remove this one, rename the other one as `ensureKeyExists`, and export it, and it will be the same, right?

TamaraFinogina: because CryptoKey should not be given to local storage

src/services/encrypted-storage.service.test.ts
TamaraFinogina: why? Those tests were moved from local storage tests, and all tests there use it

larryrider: its just a naming convention, it works the same but we use the 'it' for `it(should do ... when ...)` and 'test' for `test(When .... then ....)`

src/services/local-storage-crypto.ts
larryrider: Then maybe its a better idea to just remove this method, and add the `ensureKey` to the `getKey` method. 
That way every time the `getKey` is called we are sure the Key exists, and then we dont need to use the `ensureKeyExists` exported method

wdyt? maybe im missing something 🤔

TamaraFinogina: Not sure about getKeys. We call this method for both get and set. If no keys, then the get should fall with "no keys" instead of "failed to decrypt" as it will happen if we just generate a random new key. The set is already calling ensureKey all the time

src/services/encrypted-storage.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    const unencryptedToken = localStorage.getItem(LocalStorageItem.NewToken);
+    if (unencryptedToken) {
+      await setToken(unencryptedToken);
+      localStorage.removeItem(LocalStorageItem.NewToken);
+    }
+  }
+}
+
+function getToken(): string | undefined {
+  return tokenCache ?? undefined;
```
CandelR: nit: use arrow functions to fit newest codebase style

CandelR (dismissed):

src/services/sockets/socket.service.ts
```diff
@@ -71,5 +71,5 @@ export default class RealtimeService {
 }
 
 function getToken(): string {
-  return localStorageService.getToken() as string;
+  return encryptedStorageService.getToken() ?? '';
```
sg-gs, line 74: Why?

TamaraFinogina: `encryptedStorageService.getToken()` returns `string | undefined`, which, with `undefined`, will still be undefined after `as string`.  The return type would not be correct, so it will be runtime error

src/index.tsx
```diff
@@ -56,44 +57,49 @@ if (typeof Node !== 'undefined' && Node.prototype.removeChild) {
   };
 }
 
-enforceCanonicalDriveDomain();
+async function bootstrap() {
+  await encryptedStorageService.hydrateEncryptedStorageCache();
```
sg-gs, line 61: Does this last / adds latency? Maybe it's better to do this call after the render and not before

src/services/encrypted-storage.service.test.ts
```diff
@@ -0,0 +1,52 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { LocalStorageProtectedItem } from 'app/core/types';
+import encryptedStorageService from './encrypted-storage.service';
+
+beforeEach(() => {
+  vi.restoreAllMocks();
+  localStorage.clear();
+});
+
+describe('Testing the local storage service', () => {
```
sg-gs: Testing the _encrypted_ local storage service, I assume

src/index.tsx
TamaraFinogina: Not much; it's 1 fast decryption

TamaraFinogina: I think I put it on top because one of the functions below required token

src/services/encrypted-storage.service.test.ts
TamaraFinogina: Fuxed

src/services/encrypted-storage.service.ts
TamaraFinogina: Fixed

xabg2 (dismissed):
LGTM (Only checked the payments-related changes)

src/index.tsx
TamaraFinogina: @sg-gs yes, a token is used during the App creation:

Screenshot 2026-07-28 at 11 24 20

src/services/encrypted-storage.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  const encryptedValue = await encryptEntry(value);
+  localStorage.setItem(key, encryptedValue);
+};
+
+const setToken = async (token: string): Promise<void> => {
+  tokenCache = token;
+  return setAndEncrypt(LocalStorageProtectedItem.EncryptedToken, token);
+};
+
+const hydrateEncryptedStorageCache = async (): Promise<void> => {
```
CandelR, line 21: What happens if the token hydration fails? Does the app get stuck without a token? Perhaps the token should be left null so that any call to the backend fails and the user is automatically logged out

TamaraFinogina: getToken will return undefined, but hydrateEncryptedStorageCache will throw an error (added test to confirm). I can make it a silent error, i.e., only logs, no throw

TamaraFinogina: @CandelR I changed the function so it never throws now; it just sets the cached token to null. Tested by invalidating the token and doing a reload; the app is redirected to the login page, no blank screen

CandelR (approved):

sg-gs (approved):

PR #2033, opened 2026-07-09
[PB-6699] Remove old key fields from UserSettings

TamaraFinogina:
## Description

This PR reduces the amount of data in UserSettings by removing old fields (`privateKey`, `publicKey`, and `revocationKey` outside of the `keys` structure). Will help once we start encrypting UserSettings.

P.S. Most changes are from deleted tests that were for testing that the old key structure works (without the keys field but with privateKey, publicKey, and revocationKey outside of the keys structure)

## Related Issues

[PB-6505](https://inxt.atlassian.net/browse/PB-6505)

## Related Pull Requests

- [Minimize workspace data](https://github.com/internxt/drive-web/pull/2020)
- [Encrypt token](https://github.com/internxt/drive-web/pull/2031)

## Checklist

- [ ] QA Passed

## Testing Process

Check that login, signup, key backup and password change are working.

## Additional Notes

We switched to keys in UserSettings over a year ago. There should be no users with an ongoing session without a Kyber key or keys field.


larryrider (dismissed):

larryrider (dismissed):

src/app/share/services/share.service.ts
```diff
@@ -584,7 +584,7 @@ export const decryptMnemonic = async (encryptionKey: string): Promise<string | u
   const user = localStorageService.getUser();
   if (user) {
     let decryptedKey;
-    const privateKeyInBase64 = user.keys?.ecc?.privateKey ?? user.privateKey;
+    const privateKeyInBase64 = user.keys?.ecc?.privateKey ?? '';
```
sg-gs: If these should be always present, which I pressume yes, this should stop being optional and do not allow these empty values. 

The source where they come from should enforce its presence. If they are not present, an error should be thrown.

sg-gs (changes requested):

src/app/share/services/share.service.ts
TamaraFinogina: It's already obligatory in the UserSettings type. Removed ? marks and parser function

src/views/Signup/hooks/useSignup.ts
```diff
@@ -90,14 +49,19 @@ export function useSignUp(referrer?: string): {
...
-    // TODO: Remove or modify this when the backend is updated to add kyber keys
-    const parsedUser = parseUserSettingsEnsureKyberKeysAdded(user);
-    parsedUser.mnemonic = decryptTextWithKey(parsedUser.mnemonic, password);
+    user.mnemonic = decryptTextWithKey(user.mnemonic, password);
 
-    return { xUser: parsedUser, xToken: token, xNewToken: newToken, mnemonic: user.mnemonic };
+    return {
+      xUser: {
+        ...user,
+        bucket: '',
```
sg-gs: Why is this empty here?

sg-gs:
Nice cleanup here @TamaraFinogina

src/views/Signup/hooks/useSignup.ts
TamaraFinogina: Because the endpoint is not returning any value for the bucket field, and the UserSettings type needs it. Before, it was masked due to a type cast to UserSettings, but TODO is still present:

 // TODO: need to update user type of register to include bucket field

sg-gs (dismissed):

sg-gs (approved):

PR #2040, opened 2026-07-13
[BR-2165]:feat/add multipart download for shared files

jaaaaavier:
## Description
Adds multipart download support to public shared file links.

Changes:
- NetworkFacade.downloadChunk: honor options?.key as an override for the SDK-derived decryption key, mirroring the existing behavior in download(). Needed because shared links have no mnemonic to derive a key from the real key is the link's own encryptionKey.
- download/v2.ts: split multipartDownload into multipartDownloadOwnFile (existing behavior) and a new multipartDownloadSharedFile, dispatching on token && encryptionKey, the same discriminator already used by downloadFile() for the non-multipart case.
- network/index.ts: re-export multipartDownloadFile from the network facade so it's actually reachable from views (it already existed in `download.ts` but wasn't exposed).
- ShareFileView.tsx: use network.multipartDownloadFile when fileInfo.item.size >= MIN_DOWNLOAD_MULTIPART_SIZE`, otherwise keep the regular single-request `downloadFile`.

## Checklist

- [ ] QA Passed

## Testing Process
Manually verified in the browser: shared files above the threshold trigger multiple parallel `206 Partial Content` range requests instead of one single request, and the resulting file downloads and decrypts correctly.


src/app/network/NetworkFacade.test.ts
```diff
@@ -202,6 +202,33 @@ describe('NetworkFacade', () => {
...
+      const mockResponse = { status: 206, body: mockStream };
+      const sdkDerivedKey = Buffer.alloc(32, 0x01);
+      const overrideKey = Buffer.alloc(32, 0x02);
+      const iv = Buffer.alloc(16);
+
+      (global.fetch as any).mockResolvedValue(mockResponse);
+      (downloadFile as any).mockImplementation(
+        async (fileId, bucketId, mnemonic, network, cryptoLib, bufferFrom, downloadCallback, decryptCallback) => {
+          await downloadCallback([{ url: downloadExample }]);
+          await decryptCallback('aes256ctr', sdkDerivedKey, iv, 2048);
```
sg-gs, line 216: Shouldn't this be called with the overrideKey?

terrerox: I dont think so, because that's the purpose of the test: to check if the derived key is overridden by the override key. If we changed that line to the override key, it would be checking the same key — `overrideKey === overrideKey`.

src/app/network/download/v2.ts
```diff
@@ -42,22 +42,29 @@ type DownloadFileFunction = (
...
 
-const downloadSharedFile: DownloadSharedFileFunction = (params) => {
-  const { bucketId, fileId, encryptionKey, token, options } = params;
-
-  return new NetworkFacade(
+/**
+ * Creates a NetworkFacade for the bridge. Pass auth for downloads of the user's own files;
+ * omit it for shared-link downloads, which authenticate through the share token instead.
+ */
+const createNetworkFacade = (auth?: { username: string; password: string }): NetworkFacade =>
```
TamaraFinogina, line 49: Maybe it's easier to remove the shared download from the network than create an empty auth

src/app/network/NetworkFacade.ts
```diff
@@ -415,7 +415,12 @@ export class NetworkFacade {
         }
       },
       async (algorithm, key, iv, fileSize) => {
-        fileStream = decryptStream(encryptedContentStreams, key as Buffer, iv as Buffer, chunkStart);
+        fileStream = decryptStream(
+          encryptedContentStreams,
+          (options?.key || key) as Buffer,
```
TamaraFinogina, line 420: why override with options.key? It wasn't there before

src/app/network/NetworkFacade.test.ts
TamaraFinogina: You added the test because you added the key override to decryptStream. Why was override added to decryptStream?

TamaraFinogina:
Screenshot 2026-08-18 at 10 43 56
I uploaded a 1,2GB video into a business workspace (not workspace owner, invited user), and video streaming was failing

src/app/network/NetworkFacade.ts
terrerox: As this change adds multipart download for shared files, this wasn't needed before, multipart was only used for own files, where the SDK derives the key from the mnemonic. Shared files don't have a mnemonic; the share link provides the key directly as `encryptionKey`, which we pass as `options.key`, so `downloadChunk` needs to honor it. It's the same `options?.key || key` logic the single-part `download()` already had. Own-file downloads never set `options.key`, so nothing changes for them.

src/app/network/download/v2.ts
terrerox: That empty auth was already there before this PR tbh, downloadSharedFile on master did the same thing, I just moved it into createNetworkFacade. And we can't really avoid it, the SDK client requires those fields and for shared files the backend only gives us the key + token, no network creds. The token is what authorizes the download so the auth is ignored anyway.

src/app/network/NetworkFacade.test.ts
sg-gs: Answer this @terrerox pls

TamaraFinogina: @sg-gs This PR was clashing with changes I made for new private sharings. So, I've incorporated multi-part downloads into new private sharings: https://github.com/internxt/drive-web/pull/2116

We can close this PR

TamaraFinogina: P.S. It's working, and server changes were already merged, but QA was postponed due to Mail & Mobile

PR #2038, opened 2026-07-13
[PB-5660] feat: favorites page

victor-ferro:
## Description

Adds a Favorites page that lists the user's favorite folders and files.

- New route /favorites + sidenav entry (views.ts, paths.json, useSidenavNavigation)
- src/views/Favorites/:
  - FavoritesView rendered through the existing DriveExplorer
  - fetchFavoriteFolders / fetchFavoriteFiles services on top of the new storage client endpoints
  - fetchFavoritesThunk populating the storage slice (new favorites file/folder state)
- Translations for the new view in all locales (de, en, es, fr, it, ru, tw, zh)

## Related Issues

Relates to PB-5660

## Checklist

- [x] Changes have been tested locally.
- [x] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [ ] SonarCloud issues have been reviewed and addressed.
- [ ] QA Passed

## Testing Process

- yarn vitest run src/views/Favorites — 2 files, 8 tests passing (against linked SDK)
- ESLint clean on all touched files

## Additional Notes

- ⚠️ Depends on unreleased @internxt/sdk changes (getFavoriteFolders / getFavoriteFiles on the new storage client). Developed against a yarn-linked local SDK; CI typecheck/tests will fail until the SDK version with these endpoints is published and bumped here.

PR #2041, opened 2026-07-14
[PB-5661] feat: favorite toggle from context menu and F shortcut

victor-ferro:
## Description

Allows users to favorite/unfavorite files and folders from the item context menu, plus an `F` keyboard shortcut.

- New menu entry **"Add to Favorites" / "Remove from Favorites"** (star / crossed-out star icon, `F` shortcut label right-aligned) in every single-item menu — right-click and three-dot — across Drive, Recents, Favorites and Shared
- Excluded from multi-selection menus (ambiguous label with mixed selections), workspace menus (favorites API is personal-scoped) and Trash
- `F` with exactly one item selected toggles it; guarded against `⇧F` (new folder), open dialogs and Trash
- Immediate UI update, no reload: `setItemFavorite` service + `toggleFavoriteThunk` patch the item in place and keep the favorites listing in sync (`patchItem`, `addFavorites` / new `removeFavorites` reducer); error toast on failure
- `isFavorite` added to drive/shared item types; translations in all locales

## Related Issues

Relates to PB-5661

## Related Pull Requests

- [PB-5660-favorites-page](https://github.com/internxt/drive-web/pull/2038) — stacked on this one, merge first

## Checklist

- [ ] QA Passed

## Testing Process

- `yarn vitest run src/views/Favorites src/app/store/slices/storage/storage.slice.test.ts src/views/Shared/hooks/useSharedContextMenu.test.ts` — 6 files, 45 tests passing (against linked SDK)
- ESLint clean on all touched files; `yarn check:i18n` in sync
- Manual testing against the real backend: toggle from both menus on files and folders, `F` shortcut, immediate update in Drive and Favorites views, label/icon switching with item state

## Additional Notes

- **Custom icon**: `@phosphor-icons/react` has no crossed-out star (`StarSlash` doesn't exist), so `src/assets/icons/star-slash.svg` was hand-made following Phosphor's stroke style (20×20, `currentColor`) and imported with `?react` like the existing `move.svg`. If design wants a different one, swapping it is trivial — a single SVG referenced from one place.
- ⚠️ Same SDK dependency as #2038: `markFileAsFavorite` / `unmarkFileAsFavorite` / `markFolderAsFavorite` / `unmarkFolderAsFavorite` come from unreleased `@internxt/sdk` changes (developed against a yarn-linked local SDK). CI will fail until that SDK version is published and bumped.


src/app/store/slices/storage/index.ts
```diff
@@ -136,6 +136,11 @@ export const storageSlice = createSlice({
       const favoriteItems = state.favorites.concat(action.payload);
       state.favorites = removeDuplicates(favoriteItems);
     },
+    removeFavorites: (state: StorageState, action: PayloadAction<DriveItemData[]>) => {
+      state.favorites = state.favorites.filter(
+        (item) => !action.payload.find((i) => i.uuid === item.uuid && !!i.isFolder === !!item.isFolder),
```
CandelR: check sonar warning

src/views/Favorites/services/toggleFavorite.service.ts
```diff
@@ -0,0 +1,16 @@
+import { StorageTypes } from '@internxt/sdk/dist/drive';
```
CandelR, line 1: To ensure consistency with other files, the name should be ‘xxxx.service.ts’. 
If we want to change it, we can discuss it, but ideally, even if it seems redundant, they should all have the same name so we don’t get confused

src/views/Shared/components/SharedItemListContainer.tsx
```diff
@@ -229,6 +232,25 @@ const SharedItemListContainer = ({
...
+    'f',
+    (e) => {
+      if (e.shiftKey) return;
+      e.preventDefault();
+      if (!disableKeyboardShortcuts && selectedItems.length === 1) {
+        toggleFavorite(selectedItems);
+      }
+    },
+    [disableKeyboardShortcuts, selectedItems, toggleFavorite],
+  );
```
CandelR: this hotkey function seems duplicated in DriveExplorer.tsx, maybe we should extract the hotkeys to a hook in order to implement the lib in one point and import in the other places

src/views/Favorites/store/toggleFavoriteThunk.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const toggleFavoriteThunk = createAsyncThunk<void, DriveItemData[], { state: RootState }>(
+  'storage/toggleFavorite',
+  async (items: DriveItemData[], { dispatch, rejectWithValue }) => {
+    try {
+      await Promise.all(
+        items.map(async (item) => {
+          const favorite = !item.isFavorite;
+
+          await setItemFavorite(item, favorite);
```
CandelR: As several favorites can be sent at once, using `Promise.all` can mean that if one fails, we display as though everything has failed, when that is not the case. I think we should replace `Promise.all` with `Promise.allSettled`, so that the dispatches are applied only to those that were successful.
If they all fail, we could display a ‘rejected’ thunk (current toast), and if only some fail, perhaps display a ‘fulfilled’ thunk but with a different toast, such as ‘some favourites could not be updated’

src/app/store/slices/storage/index.ts
victor-ferro: Done!

src/views/Favorites/services/toggleFavorite.service.ts
victor-ferro: Done!

src/views/Shared/components/SharedItemListContainer.tsx
victor-ferro: Extracted the duplicated F shortcut into a shared useToggleFavoriteHotkey hook used by both DriveExplorer and SharedItemListContainer. Or were you referring to a hook covering all the useHotkeys we have across the project?

src/views/Favorites/store/toggleFavoriteThunk.ts
victor-ferro: You're right. It's a use case that didn't affect this branch yet because we only allowed sending a single item to this function, but it's true that it's better to do it properly for the future. In fact, we already ran into this case in the PB-5662 branch. Done!

src/views/Shared/components/SharedItemListContainer.tsx
CandelR: No, not for all, for now, let’s stick to this one so we don’t stray from the scope of the task, and we’ll get round to the others later

CandelR (approved):

PR #2042, opened 2026-07-14
[PB-6545]: feature/account status update

xabg2:
## Description

Add a new dialog to warn users when their subscription is scheduled to end in 30 days or less. Also, add a new account status in the Account section (Settings) to display the scheduled cancellation date, so users can see when their subscription will end.

## Checklist

- [ ] QA Passed


src/app/core/layouts/HeaderAndSidenavLayout/HeaderAndSidenavLayout.tsx
```diff
@@ -27,13 +34,33 @@ export default function HeaderAndSidenavLayout(props: HeaderAndSidenavLayoutProp
...
+  const { isReactivatingSubscription, reactivateUserSubscription } = useSubscriptionCancellation({
+    onModalClose: () => setIsSubscriptionEndingModalClosed(true),
+  });
+
+  const cancellationDate = individualPlan?.commitment?.cancellationDate;
+  const daysUntilCancellation = cancellationDate ? dateService.getDaysUntilExpiration(cancellationDate) : null;
+  const isSubscriptionEndingModalOpen =
+    !isSubscriptionEndingModalClosed &&
+    daysUntilCancellation !== null &&
+    SUBSCRIPTION_ENDING_WARNING_DAYS.includes(daysUntilCancellation);
```
CandelR: Maybe will be better to encapsulate the logic within a wrapper for the modal, so that this component remains unaware of anything, that way, everything that manages the modal’s state remains within its wrapper

xabg2: By creating a wrapper, we would move the logic from the parent to a child, which would have a dumb component. Instead, I can extract that logic to a custom hook to keep all we need regarding the Cancellation Modal in one place and the Cancellation Modal component still be dumb.

CandelR (approved):

PR #2048, opened 2026-07-15
[PB-6619]: refactor/refresh user token correctly

xabg2:
## Description

Refresh the user's access token when the app is opened if the refresh token is available and the access token has expired.
If the refresh token is expired or the token refresh fails, the existing logout flow will be triggered when the app attempts to refresh the user's data, redirecting the user to the login screen, as it did previously.

## Checklist

- [ ] QA Passed


yarn.lock
```diff
@@ -1929,10 +1929,10 @@
     "@typescript-eslint/parser" "^5.4.0"
     eslint-config-prettier "^8.3.0"
 
-"@internxt/lib@1.4.2":
-  version "1.4.2"
-  resolved "https://registry.yarnpkg.com/@internxt/lib/-/lib-1.4.2.tgz#fd178ff4e71381509648e469ec8b574082cd6eb2"
-  integrity sha512-3eYwSSl3NMwCyq+W8QkOrV0V0tCAcDfzmo8JH6BwU+rd4fMmJvriOnbFu9cKLJmXLk4GhCDEL5cr8rpFt4cAZQ==
+"@internxt/lib@=1.5.0":
```
larryrider: use the latest 1.5.1

larryrider (changes requested):

src/app/store/slices/user/index.ts
```diff
@@ -65,13 +65,16 @@ export const refreshUserThunk = createAsyncThunk<void, { forceRefresh?: boolean
...
-    const isExpired = isTokenExpired(userToken);
-
     const currentUser = getState().user.user;
-    if (!currentUser) throw new Error('Current user is not defined');
+
+    if (!currentUser || !userToken) throw new Error('Current user is not defined');
+
+    const tokenStatus = auth.validateTokenAndCheckExpiration(userToken);
+
+    const isRefreshRequired = tokenStatus === TokenStatus.REFRESH_REQUIRED || tokenStatus === TokenStatus.EXPIRED;
```
larryrider: if the TokenStatus === EXPIRED then we cannot use that token to get a new one, so in that case we need to logout the current user (or throw an ExpiredCredentials error)

larryrider: you could even check the invalid status too:
if TokenStatus === INVALID || TokenStatus === EXPIRED -> logout
else if TokenStatus === REFRESH_REQUIRED -> refreshToken....

larryrider (changes requested):

src/app/store/slices/user/index.ts
xabg2: But the user will get 401 for expired tokens, so the log out will be executed directly. 

But yeah, we can execute the logout directly instead. Let me improve it.

larryrider: yeah for me its better that way, if we know the token is expired then its better to do it right and do not rely on calling the backend again 👍

larryrider (dismissed):

CandelR (dismissed):

larryrider (approved):

CandelR (approved):

PR #2049, opened 2026-07-15
[MKT_928]:fix/ga services

jaaaaavier:
## Description
Purchases were not being tracked in Google Ads. This PR fixes the three issues that caused it:

- **`index.html`**: the Google Ads tag (`AW-XXXX`) was never configured with `gtag('config')`, so gtag silently dropped every event sent to the Ads destination, purchase conversions never left the browser. The account id is derived from`REACT_APP_GA_CONVERSION_TAG`.
- **`ga.service.ts`**: `begin_checkout` was sent with the Ads conversion label, so every checkout page view would be counted by Google Ads as a full-price purchase. It now goes to GA4 only.
- **`ga.service.ts`**: the `transaction_id` fallback used the user uuid, which repeats across purchases. Google Ads discards conversions with duplicated transaction ids, silently dropping repeat purchases. The fallback is now unique per purchase.

## Checklist

- [ ] QA Passed

## Testing Process
After carrying out several local tests, I was able to confirm that the data is now being sent correctly and that Google is processing it correctly, as verified by the marketing team.

## Additional Notes
I need to update the value of a production env.variable and create it in the preview and i dont have permission to do that, so that I can simulate a final purchase.

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,6 +77,10 @@
       gtag('js', new Date());
       gtag('config', '%REACT_APP_GA_ID%', { debug_mode: '%NODE_ENV%' === 'development' });
       gtag('config', '%REACT_APP_GA_BLOG_ID%', { debug_mode: '%NODE_ENV%' === 'development', groups: 'Blog' });
+      var adsAccountId = '%REACT_APP_GA_CONVERSION_TAG%'.split('/')[0];
+      if (adsAccountId.indexOf('AW-') === 0) {
+        gtag('config', adsAccountId, { allow_enhanced_conversions: true });
+      }
     </script>
 
     <!-- MailerLite Universal -->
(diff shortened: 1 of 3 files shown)


CandelR (approved):

index.html
```diff
@@ -77,6 +77,10 @@
       gtag('js', new Date());
       gtag('config', '%REACT_APP_GA_ID%', { debug_mode: '%NODE_ENV%' === 'development' });
       gtag('config', '%REACT_APP_GA_BLOG_ID%', { debug_mode: '%NODE_ENV%' === 'development', groups: 'Blog' });
+      var adsAccountId = '%REACT_APP_GA_CONVERSION_TAG%'.split('/')[0];
```
larryrider, line 80: use `const` instead of `var`

PR #2050, opened 2026-07-16
[MKT_923]:feat/new checkout design

jaaaaavier:
## Description
Redesign of the checkout page to reduce friction, the billing address form is gone for standard purchases.

- **Removed the global `AddressElement`** (billing name + full address) and the section titles/plan-features list from the checkout layout. The Pay button moved into the order summary card on desktop.
- **Billing country is now derived from the user's IP** (`useUserLocation`). The payments backend only requires `country` + `captchaToken` on `POST /customer`; all address fields are optional there.
- **Conditional postal code input**: shown only when the detected country is US, CA or IN, where Stripe Tax requires country + postal code to calculate tax (everywhere else the country alone is enough). Validated before payment.
- **Conditional address form for crypto**: the `AddressElement` (name + full billing address) now renders only when the crypto payment dropdown is opened. Bit2Me requires full name and complete postal address for KYC (see internxt/payments#322, #327, #328 — PB-5312), so the strict address validation now applies only to crypto purchases.
- **Customer name fallback chain**: `companyName` → typed name (crypto form) → authenticated user's name → email.
- **Tax preview**: initial price+tax preview keeps working via IP (already supported by `GET /price-by-id`); the recalculation effect now reacts to both the conditional postal code input and the crypto address form, debounced (500 ms).
- **Local type relaxation** in `checkout.service.ts` (`CreateCheckoutCustomerPayload`): the SDK's `CreateCustomerPayload` still marks the address fields as required even though the backend does not. To be removed once the SDK type is updated.
- New translation keys (`checkout.postalCode.*`, `checkout.error.postalCodeRequired`, `checkout.error.countryRequired`) in all 8 locales.

## Checklist

- [ ] QA Passed

## Testing Process
For manual verification in Stripe test mode, the flows to cover are the following. For an EU card purchase (subscription and lifetime), no address fields are shown, the customer is created with the IP-derived country, tax is calculated and the payment confirms. With a US IP, the postal code input appears under the payment method, paying without it shows the `postalCodeRequired` error, and once a ZIP is entered the tax preview recalculates. For a crypto lifetime purchase, opening the crypto dropdown shows the billing address form, submitting an incomplete address shows the `addressRequired` error, and a complete address reaches Bit2Me invoice creation. Finally, for logged-in vs. new users, the customer name falls back correctly (account name / email).


src/views/Checkout/services/checkout.service.ts
```diff
@@ -24,6 +24,12 @@ const fetchPromotionCodeByName = async (priceId: string, promotionCodeName: stri
   return paymentClient.fetchPromotionCodeByName(priceId, promotionCodeName);
 };
 
+export type CreateCheckoutCustomerPayload = Pick<
```
xabg2, line 28: It's better to clean up the types from the SDK/payments API if they're no longer needed. That way, we can avoid maintaining dead code.

src/views/Checkout/components/CheckoutProductCard.tsx
```diff
@@ -31,8 +32,8 @@ export const CheckoutProductCard = ({
   selectedPlan,
   couponCodeData,
   showCouponCode,
-  showHardcodedRenewal,
   couponError,
+  isButtonDisabled,
```
xabg2: Update the name for something more descriptive, like `isPaymentProcessing`

xabg2 (commented):
ℹ️  Also, remove the useless translations for each JSON if needed

src/views/Checkout/services/checkout.service.ts
jaaaaavier: Yeah better

src/views/Checkout/components/CheckoutProductCard.tsx
jaaaaavier: Oki

jaaaaavier:
I updated the SDK version and checked everything works as expected with the last deploy on payments @xabg2

src/views/Checkout/views/CheckoutViewWrapper.tsx
```diff
@@ -306,10 +318,25 @@ const CheckoutViewWrapper = () => {
...
+      const billingPostalCode = address?.postal_code ?? (postalCode.trim() || undefined);
+
+      if (!billingCountry) {
+        throw new Error(translate('checkout.error.countryRequired'));
+      }
+
+      if (isPostalCodeRequired && !billingPostalCode) {
+        throw new Error(translate('checkout.error.postalCodeRequired'));
+      }
+
```
CandelR, line 339: CheckoutViewWrapper is taking on too many unrelated responsibilities, a lot of useEffect plus the crypto, postcode, auth and submit state, all within the same component. The file itself already uses the correct pattern for others (useCheckout, useAuthCheckout, useProducts), I would follow that same approach

jaaaaavier: Okay, I'll go over it

jaaaaavier:
Extracted only the address/zip code portion of the task to a hook (useBillingDetails), following the pattern of the existing hooks.

jaaaaavier:
Added a new order for the payment methods depending on the country of the customer

xabg2 (approved):

PR #2052, opened 2026-07-17
feat:(PB-6297) preview public shared folder contents

victor-ferro:
## Description

First part of PB-6297 (public shared folder preview). Until now, a public folder link only showed a card with the folder name and a Download button — the recipient had to download the whole folder as a zip without knowing its contents.

This PR adds a "View" button next to Download that opens a read-only listing of the shared folder contents, fetched without authentication:

- `usePublicSharedFolderContent` (new hook): fetches folders first and then files per level via `getPublicSharedFolderContent`, handling pagination (30 items/page), per-level resource tokens (each response provides the token that authorizes listing the next level) and network credentials (returned by the files request, which carries the link `code`).
- `PublicSharedFolderContent` (new): breadcrumbs, item selection, sorting, and navigation into subfolders. Exiting the preview returns to the classic card.
- `PublicSharedItemList` (new): list UI for the items.
- `ShareFolderView`: renders the new preview when the eye button is clicked.
- `ShareFileView`: Phosphor icon imports renamed to their non-deprecated `*Icon` variants (no behavior change).

Scope note: this PR only lists contents. Image thumbnails and per-item download/preview come in two follow-up stacked PRs.

## Related Issues

Relates to PB-6297.

## Related Pull Requests

Stacked on top of this PR (will be opened next):
- `PB-6297-thumbnails` — image thumbnails in the listing.
- `PB-6297-preview-download` — per-item download and file preview.

## Checklist

- [x] Changes have been tested locally.
- [x] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [ ] SonarCloud issues have been reviewed and addressed.
- [ ] QA Passed

## Testing Process

- Unit tests: `yarn vitest run src/views/PublicShared/hooks/usePublicSharedFolderContent.test.ts` (4 tests: initial folders→files fetch sequence, subfolder navigation with the parent level token + breadcrumb back-navigation, no-op navigation guards, and pagination).
- `yarn typecheck` and `yarn lint` clean; full unit suite run by the pre-commit hook.
- Manual: opened a public shared folder link (`/sh/folder/:token/:code`), clicked View, browsed root and subfolders, checked pagination on folders with 30+ items, breadcrumb navigation back to parent levels, and exit back to the download card. Verified with and without link password.

PR #2053, opened 2026-07-17
feat:(PB-6297) show image thumbnails in public shared folder preview

victor-ferro:
## Description

Second part of PB-6297, stacked on #2052. The public shared folder listing now shows real image thumbnails instead of the generic file-type icon.

Thumbnails are encrypted files in the network like any other, so downloading one needs the same three inputs as a regular file: its network coordinates, the share network credentials, and the decryption key. This PR wires those for the public (unauthenticated) context:

- `derivePublicSharingKey` (new, extracted in `share.service.ts`): the `aes.decrypt(encryptionKey, code)` + sharing-version handling that lived inline in `downloadPublicSharedFolder` is now a reusable function returning a `FileKey` (`{ bucketKey }` for `inxt-v3` sharings, `{ mnemonic }` for older ones). `downloadPublicSharedFolder` now consumes it — pure refactor, no behavior change.
- `downloadPublicThumbnail` (new in `thumbnail.service.ts`): downloads and decrypts a thumbnail (`bucket_id`/`bucket_file`) with explicit credentials and key, returning a Blob — the existing `downloadThumbnail` relies on the authenticated user context, which does not exist on a public link.
- `PublicSharedItemList`: for image items with a thumbnail, downloads it and renders it via `URL.createObjectURL` (revoked on unmount); falls back to the type icon otherwise. Credentials travel per-item (set by the hook from the files response), and the share key is derived once in `PublicSharedFolderContent` (`useMemo`) and passed down.
- `ShareFolderView`: passes `encryptionKey`/`sharingVersion` from the sharing meta down to the preview.

## Related Issues

Relates to PB-6297.

## Related Pull Requests

- Stacked on #2052 (`PB-6297-folder-link-preview`) — merge that one first.
- Followed by `PB-6297-preview-download` (per-item download and file preview), to be opened next.

## Checklist

- [ ] QA Passed

## Testing Process

- Unit tests: `yarn vitest run src/app/drive/services/thumbnail.service.test.ts` — new case: "When downloading a public thumbnail, then it downloads the thumbnail file with the share credentials and key".
- `yarn typecheck` and `yarn lint` clean; full unit suite run by the pre-commit hook.
- Manual: opened a public shared folder link containing images, verified thumbnails render in the listing after the files page loads, non-image items keep their type icon, and navigation between folders loads each level's thumbnails. Checked both sharing versions derive a working key (v2 mnemonic / v3 bucketKey).


src/views/PublicShared/components/PublicSharedItemList.tsx
```diff
@@ -20,15 +26,46 @@ const skinSkeleton = [
...
+const PublicSharedListItem = ({ item, publicShareKey, onNameClicked }: PublicSharedListItemProps) => {
   const ItemIconComponent = iconService.getItemIcon(item.isFolder, item.type);
+  const [thumbnailUrl, setThumbnailUrl] = useState<string>();
   const displayName =
     transformItemService.getItemPlainNameWithExtension(item as unknown as DriveItemData) ??
     items.getItemDisplayName(item);
 
+  useEffect(() => {
+    const thumbnail = (item.thumbnails as Thumbnail[] | undefined)?.[0];
+    const isImage = !item.isFolder && !!item.type && thumbnailableImageExtension.includes(item.type.toLowerCase());
```
CandelR: not only the images has thumbnails, also the videos and some docs as pdfs

src/views/PublicShared/components/PublicSharedItemList.tsx
```diff
@@ -20,15 +26,46 @@ const skinSkeleton = [
...
+    if (!thumbnail || !isImage || !item.credentials) return;
+
+    let objectUrl: string | undefined;
+    let isUnmounted = false;
+
+    downloadPublicThumbnail(
+      thumbnail,
+      { user: item.credentials.networkUser, pass: item.credentials.networkPass },
+      publicShareKey,
```
CandelR, line 53: nit: For quick navigation and folders containing lots of images, an abort controller might come in handy (this would also apply to Drive, which doesn’t have one, and it would be good to be able to cancel requests. I’m not saying we should do the Drive part here! haha)

src/views/PublicShared/components/PublicSharedItemList.tsx
victor-ferro: Totally!!! Done!

src/views/PublicShared/components/PublicSharedItemList.tsx
victor-ferro: Done!

CandelR (approved):

xabg2 (approved):
LGTM. Only checked the payments-related changes.

PR #2054, opened 2026-07-17
feat:(PB-6297) download and preview individual items in public shared folder

victor-ferro:
## Description

Third and final part of PB-6297, stacked on #2053. The public shared folder preview becomes actionable: items can now be downloaded individually (or in multi-selection) and files can be previewed in the viewer, all without authentication.

**Per-item download**
- `downloadPublicSharedItems` (new in `share.service.ts`): generalizes the existing root-folder download to any selection — a single file downloads directly (stream → blob → save), a single folder zips through the public iterators, and a mixed selection packs everything into one zip. `downloadPublicSharedFolder` is now a thin wrapper over it (same behavior as before for the main Download button).
- Context menu (right-click / three-dots) with Download on any selection, and a `beforeunload` warning while a download is in flight.

**File preview**
- Double-click (or Preview in the menu) opens the plain `FileViewer` with the decrypted blob, download progress, and the same gates as `ShareFileView`: previewable type and size limit — otherwise the viewer opens with its fallback without fetching the blob.
- `TopBarActions` gets a dedicated share-view variant: Download + Report (reusing `handleReportShare`, now exported from `ReportButton`). This also removes a dead branch: the old copy-link button was unreachable in share views. Phosphor deprecated icon names updated (`DotsThreeIcon`, `WarningCircleIcon`).

**Cleanups (code created earlier in this stack)**
- `usePublicSharedFolderContent`: `fetchFolders`/`fetchFiles` merged into a single `fetchLevelItems(type)` — same phase machine (folders per level, then files), ~20 duplicated lines removed. Test assertions updated accordingly.
- `useWarnBeforeUnload` (new hook): replaces the three identical `handleLeavePage` + `beforeunload` effect copies in `ShareFolderView`, `ShareFileView` and the new preview; also fixes the effect depending only on `progress` instead of the full active condition.
- `isTypeAllowed` now reuses `getIsTypeAllowedAndFileExtensionGroupValues` from `fileViewerUtils` (also fixes uppercase extensions being rejected).
- `ShareLayout`: user menu dropdown gets `z-50` so it stays above the new preview content.

## Related Issues

Relates to PB-6297.

## Related Pull Requests

- Stacked on #2053 (`PB-6297-thumbnails`), which is stacked on #2052 — merge in order: #2052 → #2053 → this one.

## Checklist

- [ ] QA Passed

## Testing Process

- Unit tests: `yarn vitest run src/app/share/services/share.service.test.ts` — three new cases for `downloadPublicSharedItems` (single file with share credentials/key and plain name, single folder zipped through the public iterators, mixed selection packed into one zip closed at the end). Hook tests updated for the unified fetch.
- `yarn typecheck` and `yarn lint` clean; full unit suite run by the pre-commit hook.
- Manual: on a public shared folder link — downloaded a single file, a folder, and a mixed multi-selection (zip); previewed images/PDFs via double-click and the context menu; verified non-previewable and oversized files open the viewer fallback without downloading; downloaded from inside the viewer; checked the leave-page warning appears only while downloading; Report button in the viewer opens the report email.


src/views/PublicShared/components/PublicSharedFolderContent.tsx
```diff
@@ -80,13 +99,105 @@ const PublicSharedFolderContent = ({
...
+      options: {
+        notifyProgress: (totalBytes, downloadedBytes) => {
+          setPreviewProgress(downloadedBytes / totalBytes);
+        },
+      },
+    })
+      .then((fileStream) => binaryStreamToBlob(fileStream, shareItem.type || ''))
+      .then((fileBlob) => {
+        setPreviewBlob(fileBlob);
+      })
```
CandelR: need to verify that the downloaded item is the same as actually opened, this can cause that user press one item, close instantalnealy, open another, and flash the preview of first item until the second is downloaded

src/views/PublicShared/components/PublicSharedFolderContent.tsx
```diff
@@ -80,13 +99,105 @@ const PublicSharedFolderContent = ({
...
+
+  const openPreview = (shareItem: AdvancedSharedItem) => {
+    if (shareItem.isFolder || !shareCredentials) return;
+
+    setPreviewItem(shareItem);
+
+    if (!isTypeAllowed(shareItem) || !isFileSizePreviewable(Number(shareItem.size))) return;
+
+    downloadFile({
+      bucketId: (shareItem as unknown as DriveFileData).bucket,
```
CandelR: avoid as unknown cast outside of tests

src/views/PublicShared/components/PublicSharedFolderContent.tsx
```diff
@@ -80,13 +99,105 @@ const PublicSharedFolderContent = ({
...
+      resourcesToken: nextLevelToken,
+    })
+      .catch((err) => {
+        errorService.reportError(err);
+        notificationsService.show({ text: errorService.castError(err).message, type: ToastType.Error });
+      })
+      .finally(() => {
+        setIsDownloading(false);
+      });
+  };
```
CandelR: I would extract all the download logic added for previewing into a hook

src/views/PublicShared/hooks/useWarnBeforeUnload.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react';
+
+const handleLeavePage = (e: BeforeUnloadEvent) => {
+  const confirmationMessage = '';
+
+  e.returnValue = confirmationMessage; //Trident, Chrome 34+
+  return confirmationMessage; // WebKit, Chrome <34
+};
+
+const useWarnBeforeUnload = (isActive: boolean): void => {
```
CandelR: this hook already exists: useBeforeUnload.ts

src/app/share/services/share.service.ts
```diff
@@ -706,6 +710,94 @@ export function derivePublicSharingKey({
   return { mnemonic: decrypted };
 }
 
+const getPublicItemDownloadName = (item: AdvancedSharedItem): string => {
+  const extension = item.type ? `.${item.type}` : '';
+  return `${item.plainName ?? item.name}${extension}`;
+};
+
+export async function downloadPublicSharedItems({
```
CandelR, line 717: Doesn't the DownloadManager cover this?

src/views/PublicShared/components/PublicSharedFolderContent.tsx
victor-ferro: Done!

src/views/PublicShared/components/PublicSharedFolderContent.tsx
victor-ferro: Done! It's been detected that the AdvancedSharedItem type doesn't have fileId because the SharedFiles imported from the SDK has a field called 'fileiId' but doesn't have 'fileId'

src/views/PublicShared/components/PublicSharedFolderContent.tsx
victor-ferro: Done!

src/views/PublicShared/hooks/useWarnBeforeUnload.ts
victor-ferro: I hadn't seen it. The functionality I put into the hook was also being used in other places, and I thought there wasn't already an existing hook for this. My bad!!
Done!

src/app/share/services/share.service.ts
victor-ferro: True, it can be done, but it touches several things in the main download flow. Because for public downloads we'd run into some blockers:
1. generateTasksForItem calls localStorageService.getUser() and throws 'User not found' before even looking at downloadCredentials, so anonymous visitors crash right away.
2. Progress is reported through tasksService, and TaskLogger is only mounted in HeaderAndSidenavLayout — an anonymous user would get a download with zero feedback, and the preview UI here doesn't read from tasks.
3. Single files go through the download worker, and postMessage turns the { bucketKey: Buffer } key (new sharing version) into a plain Uint8Array, which the network crypto won't accept without re-hydrating it in the worker.

If you want me to get started on it, I will. But maybe it might be work for a separate ticket

CandelR: Okay, so we’ll have to create a task to decouple DownloadManager from everything that isn’t handle ‘download', because, as we have seen, it does not allow us to reuse it 😞

src/hooks/useBeforeUnload.test.ts
```diff
@@ -0,0 +1,102 @@
+/**
+ * @jest-environment jsdom
+ */
```
CandelR: we don't use jest, this tag is not necessary

src/views/PublicShared/hooks/usePublicSharedDownload.test.ts
```diff
@@ -0,0 +1,267 @@
+/**
+ * @jest-environment jsdom
+ */
```
CandelR: same here, we don't use jest

CandelR (approved):

CandelR (approved):

PR #2074, opened 2026-07-27
[_] bugfix/url validation redirect

CandelR:
## Description

Fixes a redirect vulnerability in the login/signup flow.

## Related Pull Requests

- https://github.com/internxt/drive-web/pull/2072

## Checklist

- [ ] QA Passed

## Testing Process

- Do login from the Desktop app, should complete and hand off credentials back to the app normally.
- Do login from Mobile, should complete and open the app normally.


src/views/Login/UniversalLinkSuccessView.tsx
```diff
@@ -35,7 +36,16 @@ export default function UniversalLinkView(): JSX.Element {
...
+      const decoded = Buffer.from(redirectUri, 'base64').toString();
+      if (
+        validateUrl({
+          urlString: decoded,
+          allowedProtocols: TRUSTED_LOCALHOST_PROTOCOLS,
+          allowedHostnames: TRUSTED_LOCALHOST_HOSTNAMES,
+        })
+      ) {
+        baseURL = decoded;
+      }
```
larryrider, line 48: if there is a redirectUri but it is not valid (malicius url not pointing to internxt or localhost) then we redirect to the baseURL. Maybe it would be better to just dont redirect at all

larryrider (changes requested):

src/views/Login/UniversalLinkSuccessView.tsx
CandelR: That’s a good point, perhaps it would be better to display a ‘invalid URL’ notification and don't redirect. What do you think?

larryrider: lgtm!

pascscha:
@CandelR I wouldn't wait too long to handle this, this PR fixes a pretty nasty issue that allows to steal user session and keys. Also, even after this hotfix the desktop SSO flow is still flawed, because currently the token and encryption key land in the browser history. Please refer to my E-Mails for more info.

larryrider (dismissed):

larryrider (approved):

PR #2075, opened 2026-07-28
[PB-6707]: keep folder uploads running when individual files fail

terrerox:
## Context

Users were seeing a folder upload reported as **successful when the folder had not actually been uploaded**: a single file failing mid-upload could kill the queue while the task logger still announced success, leaving the user with no signal that content was missing. This PR implements the solution proposed by the product team: keep uploading everything that can succeed, be honest about what failed, and give the user a way to retry exactly those files.

## What changed

Folder uploads no longer abort and roll back when a single file fails. The manager marks the task with hasFailedFiles, lets the queue finish the remaining items, and announces the failure once the whole folder is done, keeping already-uploaded content. Referral tracking and network logging are skipped for partially failed folders.

Failed files land in the retry list grouped by their parent task, with duplicates replaced by taskId on re-add. Errors that can never succeed (402 payment required, file size exceeded) are marked non-retryable and shown as "Not allowed" in the retry modal instead of a retry button.

Also renames boolean flags to the is/has/should convention, fixes DEFAULT_OPTIONS being mutated via Object.assign in uploadItemsThunk, normalizes the two RetryableTask param shapes in TaskToRetryItem, and adds size props to TaskLoggerButton for the retry modal polish.
## Checklist

- [ ] QA Passed


src/app/network/RetryManager.ts
```diff
@@ -15,7 +16,11 @@ class RetryManager {
   }
 
   addTasks(tasks: RetryableTask[]) {
+    // A task can only appear once per taskId: re-adding a failed task replaces its entry
```
sg-gs: I would add JSDoc to this function rather than this comment

sg-gs (dismissed):

sg-gs (dismissed):

sg-gs (dismissed):
ℹ️ Reviewed only the network part

larryrider (approved):

PR #2076, opened 2026-07-28
[PB-6707]: surface failed uploads in the retry modal and task logger

terrerox:
## Description

Retry entries are grouped under their parent folder task via relatedTaskId, and non-retryable files show a "Not allowed" label (added to all locales) instead of a retry button.

Also normalizes the two RetryableTask param shapes in TaskToRetryItem behind a getDisplayData helper (file names now display with their extension), tightens the retry modal row padding, and adds sizeClassName/iconSize props to TaskLoggerButton.

## Checklist

- [ ] QA Passed

### DIFF
--- a/src/app/tasks/components/TaskLogger/TaskLogger.tsx
+++ b/src/app/tasks/components/TaskLogger/TaskLogger.tsx
@@ -27,7 +27,7 @@ const TaskLogger = (): JSX.Element => {
   const filesToRetryGroupedByTask = useMemo(
     () =>
       filesToRetry.reduce<Record<string, RetryableTask[]>>((acc, file) => {
-        const relatedTaskId = file.taskId;
+        const relatedTaskId = file.params?.relatedTaskId ?? file.taskId;
         if (relatedTaskId) {
           if (!acc[relatedTaskId]) {
             acc[relatedTaskId] = [];
--- a/src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
+++ b/src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
@@ -84,14 +84,13 @@ const PendingBlock = (): JSX.Element => {
 
 const SuccessBlock = ({ isHovered, magnifyingAction, infoAction, taskType, haveWarnings }): JSX.Element => {
   if (isHovered && taskType.includes('upload')) {
-    return (
-      <>
-        {haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}
-        <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
-      </>
+    return haveWarnings ? (
+      <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} iconSize={20} />
+    ) : (
+      <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
     );
   } else if (isHovered && taskType.includes('download')) {
-    return <>{haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}</>;
+    return <>{haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} iconSize={20} />}</>;
   } else {
     return (
       <div className="flex h-8 w-8 items-center justify-center">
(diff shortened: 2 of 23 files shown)


src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
```diff
@@ -84,14 +84,13 @@ const PendingBlock = (): JSX.Element => {
...
   if (isHovered && taskType.includes('upload')) {
-    return (
-      <>
-        {haveWarnings && <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} />}
-        <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
-      </>
+    return haveWarnings ? (
+      <TaskLoggerButton onClick={infoAction} Icon={InfoIcon} iconSize={20} />
+    ) : (
+      <TaskLoggerButton onClick={magnifyingAction} Icon={MagnifyingGlass} />
```
CandelR, line 90: If there’s any file with error, do we no longer display the magnifying glass icon to navigate to the folder?

src/app/tasks/components/TaskToRetryItem/TaskToRetryItem.tsx
```diff
@@ -8,33 +8,64 @@ import { TaskLoggerButton } from '../TaskLoggerButton/TaskLoggerButton';
...
+  isFolder: boolean;
+}
+
+const getDisplayData = (params: RetryableTask['params']): DisplayData => ({
+  name: params?.filecontent?.name ?? params.plainName ?? params.name,
+  type: params?.filecontent?.type ?? params.type,
+  size: params?.filecontent?.size ?? params.size,
+  modifiedAt: params?.filecontent?.content.lastModified ?? params.updatedAt,
+  isFolder: Boolean(params?.isFolder),
+});
```
CandelR: I understand that this depends on whether it’s an upload or a download, as the data arrives with the fields structured differently, but this needs to be reorganised more effectively, as looking at it like this, we can’t tell why there are so many fallback options for the same field.
Looking at this code alone, it’s impossible to tell why, for ‘name’, it checks ‘filecontent.name’, then uses ‘params.plainname’ as a fallback, and then ‘params.name’.

src/app/tasks/components/TaskLoggerActions/TaskButtonActionBlocks.tsx
terrerox: Exactly. Now we’ll enable an info icon that, when clicked, will open this modal.
Screenshot 2026-08-03 at 11 07 03 PM

src/app/tasks/components/TaskLogger/TaskLogger.tsx
```diff
@@ -27,7 +27,7 @@ const TaskLogger = (): JSX.Element => {
   const filesToRetryGroupedByTask = useMemo(
     () =>
       filesToRetry.reduce<Record<string, RetryableTask[]>>((acc, file) => {
-        const relatedTaskId = file.taskId;
+        const relatedTaskId = file.params?.relatedTaskId ?? file.taskId;
```
TamaraFinogina, line 30: Why is` file.params?.relatedTaskId` the primary option now?

src/app/tasks/components/TaskToRetryItem/TaskToRetryItem.tsx
```diff
@@ -6,35 +6,79 @@ import { formatDefaultDate } from 'services/date.service';
...
         </div>
       </div>
-      {status === 'failed' && <TaskLoggerButton onClick={() => downloadItem(file)} Icon={RestartIcon} />}
-      {status === 'retrying' && <CircleNotch size={24} className="mr-2 animate-spin text-gray-60" weight="bold" />}
+      {isNotAllowed && <span className="mr-2 text-sm font-medium text-gray-50">{t('tasks.messages.notAllowed')}</span>}
+      {!isNotAllowed && status === RetryableTaskStatus.Failed && (
+        <TaskLoggerButton onClick={() => downloadItem(file)} Icon={RestartIcon} sizeClassName="h-8 w-8" iconSize={16} />
+      )}
+      {!isNotAllowed && status === RetryableTaskStatus.Retrying && (
+        <CircleNotch size={16} className="mr-2 animate-spin text-gray-60" weight="bold" />
```
TamaraFinogina: Change it to Spinner; CircleNotch is deprecated

src/app/tasks/components/TaskLogger/TaskLogger.tsx
terrerox: Because we are moving from "all-or-nothing folder uploads" to "keep what worked, show what didn't, let me retry it", and that line is what makes the failures show up in the place the user thinks of as the upload: the folder

TamaraFinogina (approved):

CandelR (approved):

PR #2077, opened 2026-07-28
[MKT_947]:feat/New checkout improves

jaaaaavier:
## Description
In this PR, we’ve improved the checkout process. Changes:

- Until now, the postalCode was only required for a few countries, as it wasn’t always necessary, we used to dynamically display an input field that looked awful. We now retrieve this information from the widget used by Stripe.

- The AvailableCryptoCurrenciesDropdown.tsx has been improved to make it clearer which cryptocurrency is intended for use. Furthermore, as payments with cryptocurrencies require the user’s first name, surname and address, we now dynamically display these fields, reusing the AddressElement previously used in the checkout process but integrated into this component.

- Previously, when a user made a mistake whilst logging in at checkout (incorrect password, email already registered, etc.), we displayed an error showing the code, e.g. 409 now we display a short message that tells the user exactly what mistake they have made.

## Checklist

- [ ] QA Passed

## Testing Process
For testing purposes, I tried the following scenarios:
- With a VPN to Canada: I verified that the card payment uses the ZIP code entered in the Stripe widget
- Using a VPN to Canada, I verified that crypto payments request all the necessary information and proceed past the widget to complete the payment
- Without a VPN, I verified that credit card payments work
- Without a VPN, I verified that crypto payments request all the necessary information and proceed past the widget to complete the payment
- I forced login errors and verified that text appears instead of an error code


src/views/Checkout/components/AvailableCryptoCurrenciesDropdown.tsx
```diff
@@ -78,6 +95,19 @@ export const AvailableCryptoCurrenciesDropdown = ({
           </MenuItems>
         </Transition>
       </Menu>
+      {cryptoSelected && !isDropdownOpen && (
```
xabg2: Can you extract that to a constant? Something like `const shouldDisplayAddressElement = cryptoSelected && !isDropdownOpen;`

src/views/Checkout/components/AvailableCryptoCurrenciesDropdown.tsx
```diff
@@ -78,6 +95,19 @@ export const AvailableCryptoCurrenciesDropdown = ({
...
+        <div className="flex w-full flex-col gap-3">
+          <div className="flex w-full border-t border-gray-10" />
+          <AddressElement
+            options={CRYPTO_ADDRESS_ELEMENT_OPTIONS}
+            onChange={(event) => {
+              const { firstName, lastName, name, address } = event.value;
+              onUserNameChanges([firstName, lastName].filter(Boolean).join(' ').trim() || name);
```
xabg2: Extract the filter/join/trim to a constant please.

src/views/Checkout/hooks/useAuthCheckout.test.ts
xabg2, line 1: As you added an object for the translation keys for the errors, why you did not use these keys to match it against the error? I mean, instead of doing:
`expect(hookState.current.authError).toBe('checkout.authComponent.authError.invalidCredentials');`

you can do:
`expect(hookState.current.authError).toBe(AUTH_ERROR_TRANSLATION_KEY['invalidCredentials']);`

src/views/Checkout/hooks/useAuthCheckout.ts
```diff
@@ -15,8 +17,68 @@ export interface AuthCheckoutProps {
...
+  authMethod: AuthMethodTypes,
+  status?: number,
+): Promise<boolean> => {
+  const isFailedSignIn = authMethod === 'signIn' && status === STATUS_CODE_ERROR.UNAUTHORIZED;
+
+  if (!isFailedSignIn) return false;
+
+  try {
+    return await is2FANeeded(email);
+  } catch {
```
xabg2: Log the error, just in case

src/views/Checkout/hooks/useAuthCheckout.ts
```diff
@@ -37,10 +99,14 @@ export const useAuthCheckout = ({ changeAuthMethod }: Pick<AuthCheckoutProps, 'c
         token: authCaptcha,
         doSignUp: doRegister,
       });
+      setAuthError(null);
       return profileInfo.user;
     } catch (err) {
-      const error = err as Error;
-      setAuthError(error.message);
+      errorService.reportError(err);
```
xabg2: No need to report the error. There is no instrument. it just logs the error.

src/views/Checkout/hooks/useAuthCheckout.ts
```diff
@@ -37,10 +99,14 @@ export const useAuthCheckout = ({ changeAuthMethod }: Pick<AuthCheckoutProps, 'c
...
+      const { status } = errorService.castError(err);
+      const isTwoFactorEnabled = await isTwoFactorBlockingTheSignIn(email, authMethod, status);
+
+      setAuthError(translate(getAuthErrorTranslationKey({ status, authMethod, isTwoFactorEnabled })));
```
xabg2: Try to not anidate a lot of things. Extract `getAuthErrorTranslationKey` to a constant, like: `const errorKey = getAuthErrorTranslationKey` and then:
```
const errorMessage = translate(errorKey);
setAuthError(errorMessage);
```

src/views/Checkout/components/AvailableCryptoCurrenciesDropdown.tsx
```diff
@@ -78,6 +95,19 @@ export const AvailableCryptoCurrenciesDropdown = ({
...
+              onUserAddressChanges(address);
+            }}
+          />
+        </div>
+      )}
```
xabg2: Why did you move the AddressElement inside this component instead of handling it from the parent? I'm missing something? 🤔 

Keep in mind that this component is the crypto currencies Dropdown. Not sure if this component should also handle the `AddressElement`.

src/views/Checkout/views/CheckoutViewWrapper.tsx
```diff
@@ -309,16 +301,26 @@ const CheckoutViewWrapper = () => {
...
-        throw new Error(translate('checkout.error.postalCodeRequired'));
+      let paymentPostalCode: string | undefined;
+
+      if (isCryptoPurchase) {
+        paymentPostalCode = address?.postal_code;
       }
 
       if (currencyType === PaymentType['FIAT']) {
         const { error: elementsError } = await elements.submit();
+        const { confirmationToken, error: confirmationTokenError } = await stripeSDK.createConfirmationToken({ elements });
```
xabg2, line 312: How does it works? 🤔 When we need to pass the elements when currency is FIAT. Afaik, we only display the Elements when the currency is crypto isn't it?

xabg2 (changes requested):

src/views/Checkout/views/CheckoutViewWrapper.tsx
jaaaaavier: I use this confirmationToken to retrieve the ZIP code from the card payment widget created by Stripe. When a user from, say, Canada or the United States tries to make a purchase, they are automatically asked for their ZIP code, and we need it to correctly create the customer.

In the production environment right now, you can see that if you're in an area that requires a ZIP code, a new field appears for entering the ZIP code this was incorrect.

Captura de pantalla 2026-07-28 a las 14 57 20

src/views/Checkout/components/AvailableCryptoCurrenciesDropdown.tsx
jaaaaavier: The idea from the beginning was to remove the billing address, but for cryptocurrency payments, it's necessary to know the buyer's address as well as their first and last name. I decided to include it within the component and have it displayed only after the cryptocurrency is selected, since that made more sense to me than putting it in the parent component. I've checked this with Fran, and everything is correct.

Captura de pantalla 2026-07-28 a las 15 01 53

xabg2: Rename the component then.

src/views/Checkout/hooks/useAuthCheckout.test.ts
jaaaaavier: Okay

src/views/Checkout/components/AvailableCryptoCurrenciesDropdown.tsx
xabg2: Remember to rename the component @jaaaaavier

jaaaaavier: Renamed

xabg2 (approved):

PR #2081, opened 2026-07-30
[MKT_949]:feat/cloudflare turnstile integration

jaaaaavier:
## Description
In this PR, we've set up everything needed to integrate Turnstile into the checkout process. To do this, we've done the following:

- We added the Turnstile script to the project's <head> tag
- We added the REACT_APP_TURNSTILE_SITE_KEY variable, which we'll use for authentication
- We created the widget
- We created the `generateTurnstileToken` file. In this file, we defined the `generateTurnstileToken` function, which sets up the widget
- We added the `turnstileToken` to the relevant locations in both `useUserPayment` and `checkout.service`, as specified in the payments documentation and the SDK

## Checklist

- [ ] QA Passed

## Testing Process
This is part of the [Turnstile Integration](https://github.com/internxt/payments-server/pull/394) so the test are the same

## Additional Notes
Need to add `REACT_APP_TURNSTILE_SITE_KEY` to env variables

The error is because i need to publish a new version of the SDK


src/utils/generateTurnstileToken.test.ts
xabg2, line 1: Use `test('When..., then..., () => {...})` instead of using `it('should...., () => {})`

src/views/Checkout/views/CheckoutViewWrapper.tsx
```diff
@@ -367,9 +369,12 @@ const CheckoutViewWrapper = () => {
         city: address?.city,
         companyVatId,
         captchaToken: customerToken,
+        turnstileToken: customerTurnstileToken,
         metadata,
       });
 
+      const paymentTurnstileToken = await generateTurnstileToken();
```
xabg2, line 376: Why are you generating 2 different tokens? We can't use just one token for both operations as we do with captcha?

PR #2082, opened 2026-07-30
[PB-6680] Enc workspace mnemonic and credentials

TamaraFinogina:
## Description

Encrypt workspace mnemonic and credentials

## Related Issues

Relates to [PB-6680](https://inxt.atlassian.net/browse/PB-6680)

## Checklist

- [x] QA Passed

## Testing Process

- Workspace mnemonic is encrypted (check local storage)
- Workspace credentials are encrypted (check local storage)
- Inviting a user to a workspace keeps working
- Switching between workspaces works
- Removing a user from a workspace works
- Downloading and uploading files to the workspace keeps working

## Additional Notes

Workspace credentials are used to set up sdkFactory, so they should be in bootstrap

[PB-6680]: https://inxt.atlassian.net/browse/PB-6680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/encrypted-storage.service.ts
```diff
@@ -41,13 +51,59 @@ const getToken = (): string | undefined => tokenCache ?? undefined;
...
+  localStorage.removeItem(LocalStorageItem.B2BworkspaceId);
+};
+
+const setB2BWorkspace = async (workspaceID: string, workspaceMnemonic: string): Promise<void> => {
+  workspaceMnemonicCache = workspaceMnemonic || null;
+  localStorage.setItem(LocalStorageItem.B2BworkspaceId, workspaceID);
+  await setAndEncrypt(LocalStorageProtectedItem.EncryptedB2BworkspaceMnemonic, workspaceMnemonic);
+};
+
+const getWorkspaceCredentials = (): WorkspaceCredentialsDetails | null => workspaceCredentialsCache ?? null;
```
sg-gs, line 126: Why `setWorkspaceCredentials` sets and encrypts but `getWorkspaceCredentials` does not get and decrypt if cache fails?

src/app/store/slices/workspaces/workspacesStore.ts
```diff
@@ -263,7 +267,6 @@ export const workspacesSlice = createSlice({
           item.workspace = Object.assign(workspace, patch);
           if (state.selectedWorkspace?.workspace.id === workspaceId) {
             state.selectedWorkspace = item;
-            localStorageService.setB2BWorkspace(item.workspace.id, item.workspaceUser.key);
```
sg-gs, line 270: Why this does not have any equivalent on the new diff ?

src/services/encrypted-storage.service.ts
TamaraFinogina: Because `getWorkspaceCredentials` is called in `sdkFactory`, same as `getToken`. So, changing it to `async` (and it's the only way to add decryption in case the cache is not ready) would imply changing `sdkFactory` and every single part that was calling sdk methods. It was too many changes, as most of them are sync, which means too many possibilities for errors. In the end, I did for workspace credentials the same as for the token - sync but with explicit cache hydration at the app start. The rest of the secrets (workspace mnemonic, shared access tokens, user) can be made async. But token and workspace credentials are hard to change.

src/app/store/slices/workspaces/workspacesStore.ts
TamaraFinogina: I think it's not needed here at all because the patch is applied to `Partial<Workspace`>, which has no workspace user, mnemonic, or key (it's just generic info with name, createdAt, etc.). So, when this function is called, `item.workspace.id` and  `item.workspaceUser.key` should be the same as before, and we don't have to set them again. 

P.S. If they do change, then the method that changes them should handle setting the new workspace, not expect this method to do it.

src/services/encrypted-storage.service.ts
sg-gs: Which are the potential implications of it then? Are there any risks to assess?

src/services/encrypted-storage.service.ts
```diff
@@ -82,6 +92,46 @@ const clear = (): void => {
...
+const setWorkspaceCredentials = async (credentials: WorkspaceCredentialsDetails): Promise<void> => {
+  workspaceCredentialsCache = credentials;
+  await setAndEncrypt(LocalStorageProtectedItem.EncryptedWorkspaceCredentials, JSON.stringify(credentials));
+};
+
+const clearWorkspaceCredentials = (): void => {
+  workspaceCredentialsCache = null;
+  localStorage.removeItem(LocalStorageProtectedItem.EncryptedWorkspaceCredentials);
   localStorage.removeItem(LocalStorageProtectedItem.EncryptedFolderToken);
   localStorage.removeItem(LocalStorageProtectedItem.EncryptedFileToken);
```
CandelR: If `folderTokenCache` and `fileTokenCache` belong to workspaces, shouldn’t the caches be reset here as well? Otherwise, they might remain and not be updated at some point

TamaraFinogina: They shouldn't, but somehow they ended up here after conflict resolution in a merge with master

https://github.com/internxt/drive-web/pull/2082/changes/dc384eb483c81be3fa43e6218b0d13e08a8a23d1

Removed them from this function

CandelR (dismissed):

src/services/encrypted-storage.service.ts
TamaraFinogina: When I tried using async for getWorkspaceCredentials, I ran into many React reducers and actions that can only be sync and must rely on useEffect for any async action. Plus, lots of changes in the code and tests to convert all functions to async. 

I did it for getUser/setUser: https://github.com/internxt/drive-web/pull/2088, so you can take a look to get an idea.  It seems like a lot of lines, but it was actually way better than for getToken/getWorkspaceCredentials

CandelR (approved):

sg-gs (approved):

PR #2084, opened 2026-07-30
[_] bugfix/ fix workspace folder reload

TamaraFinogina:
## Description

Reloading the page while being inside a folder in a business workspace was showing fetch folder errors and temporarily showed individual workspace folders. 

## Checklist

- [ ] QA Passed

## Testing Process

Navigate to a folder in a business workspace and do reload, then go back to the workspace. 

## Additional Notes

P.S. Still way too many requests for the server, not sure why


src/app/store/slices/workspaces/workspacesStore.ts
```diff
@@ -118,6 +118,9 @@ const setSelectedWorkspace = createAsyncThunk<
   } else if (isSelectedWorkspace && selectedWorkspace?.workspace.id === workspaceId) {
     dispatch(workspacesActions.setSelectedWorkspace(selectedWorkspace));
   } else {
+    if (state.workspaces.workspaces.length === 0) {
+      await dispatch(fetchWorkspaces());
+    }
     const workspace = state.workspaces.workspaces.find((workspace) => workspace.workspace.id === workspaceId);
```
CandelR: I would say that this state will not be up to date after the fetch, as the state was retrieved earlier. You should use `getState` instead of the state retrieved previously.
Add tests to detect this

src/views/Drive/DriveView.tsx
```diff
@@ -188,7 +199,13 @@ const DriveView = (props: DriveViewProps) => {
...
         <link rel="canonical" href={`${envService.getVariable('hostname')}`} />
       </Helmet>
-      <DriveExplorer title={<BreadcrumbsDriveView namePath={namePath} />} isLoading={isLoading} items={items} />
+      {isDriveContextReady ? (
+        <DriveExplorer title={<BreadcrumbsDriveView namePath={namePath} />} isLoading={isLoading} items={items} />
+      ) : (
+        <div className="flex h-full w-full items-center justify-center">
+          <Loader />
+        </div>
+      )}
```
CandelR, line 208: this change has design dept. approvement? As this is a bug, we can merge it, but it will need to be reviewed to ensure a sound design solution is found

TamaraFinogina: Without it, we get folders from the individual workspace for a split-second

src/app/store/slices/workspaces/workspacesStore.ts
TamaraFinogina: What do you mean? This if should be accessed only if the first fetch workspace is in progress to prevent null/underfined workspace id

CandelR (approved):

PR #2110, opened 2026-08-11
[PB-6719]: fix/apple pay on ios

jzunigax2:
## Description

Fixes two defects in the checkout payment flow.

**1. Apple Pay failed at confirmation**
`onCheckoutButtonClicked` collected the payment method from Elements twice: first on `elements.submit()` + `stripe.createConfirmationToken({ elements })` this first call prompts the Apple Pay sheet to appear, the resulting ConfirmationToken was then used only to read a postal code and discarded. Then confirmPayment({ elements, … })` at the end of the flow by passing `elements` asks Stripe.js to collect again, which for Apple Pay means opening a second `ApplePaySession`.

By that point the flow has awaited a captcha, an optional sign up, a second captcha, customer creation and intent creation, so Safari refuses the second sheet (`IntegrationError: show() must be called from a user gesture handler`) and the payment is never confirmed. Known issue:
[stripe-js#289](https://github.com/stripe/stripe-js/issues/289),

Confirmation now uses the ConfirmationToken that was already created, per Stripe's documented [two-step confirmation](https://docs.stripe.com/payments/build-a-two-step-confirmation) flow:

```ts
await confirmIntent({
  clientSecret,
  confirmParams: { confirmation_token: confirmationTokenId, return_url },
});
```

**2. Wallets showed the pre-discount total when a 100% OFF coupon was applied.**
The elements `amount` is synced into the already mounted Stripe Elements instance whenever the total changes, but the sync was skipped for any amount below Stripe's minimum charge, so a 100% OFF code left the Apple Pay / Google Pay sheet showing the old, undiscounted total.

Stripe accepts an amount of `0` when the Elements instance is in `subscription` mode, so the gate introduced in #2106 now allows `0` when the elements were mounted in `subscription` mode, and keeps rejecting it for `payment` mode (lifetime) plus anything in the 1–49 minor-unit window, which is invalid in both modes.

## Checklist

- [x] QA Passed

## Testing Process
Attempt a checkout on apple pay from an iOS device in Safari, checkout should now complete. Since this refactors checkout a bit all payment methods should undergo a regression test

## Additional Notes
- The Apple Pay defect is pre-existing, not a regression from the discounted-total fix from before.
- A lifetime plan with a 100% OFF coupon still shows the undiscounted total in the wallet sheet, because `mode: 'payment'` genuinely rejects an amount of `0`. I deliberately left this out of scope


src/views/Checkout/hooks/useUserPayment.ts
```diff
@@ -173,11 +159,21 @@ export const useUserPayment = () => {
 
     switch (subscription.type) {
       case 'payment':
-        await confirmStripePaymentIntent(elements, subscription.clientSecret, confirmPayment);
+        await confirmStripeIntent(
+          confirmationTokenId,
+          subscription.clientSecret,
+          confirmPayment,
+          'Missing payment details to confirm the payment',
```
xabg2: Passing the error message as prop is a bit weird. Also, we need to add translations for that, so the error is showed in the correct language.

jzunigax2: addressed

xabg2 (approved):
LGTM 🚀

larryrider (approved):

PR #2119, opened 2026-08-20
[MKT_1008]:feat:/redirect checkout

jaaaaavier:
## Description
The purpose of this PR is to redirect to another plan during checkout. The reason is that we cannot simply modify the URL, we need this change to be made internally.

To do this, we’ve done the following:
- Created a map of deprecated plans and associated them directly with the plan we want to redirect to.
- A function that handles this redirection if it detects the priceId of the deprecated product
- A test that covers the cases in which this function is triggered

It’s also important that the user doesn’t notice the redirect we’re performing… so we created the `deprecatedPlanIdsPlugin` to handle this. This plugin is based on the `transformIndexHtml` hook, which allows us to modify `index.html` during the build or when the development server starts up

## Checklist

- [ ] QA Passed

## Testing Process
Opened the link https://b9258af4.drive-web.pages.dev/checkout?planId=price_1T1xQtFAOdcgaBMQ1r2JnHsE&couponCode=SAVE&planType=individual&currency=eur&mode=payment&irclickid=VDATvXyxdxyZRKTUCo0LBx1tUkr0ebVJAzMh2U0&irgwc=1&afsrc=1&utm_source=Impact&utm_medium=referral&utm_campaign=3094156

And checked that redirects correctly to

 https://b9258af4.drive-web.pages.dev/checkout?planId=price_1U6Ev3FAOdcgaBMQHxOAmWPO&couponCode=SAVE&planType=individual&currency=eur&mode=payment&irclickid=VDATvXyxdxyZRKTUCo0LBx1tUkr0ebVJAzMh2U0&irgwc=1&afsrc=1&utm_source=Impact&utm_medium=referral&utm_campaign=3094156


src/utils/deprecatedPlanIds.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  beforeAll(() => {
+    originalUrl = window.location.href;
+  });
+
+  afterEach(() => {
+    window.history.replaceState(null, '', originalUrl);
+  });
+
+  describe('When a user lands on the checkout with a price that is no longer on sale', () => {
+    it('Then the checkout opens with the price that replaced it', () => {
```
larryrider: use the same name convention we are using here at drive-web

describe('group of tests')

test('when... then...')

vite.config.mts
```diff
@@ -2,19 +2,33 @@ import react from '@vitejs/plugin-react';
...
+      children: `(${rewriteDeprecatedPlanId.toString()})(${JSON.stringify(DEPRECATED_PLAN_IDS)});`,
+      injectTo: 'head-prepend',
+    },
+  ],
+});
+
 export default defineConfig({
   base: process.env.PUBLIC_URL ?? '/',
   plugins: [
+    deprecatedPlanIdsPlugin(),
```
larryrider: im not sure about adding that functionality here, even if the user should not see that redirect, I think we can manage routes and redirects like that on the https://github.com/internxt/drive-web/blob/master/src/App.tsx without needing to add it as a plugin here 🤔

jaaaaavier: Yes, better

jaaaaavier:
OK, the Vite plugin is no longer there and the price change is now handled in `DeprecatedPlanIdRedirect.tsx`, which wraps the `<Switch>` in `App.tsx`. It wraps it rather than placing it inside because `App` is the parent of the `<Router>` and does not re-render when the history changes, so a `<Redirect from="/checkout">` would still take precedence and the checkout would never actually load.

larryrider:
> OK, the Vite plugin is no longer there and the price change is now handled in `DeprecatedPlanIdRedirect.tsx`, which wraps the `<Switch>` in `App.tsx`. It wraps it rather than placing it inside because `App` is the parent of the `<Router>` and does not re-render when the history changes, so a `<Redirect from="/checkout">` would still take precedence and the checkout would never actually load.

are you sure about this? As I can see, inside the <DeprecatedPlanIdRedirect> there is also a <Redirect> so whats the difference between being the parent or not? 🤔 If there are other conflicting routes then this one could be the first

larryrider (approved):

PR #2039, opened 2026-07-13
[PB-6531]feat/skip item option in name collision

jaaaaavier:
## Description

Stacked on #2148 (which sits on #2147 and #2146). Adds a Skip option to the name-collision dialog, next to Replace and Keep both, for both upload and move collisions. Duplicates are now resolved one at a time, and a new "Apply this action to all duplicates" checkbox handles them all at once, closing the dialog immediately and running the work in the background.

The container logic lives in `nameCollision.actions.ts` and `nameCollision.utils.ts` (extracted in #2146 and #2147), so this PR adds the skip branch to the resolvers, the pending-group bookkeeping for the one-at-a-time flow, and their unit tests. The name-based matching and batching fixes that used to be bundled here are now #2148.

Also: the renameModal i18n block is renamed to alreadyExistsModal (the dialog doesn't rename anything), the description text is operation-neutral, and the apply-to-all label toggles the checkbox.

Skipping a folder is a no-op here; merging its new content into the existing folder comes in #2129, stacked on this one.

## Related Pull Requests

- Previous: #2148
- Next: #2129

## Checklist

- [x] Changes have been tested locally.
- [ ] Unit tests have been written or updated as necessary.
- [x] The code adheres to the repository's coding standards.
- [ ] Relevant documentation has been added or updated.
- [x] No new warnings or errors have been introduced.
- [ ] SonarCloud issues have been reviewed and addressed.
- [ ] QA Passed

## Testing Process
Tested uploading existing files, then, verified that all functions work as expected

PR #2129, opened 2026-08-26
[PB-6531]: merge folder uploads into the existing folder on skip

terrerox:
## Description

Stacked on #2150, which sits on #2039. Skipping a folder that already exists now merges it into the existing one: existing files are ignored, new files and subfolders are uploaded, and colliding subfolders are merged recursively so the structure is preserved. The merge helpers live in `nameCollision.actions.ts` and upload skips route to them; the unit suite covers the recursion.

Second commit adds the Playwright coverage that needs this PR or its mock extensions: keep both for files and folders, replacing a folder, merging a skipped folder into the existing one, replace, keep both and skip for moved files, and versioned replace. The Drive mock from #2150 gains folders, folder creation, moves and a versioning flag; moves are driven by dispatching the HTML5 drag events on the row's drop zone (a `data-test` hook on it). Flows that wait for a bucket upload never settle with the bridge blocked, so those specs assert on recorded requests and the task panel instead of the dialog closing; the versioned replace check is Chromium-only.

## Related Pull Requests

- Previous: #2150

## Checklist

- [ ] QA Passed

## Testing Process

`yarn vitest run src/app/drive/components/NameCollisionDialog` and `yarn playwright test test/e2e/tests/specs/DRIVE-internxt-name-collision-resolutions.spec.ts test/e2e/tests/specs/DRIVE-internxt-name-collision-skip.spec.ts --project "Internxt E2E tests on chromium"` with the dev server running.

### DIFF
--- a/src/app/drive/components/NameCollisionDialog/NameCollisionContainer.tsx
+++ b/src/app/drive/components/NameCollisionDialog/NameCollisionContainer.tsx
@@ -21,6 +21,10 @@ import { getUniqueFilename } from 'app/store/slices/storage/fileUtils/getUniqueF
 import { checkDuplicatedFiles } from 'app/store/slices/storage/fileUtils/checkDuplicatedFiles';
 import { items as itemUtils } from '@internxt/lib';
 import { CollisionGroup } from 'app/store/slices/storage/storage.model';
+import {
+  handleRepeatedUploadingFiles,
+  handleRepeatedUploadingFolders,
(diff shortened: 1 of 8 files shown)


CandelR (commented):
I find it odd that the ‘skip’ feature doesn’t skip uploading the folder, instead, it skips the ‘replace’ step and uploads the items into the existing folder. Has this been confirmed by Product? I find it confusing

PR #2142, opened 2026-09-09
[BR-2247]: abort stalled upload requests instead of waiting forever

terrerox:
## Description

PUT requests to the storage host had no timeout, so a connection that went silent without closing kept the upload promise pending forever. No error meant no retry, and the task sat at the same percentage with nothing in the console. Reproduced in Chromium by freezing the storage connections mid-upload: a folder of small files and a 150 MB multipart file both hung indefinitely.

Add a stall watchdog that aborts the request when no bytes move for UPLOAD_IDLE_TIMEOUT_MS (60 s), re-armed on every progress event so slow but live uploads are unaffected. The request rejects with "Upload stalled", which the existing retry logic already handles. With the same freeze both scenarios now log a failed attempt after a minute, retry over fresh connections and complete.

## Checklist

- [ ] QA Passed


src/app/network/upload-utils.test.ts
```diff
@@ -60,3 +60,69 @@ describe('uploadFileUint8Array error handling', () => {
     ).rejects.toThrow('Unknown error');
   });
 });
+
+describe('uploadFileUint8Array stall handling', () => {
```
CandelR: The two new tests cover the happy path and the stall itself, but the branches this PR actually changed are still untested. The request no longer receives opts.abortController.signal directly, it now goes through the watchdog. Add more tests to check all possible paths

src/app/network/upload-utils.ts
```diff
@@ -1,15 +1,22 @@
...
 
+export const UPLOAD_STALLED_ERROR_MESSAGE = 'Upload stalled';
+
 export async function uploadFileUint8Array(
   content: Uint8Array,
   url: string,
   opts: {
     progressCallback: UploadProgressCallback;
     abortController?: AbortController;
+    idleTimeoutMs?: number;
```
CandelR, line 15: this arg is not used, if it is for the time of tests, with faketimers I think is enough

CandelR (dismissed):

CandelR (dismissed):

CandelR (approved):

PR #2145, opened 2026-09-11
[PB-6531]: extract name collision logic out of the container

terrerox:
## Description

Stacked on #2039. Pure refactor, no behaviour change.

The name-collision container had the same `operationType + operation` switch twice (once for "apply to all", once for the one-at-a-time flow) and held all the item matching and per-operation side effects inline. This moves that logic out of the component:

- `nameCollision.utils.ts`: pure helpers for matching an incoming item to its existing counterpart, pairing items, finding the next pending group and computing the remaining groups after a decision.
- `nameCollision.actions.ts`: the side effects (keep/replace for move and upload, versioned replace) behind a single `resolveCollision` that takes the items to handle plus a small context (dispatch, workspace, upload limit, versioning flag). Both dialog paths call it, so the duplicated switch is gone.
- `NameCollisionContainer.tsx` now only reads store state, builds the context and decides whether to close or reopen the dialog.

Both modules have unit tests (100% line coverage on the new files). #2129 is rebased on top of this so the merge-on-skip logic lands as one extra branch in `resolveCollision`.

## Related Pull Requests

- [internxt/drive-web#2039](https://github.com/internxt/drive-web/pull/2039)
- [internxt/drive-web#2129](https://github.com/internxt/drive-web/pull/2129)

## Checklist

- [ ] QA Passed

## Testing Process

`yarn vitest run src/app/drive/components/NameCollisionDialog` plus the same manual flows as #2039 (upload/move collisions with keep, replace, skip, per item and apply to all).


terrerox:
Superseded by the reordered stack: the refactor now lands before the feature as #2146 and #2147, with the matching/batching fixes in #2148, followed by #2039 and #2129.

PR #2146, opened 2026-09-11
[PB-6531]: extract name collision move actions out of the container

terrerox:
## Description

First of a stack that ends in #2039 (skip option) and #2129 (merge folders on skip). Pure refactor, no behaviour change.

The name-collision container held every side effect inline. This moves the move-collision handling (keep both with a unique name, replace by trashing the existing item) into `nameCollision.actions.ts` behind `resolveMoveCollision`, which takes the items plus a small context (dispatch, workspace, upload limit, versioning flag) instead of closing over component state. Upload collisions stay in the container for now and move in the next PR.

Unit tests cover both move resolutions.

## Related Pull Requests

- Next: refactor\/extract-name-collision-upload-actions

## Checklist

- [ ] QA Passed

## Testing Process

`yarn vitest run src/app/drive/components/NameCollisionDialog`, plus the manual upload/move collision flows (keep both, replace, with and without versioning).

### DIFF
--- a/src/app/drive/components/NameCollisionDialog/nameCollision.actions.test.ts
+++ b/src/app/drive/components/NameCollisionDialog/nameCollision.actions.test.ts
@@ -1,0 +1,89 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { getDriveItemData } from 'testUtils/fixtures/drive.fixtures';
+import { NameCollisionContext, ResolveMoveCollisionParams, resolveMoveCollision } from './nameCollision.actions';
+
+const mocks = vi.hoisted(() => ({
+  moveItemsToTrash: vi.fn(),
+  moveItemsThunk: vi.fn(),
+  popItemsToDelete: vi.fn(),
+  checkDuplicatedFiles: vi.fn(),
+  getUniqueFilename: vi.fn(),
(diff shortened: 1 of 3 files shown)


src/app/drive/components/NameCollisionDialog/nameCollision.actions.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const getContext = (): NameCollisionContext => ({
+  dispatch: vi.fn() as unknown as NameCollisionContext['dispatch'],
+  selectedWorkspace: null,
+  maxUploadFileSize: 5000,
+  isVersioningEnabled: false,
+});
+
+const resolve = (params: Omit<ResolveMoveCollisionParams, 'destinationUuid'>) =>
+  resolveMoveCollision({ ...params, destinationUuid: DESTINATION }, getContext());
```
CandelR, line 43: the context is created inside resolve, so the tests never get hold of dispatch. They check that moveItemsThunk and popItemsToDelete are called, but not that their results are dispatched. If the dispatch(...) wrapper is dropped in` nameCollision.actions.ts`, both tests still pass while nothing is actually moved or removed from the pending list.

CandelR (approved):

PR #2148, opened 2026-09-11
[PB-6531]: match name collisions by name and type and batch their resolution

terrerox:
## Description

Stacked on `refactor/extract-name-collision-upload-actions`. Two fixes to how name collisions are resolved, split out of #2039 so that PR only adds the Skip option.

- Duplicated items were matched to their existing counterpart by array index, which could make Replace trash the wrong file. They are now paired by name and type (name only for folders) via `findExistingItemFor` in `nameCollision.utils.ts`, and items without a match are left out of the replace.
- Each resolution goes through the array-based primitives (`moveItemsToTrash`, `moveItemsThunk`, the upload managers) in one call instead of one call per item, so batching, concurrency limits and per-item error handling are inherited rather than reimplemented. Versioned replacements still run one at a time because that upload bypasses the queue. The destination folder is refreshed once per group instead of once per item.

Unit tests cover the matching rules and every resolution path.

## Related Pull Requests

- Previous: refactor\/extract-name-collision-upload-actions
- Next: #2039

## Checklist

- [ ] QA Passed

## Testing Process

`yarn vitest run src/app/drive/components/NameCollisionDialog`, plus the manual upload/move collision flows (keep both, replace, with and without versioning).


src/app/store/slices/storage/storage.thunks/renameItemsThunk.ts
```diff
@@ -17,6 +17,13 @@ import { getUniqueFolderName } from '../folderUtils/getUniqueFolderName';
 import { CollisionGroup, StorageState } from '../storage.model';
 import { IRoot } from '../types';
 
+/**
+ * The duplicate checks return raw API items without isFolder, so existing items are tagged
+ * here to let consumers tell colliding files and folders apart.
+ */
+const asExistingItems = (items: (DriveFileData | DriveFolderData)[], isFolder: boolean): DriveItemData[] =>
```
CandelR, line 24: nit: asExistingItems reads like a cast but it's actually adding isFolder field, maybe addIsFolderField or similar would be more honest about what it does

src/app/drive/components/NameCollisionDialog/nameCollision.actions.ts
```diff
@@ -58,30 +58,36 @@ const getUniqueNameMovePayload = async (item: DriveItemData, destinationUuid: st
...
-) => {
-  for (const item of items) {
-    const itemParsed = await getUniqueNameMovePayload(item, destinationUuid);
-    await dispatch(storageThunks.moveItemsThunk({ items: [itemParsed], destinationFolderId: destinationUuid }));
-  }
+): Promise<void> => {
+  if (items.length === 0) return;
+
+  const itemsParsed = await Promise.all(items.map((item) => getUniqueNameMovePayload(item, destinationUuid)));
+  await dispatch(storageThunks.moveItemsThunk({ items: itemsParsed, destinationFolderId: destinationUuid }));
```
CandelR, line 71: maybe would be fine to check if some items have same name+extension, because I think that from trash could go items with same name and extension to same folder and will produce errors

terrerox: I Fixed in #2159, with e2e in #2160 because this PR will exceed kloc limit.

CandelR (approved):

PR #2161, opened 2026-09-21
[PB-6882]: prevent account blocked message and hide turnstile widget on success

apsantiso:
## Description

Fixes the Turnstile challenge widget staying visible on screen after the check completes, and the login button getting stuck in a loading state.

- Prevents moving users to account blocked screen when a 403 code is received. Any invalid turnstile token returns a 403 code tood so the check was being triggered. I modified the backend so it returns an error unique code in the "error" field, it's better than relying on the http code.
- `TurnstileWidget`: The widget container is now hidden by default (previous it was hidden but taking space) and only expands when Cloudflare actually requires interaction (`onBeforeInteractive`), and collapses again once a token is obtained. `getToken()`. Every submittion after the first try will request a token again.
- LogIn.tsx: Skip the redundant /login call when a user logs in. Previously, `is2FANeeded` called /login, and the SDK's authentication function called it again afterward.

## Checklist

- [ ] QA Passed


apsantiso:
@CandelR @sg-gs Feel free to test it using -> https://fix-prevent-account-blocked.drive-web.pages.dev/

You can test the behaviour when the widget was not able to load by using this URL -> https://963f2b2d.drive-web.pages.dev/

You can force the interaction by changing the value of `REACT_APP_TURNSTILE_SITE_KEY` on the preview environment to `3x00000000000000000000FF`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "drive-web",
-  "version": "v1.1.0",
+  "version": "v2.0.0",
```
CandelR: major version? I don’t think it’s that big a change haha

sg-gs (approved):

package.json
sg-gs: This is for another reason, but we should be able to avoid it @apsantiso

CandelR (approved):

CandelR (approved):

PR #2165, opened 2026-09-21
[PB-6867]:feat/ add brazilian portuguese as language

jaaaaavier:
## Description
Adds Brazilian Portuguese (pt-BR) as the 9th supported language.

The full 1193-key `en.json` was translated into `src/app/i18n/locales/pt.json`, keeping brand names (Internxt, Drive, VPN, Antivirus, Cleaner, Meet, Mail), plan names , the 41 `{{placeholders}}` and the single HTML-bearing key untouched.

The locale is registered as `pt-BR` in i18next and as `pt-br` in the selector,mirroring the existing `zh-TW` / `zh-tw` asymmetry rather than introducing a new convention. The file is named `pt.json` following the repo's `tw.json` precedent.

`pt-BR` carries a region suffix, and dayjs locale resolution did not handle it: the `.split('-')` was only applied to `navigator.language`, never to the value read from localStorage — which is exactly where regional codes are stored. 

Any user on `es-ES`, `fr-FR`, `de-DE` or `zh-TW` got English dates under a translated UI. Fixing it was a prerequisite for pt-BR;it corrects the other locales as a side effect.

`getCalendarLocale` needed the same treatment: `pt-BR` resolved to `pt` (European Portuguese), so regional locales are now matched from a list.

## Checklist

- [ ] QA Passed

## Testing Process
- `yarn check:i18n` — all keys in sync
- `yarn typecheck` — clean
- `yarn build` — builds


CandelR (commented):
Yep @jaaaaavier , there’s a bug that causes the language translation to not display correctly in the language selector if the browser’s default language is Brazilian Portuguese. It seems to happen with other languages too, so it’s been an issue for a while. Have a look at it, so we don’t release this new language with a bug that users with their browser set to Brazilian Portuguese will spot straight away. To test it, set your browser language to Brazilian Portuguese and go into incognito mode. I’ve attached a screenshot.

Image

jaaaaavier:
> Yep @jaaaaavier , there’s a bug that causes the language translation to not display correctly in the language selector if the browser’s default language is Brazilian Portuguese. It seems to happen with other languages too, so it’s been an issue for a while. Have a look at it, so we don’t release this new language with a bug that users with their browser set to Brazilian Portuguese will spot straight away. To test it, set your browser language to Brazilian Portuguese and go into incognito mode. I’ve attached a screenshot.
> 
> Image

Fixed by updating the `sanitizeLanguage`, now converts the code to lowercase, but only if that version exists in the list of supported languages.

PR #2171, opened 2026-09-21
[MKT-1095]feat/create new checkout page adding view and route

internxt-yu:
## Description

We are creating a new checkout page with the goal to test whether a shorter, more urgent, lower-friction checkout page converts better than our current flow. Having the new checkout in route/checkout-bf.
We aim to two type of client, mainstream and adult client who is more likely to store adult content. So in the new checkout we change the image depeding the the type of client using parameter variant in URL.
The new checkout page should looking like the image below:

ChatGPT Image 17 sept 2026, 15_48_38

## Checklist

- [ ] QA Passed

## Testing Process

1. Access to https://c5a49407.drive-web.pages.dev/checkout-uc?planId=<Price_id>&couponCode=\<CouponCode\>&variant=\<Audience\>, replace the <Price_id> and \<CouponCode\> with a valid price_id and couponCode in Stripe, and for variant it could be 'mainstream' or 'adult' using differents images for different type of audience.
2. Do a test purchase in the page and see if it succed.


jaaaaavier (changes requested):
Use common names, not BFxxx

jaaaaavier (commented):
Will need some proves that the new checkout really works, not only free customers you can create a 0.01 code, make the refund, test with stripe test mode or something, attach a video or stripe link where we can see the details on the task

PR #2173, opened 2026-09-22
[MKT-1095]feat/create new checkout page adding components

internxt-yu:
## Description

Added new components to be used in the new checkout page.

BfCheckoutHeader:
Page header, with Internxt logo, campaign tagline, and three trust badges (European based, GDPR compliant, trusted by millions).

BfCheckoutHero:
Left column. Renders the discount pill, the headline and subtitle interpolated with storage and price, three selling points (privacy, storage, offer), the campaign creative, and four footer badges.

BfCheckoutOrderSummary:
discount chip, plan name with the original price struck through, billing row, estimated tax, savings and total.

BfCheckoutAccountFields:
Account block, to the login of user who already has an account or register of user who doesn'st has an account. The same as the original checkout page.

BfCheckoutPaymentSection:
Payment area, Renders the Stripe PaymentElement, the crypto selector, the accepted card brands, the submit button labelled with storage and price, the renewal note for recurring plans, and the money-back guarantee.

## Checklist

- [ ] QA Passed


src/views/UrgentCheckout/components/UrgentCheckoutPaymentSection.tsx
```diff
@@ -0,0 +1,126 @@
...
+          options={URGENT_CHECKOUT_PAYMENT_ELEMENT_OPTIONS}
+          onChange={(event) => {
+            if (!event.collapsed) {
+              onPaymentExpanded();
+            }
+          }}
+        />
+        {availableCryptoCurrencies && (
+          <div
+            data-testid="bf-crypto-section"
```
CandelR: I think that this data-testid has slipped in

src/views/UrgentCheckout/components/UrgentCheckoutHero.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div className="flex w-fit flex-row items-center rounded-full border border-[#1E4A8F] bg-[#0E1E38] px-5 py-2">
+        <p className="text-sm font-semibold text-[#4D9BFF]">{offerPillLabel}</p>
+      </div>
+
+      <div className="flex flex-col gap-4">
+        <h1 className="text-4xl font-bold leading-tight text-white sm:text-5xl">
+          {translate('urgentCheckout.hero.title')}{' '}
+          <span className="text-[#4D9BFF]">{translate('urgentCheckout.hero.titleHighlight')}</span>
+        </h1>
+        <p className="text-xl leading-snug text-[#C4CEDD]">
```
CandelR: Colours are normally taken from Tailwind config. 
If it’s only for this checkout, that’s fine, but for the next one, if any colours are to be added to the colour palette that will be reused frequently, it’s better to add them to the Tailwind configuration. I’m just mentioning this so you bear it in mind.

CandelR (dismissed):

src/views/UrgentCheckout/components/UrgentCheckoutPaymentSection.tsx
internxt-yu: It was a id introduced for a test in UrgentCheckoutView.test.tsx in the next PR, now i'm thinking is not proper introducing ids for test purpose, it's removed and updated test in the next PR without using id.

src/views/UrgentCheckout/components/UrgentCheckoutHero.tsx
internxt-yu: It's used only here, i'll keep this in mind if in the future we want to reuse them.

CandelR (approved):

